PR #122, opened 2019-09-09
Move from dep to go mod

### LINKED ISSUE #103: Could we move to go modules?
Currently, the instructions to fetch the repo, play with samples, and run the `logs` CLI are:

```
git clone https://github.com/pivotal/kpack \
   $GOPATH/src/github.com/pivotal/kpack
cd $GOPATH/src/github.com/pivotal/kpack
dep ensure
go run ./cmd/logs
```

If we move to go modules, then I think these instructions become nicer, and less dependent on old `$GOPATH`:

```
git clone https://github.com/pivotal/kpack
cd kpack
go run ./cmd/logs
```

joaopapereira:
Closes #103


hack/update-codegen.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright 2018 The Knative Authors
+# Copyright 2017 The Kubernetes Authors.
```
ashwin-venkatesh: hehe

PR #129, opened 2019-09-13
Prevent duplicate entries in .docker/config.json

matthewmcnew:
- build-init does not write the same registry to .docker/config.json if it already exists in a different format
- Use builderPullSecrets when fetching builder metadata in build-init

Signed-off-by: Joao Pereira <jdealmeidapereira@pivotal.io>

### DIFF
--- a/pkg/dockercreds/docker_creds.go
+++ b/pkg/dockercreds/docker_creds.go
@@ -14,10 +14,8 @@ import (
 type DockerCreds map[string]entry
 
 func (c DockerCreds) Resolve(reg name.Registry) (authn.Authenticator, error) {
-	registryMatcher := RegistryMatcher{}
-
 	for registry, entry := range c {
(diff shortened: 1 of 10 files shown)


pkg/dockercreds/docker_creds.go
```diff
@@ -47,15 +53,41 @@ func (c DockerCreds) AppendToDockerConfig(path string) error {
...
+	}
+
+	u, err := url.Parse(reg)
+	if err != nil {
+		return false, err
 	}
-	return c
+
+	for existingRegistry := range c {
+		if (RegistryMatcher{}.Match(u.Host, existingRegistry)) {
```
ashwin-venkatesh: Can we update this method to new require a receiver and just be a public method in the pkg?

matthewmcnew: Fixed 🙂

PR #132, opened 2019-09-16
Update image status if builder is not ready

ashwin-venkatesh:
Update image status to `builderNotReady` if the builder is not ready.

Additionally, if the builder metadata is not available (because of incorrect credentials on a private builder), this does not try to constantly re-enqueue the builder/clusterbuilder resource

Additionally added a sleep in our tests as we now wait for the builders to be ready for images to get scheduled.

### DIFF
--- a/pkg/apis/build/v1alpha1/image_builds.go
+++ b/pkg/apis/build/v1alpha1/image_builds.go
@@ -28,6 +28,7 @@ type AbstractBuilder interface {
 	ImageRef() BuilderImage
 	Ready() bool
 	BuildpackMetadata() BuildpackMetadataList
+	GetName() string
 }
 
 func (im *Image) buildNeeded(lastBuild *Build, sourceResolver *SourceResolver, builder AbstractBuilder) ([]string, bool) {
--- a/pkg/apis/build/v1alpha1/image_lifecycle.go
+++ b/pkg/apis/build/v1alpha1/image_lifecycle.go
@@ -22,14 +22,3 @@ func (im *Image) BuilderNotFound() duckv1alpha1.Conditions {
 		},
 	}
 }
-
-func (im *Image) BuilderNotReady() duckv1alpha1.Conditions {
-	return duckv1alpha1.Conditions{
-		{
-			Type:    duckv1alpha1.ConditionReady,
-			Status:  corev1.ConditionFalse,
-			Reason:  BuilderNotReady,
-			Message: fmt.Sprintf("Builder %s is not ready.", im.Spec.Builder.Name),
-		},
-	}
-}
--- a/pkg/apis/build/v1alpha1/image_lifecycle.go
+++ b/pkg/apis/build/v1alpha1/image_lifecycle.go
@@ -33,14 +33,3 @@ func (im *Image) BuilderNotReady() duckv1alpha1.Conditions {
 		},
 	}
 }
-
-func (b *Build) BuilderNotFound() duckv1alpha1.Conditions {
-	return duckv1alpha1.Conditions{
-		{
-			Type:    duckv1alpha1.ConditionSucceeded,
-			Status:  corev1.ConditionFalse,
-			Reason:  BuilderNotFound,
-			Message: fmt.Sprintf("Unable to find builder %s.", b.Spec.Builder.Image),
-		},
-	}
-}
--- a/pkg/apis/build/v1alpha1/image_lifecycle.go
+++ b/pkg/apis/build/v1alpha1/image_lifecycle.go
@@ -9,6 +9,7 @@ import (
 
 const (
 	BuilderNotFound = "BuilderNotFound"
+	BuilderNotReady = "BuilderNotReady"
 )
 
 func (im *Image) BuilderNotFound() duckv1alpha1.Conditions {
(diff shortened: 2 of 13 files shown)


pkg/reconciler/v1alpha1/builder/builder.go
```diff
@@ -122,16 +122,16 @@ func (c *Reconciler) reconcileBuilderStatus(builder *v1alpha1.Builder) reconcile
...
-						Status: corev1.ConditionFalse,
+						Type:    duckv1alpha1.ConditionReady,
+						Status:  corev1.ConditionFalse,
+						Message: err.Error(),
 					},
 				},
 			},
 		}
 
 		return reconciledBuilderResult{
```
matthewmcnew: If we never return an error anymore... Let's remove the reconciledBuilderResult.

pkg/reconciler/v1alpha1/image/image.go
```diff
@@ -125,12 +125,20 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
-	if k8serrors.IsNotFound(err) {
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return nil, err
+	} else if k8serrors.IsNotFound(err) {
 		image.Status.Conditions = image.BuilderNotFound()
 		image.Status.ObservedGeneration = image.Generation
 		return image, nil
 	}
 
+	if !builder.Ready() {
```
matthewmcnew: Doing this check here will short circuit all the the other image reconciliation logic. 
While a builder 'is not ready': 
- this will cause existing Conditions to be overwritten 
- prevent valuable updates from taking place to the existing image object (like latestImage) 

Could we append this to the conditions when processing the build? 
Perhaps append on this line?https://github.com/pivotal/kpack/blob/d2b6381a892a9a1659ea2302d276f9167ffff918/pkg/apis/build/v1alpha1/image_reconcile_build.go#L56

This would also enable use to add a condition when the sourceresolver is not ready.

test/execute_build_test.go
```diff
@@ -189,6 +188,8 @@ func testCreateImage(t *testing.T, when spec.G, it spec.S) {
 				},
 			}
 
+			time.Sleep(10 * time.Second)
```
matthewmcnew: What is this sleep for?

pkg/apis/build/v1alpha1/image_lifecycle.go
```diff
@@ -22,6 +23,17 @@ func (im *Image) BuilderNotFound() duckv1alpha1.Conditions {
...
+		{
+			Type:    duckv1alpha1.ConditionReady,
+			Status:  corev1.ConditionFalse,
+			Reason:  BuilderNotReady,
+			Message: fmt.Sprintf("Builder %s is not ready.", im.Spec.Builder.Name),
+		},
+	}
+}
+
 func (b *Build) BuilderNotFound() duckv1alpha1.Conditions {
```
matthewmcnew: I realize that you didn't add this now, but while we are here can we remove the `build.BuilderNotFound` method?

pkg/apis/build/v1alpha1/image_reconcile_build.go
```diff
@@ -58,6 +64,16 @@ func (r upToDateBuild) Apply(creator BuildCreator) (ReconciledBuild, error) {
 }
 
 func (r upToDateBuild) conditions() duckv1alpha1.Conditions {
+	if !r.builder.Ready() {
+		return duckv1alpha1.Conditions{
+			{
+				Type:   duckv1alpha1.ConditionReady,
+				Status: corev1.ConditionFalse,
+				Reason: BuilderNotReady,
```
matthewmcnew: Super small thing here but, because this condition is largely for human consumption can we add an easy message here?

pkg/reconciler/v1alpha1/image/image.go
```diff
@@ -125,7 +125,9 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 
 func (c *Reconciler) reconcileImage(image *v1alpha1.Image) (*v1alpha1.Image, error) {
 	builder, err := c.getBuilder(image)
-	if k8serrors.IsNotFound(err) {
+	if err != nil && !k8serrors.IsNotFound(err) {
```
matthewmcnew: Good catch

pkg/apis/build/v1alpha1/image_types.go
```diff
@@ -89,3 +89,5 @@ func (*Image) GetGroupVersionKind() schema.GroupVersionKind {
 func (i *Image) NamespacedName() types.NamespacedName {
 	return types.NamespacedName{Namespace: i.Namespace, Name: i.Name}
 }
+
+const ConditionBuilderReady duckv1alpha1.ConditionType = "Builder Ready"
```
matthewmcnew: Is it common to have a space here? Most conditions I have seen would be BuilderReady

pkg/apis/build/v1alpha1/image_builds.go
```diff
@@ -28,6 +28,7 @@ type AbstractBuilder interface {
 	ImageRef() BuilderImage
 	Ready() bool
 	BuildpackMetadata() BuildpackMetadataList
+	GetName() string
```
matthewmcnew: Super pedantic but, get Getters in go shouldn't start with 'Get`.

https://golang.org/doc/effective_go.html#Getters

PR #133, opened 2019-09-16
Remove dependency on buildpack/imgutil

### LINKED ISSUE #128: Builder.Spec.Image does not support images with digest
Currently, kpack cannot process builders with an image that includes a digest. 

```yaml
apiVersion: build.pivotal.io/v1alpha1
kind: Builder
metadata:
  name: sample-builder
spec:
  image: cloudfoundry/cnb:bionic@sha256:b100bb7ba4d9109146965dbba0c91784bef2789b4fca759f4e734d55ea89a6af
```
Will fail to successfully reconcile the builder. 

kpack controller logs:
```
a digest must contain exactly one '@' separator (e.g. registry/repository@digest) saw: cloudfoundry/cnb:bionic@sha256:b100bb7ba4d9109146965dbba0c91784bef2789b4fca759f4e734d55ea89a6af@sha256:b100bb7ba4d9109146965dbba0c91784bef2789b4fca759f4e734d55ea89a6af
creating digest reference
```

matthewmcnew:
- Create GGCRImage that implements RemoteImage interface
- Fixes #128
- Upgrades lifecycle to 0.0.4

### DIFF
--- a/pkg/cnb/cnb_metadata.go
+++ b/pkg/cnb/cnb_metadata.go
@@ -57,7 +57,7 @@ func (r *RemoteMetadataRetriever) GetBuilderImage(repo registry.ImageRef) (Build
 
 	return BuilderImage{
 		BuilderBuildpackMetadata: metadata.Buildpacks,
-		Identifier:               identifier.String(),
+		Identifier:               identifier,
 	}, nil
 }
 
@@ -90,7 +90,7 @@ func (r *RemoteMetadataRetriever) GetBuiltImage(ref registry.ImageRef) (BuiltIma
 	}
 
 	return BuiltImage{
-		Identifier:        identifier.String(),
+		Identifier:        identifier,
 		CompletedAt:       imageCreatedAt,
 		BuildpackMetadata: metadata.Buildpacks,
 	}, nil
--- a/pkg/cnb/cnb_metadata.go
+++ b/pkg/cnb/cnb_metadata.go
@@ -29,11 +29,11 @@ type BuilderImage struct {
 type BuilderMetadata []BuildpackMetadata
 
 type RemoteMetadataRetriever struct {
-	LifecycleImageFactory registry.RemoteImageFactory
+	RemoteImageFactory registry.RemoteImageFactory
 }
 
 func (r *RemoteMetadataRetriever) GetBuilderImage(repo registry.ImageRef) (BuilderImage, error) {
-	img, err := r.LifecycleImageFactory.NewRemote(repo)
+	img, err := r.RemoteImageFactory.NewRemote(repo)
 	if err != nil {
 		return BuilderImage{}, errors.Wrap(err, "unable to fetch remote builder image")
 	}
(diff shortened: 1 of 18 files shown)


pkg/registry/ggcr_image.go
```diff
@@ -0,0 +1,109 @@
...
+	"time"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/pkg/errors"
+)
+
+type GGCRImage struct {
```
ashwin-venkatesh: Not sure if `GGCR` is how the community refers to the library. We could call it `RegistryImage` possibly? That or even `RemoteImage` 🤷‍♂

pkg/registry/ggcr_image.go
```diff
@@ -0,0 +1,109 @@
...
+	image    v1.Image
+	repoName string
+}
+
+func NewGGCRImage(repoName string, keychain authn.Keychain) (*GGCRImage, error) {
```
ashwin-venkatesh: Do we want to call it `GetRemoteImage` or something similar, because we do not want to create an image in case one doesn't exist right? We only want to fetch an already existing one.

pkg/registry/image_factory.go
```diff
@@ -3,19 +3,16 @@ package registry
...
-	"github.com/buildpack/imgutil/remote"
 	"github.com/google/go-containerregistry/pkg/authn"
-	"github.com/pkg/errors"
 )
 
 type ImageFactory struct {
 	KeychainFactory KeychainFactory
 }
 
 func (f *ImageFactory) NewRemote(imageRef ImageRef) (RemoteImage, error) {
```
ashwin-venkatesh: As an extension of our previous comment, can we update this as well to say `Get/FetchRemote` because it will never be our intent on actually creating a new remote image

pkg/cnb/cnb_metadata.go
```diff
@@ -57,23 +57,23 @@ func (r *RemoteMetadataRetriever) GetBuilderImage(repo registry.ImageRef) (Build
 
 	return BuilderImage{
 		BuilderBuildpackMetadata: metadata.Buildpacks,
-		Identifier:               identifier.String(),
+		Identifier:               identifier,
```
ashwin-venkatesh: should this be Image too?

pkg/dockercreds/access_checker.go
```diff
@@ -2,27 +2,31 @@ package dockercreds
 
 import (
 	"fmt"
+	"github.com/google/go-containerregistry/pkg/name"
```
ashwin-venkatesh: import order nit

pkg/cnb/cnb_metadata.go
matthewmcnew: That is an actual built image identifier. Not the same thing as an image because it always includes the digest

out.yaml
```diff
@@ -0,0 +1,217 @@
+apiVersion: v1
```
ashwin-venkatesh: Checked in by mistake?

pkg/dockercreds/access_checker.go
```diff
@@ -60,21 +64,3 @@ func HasWriteAccess(tagName string) (bool, error) {
 
 	return true, nil
 }
-
-type unAuthorizedWithoutErrorCodeTransportChecker struct {
```
c-abramson: does the new version of ggcr make this obsolete?

pkg/registry/ggcr_image.go
```diff
@@ -0,0 +1,109 @@
...
+	"time"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/pkg/errors"
+)
+
+type GoContainerRegistryImage struct {
```
c-abramson: can we rename this file/struct ... maybe RemoteRegistryImage?

out.yaml
matthewmcnew: I finally accidentally did. Gone now :)

pkg/dockercreds/access_checker.go
matthewmcnew: Sort of. We can get the status code now directly on the transport error.

ashwin-venkatesh:
🤔 Does `dep ensure` cause a diff on your local against this branch right now? The CI pipeline failing is so bizarre.

pkg/registry/ggcr_image.go
matthewmcnew: My thought is the consumers of this object will consume it through the RemoteImage interface. What makes this struct unique is that it implements that interface using  GoContainerRegistry library.

PR #136, opened 2019-09-19
Add a default ClusterBuilder

matthewmcnew:
- kpack users can create images with cloudfoundry/cnb:bionic in any namespace


matthewmcnew:
Closing due to possible race condition.

PR #140, opened 2019-09-23
Add run image into builder status successful build

ashwin-venkatesh:
The fully qualified run image has been added to the builder status and the status of the successful builds to determine if a rebase is required once a new run image is available. Where the actual performing of the rebase will happen has not yet been determined.


pkg/apis/build/v1alpha1/build_types.go
```diff
@@ -54,11 +54,19 @@ type BuildSpec struct {
 	CacheName      string                      `json:"cacheName"`
 	Env            []corev1.EnvVar             `json:"env"`
```
joaopapereira: Maybe we can add the RunImage field to the Build Object itself. What do yll think about it?

This way we do not need the Rebase.LatestRunImage

pkg/apis/build/v1alpha1/build_types.go
```diff
@@ -54,11 +54,19 @@ type BuildSpec struct {
...
 	Resources      corev1.ResourceRequirements `json:"resources"`
+	Rebase         *RebaseConfig               `json:"rebase,omitempty"`
+}
+
+type RebaseConfig struct {
+	ImageRef         string `json:"imageRef"`
```
joaopapereira: What do we use this for?

joaopapereira: Maybe the name could be `ImageToRebaseRef`? to be more explicit

pkg/cnb/cnb_metadata.go
```diff
@@ -18,11 +19,21 @@ type BuildpackMetadata struct {
 }
 
 type BuilderImageMetadata struct {
-	Buildpacks []BuildpackMetadata `json:"buildpacks"`
+	Buildpacks []BuildpackMetadata    `json:"buildpacks"`
+	Stack      lcyclemd.StackMetadata `json:"stack"`
+}
+
+type Stack struct {
```
joaopapereira: What are we going to use this for?

pkg/cnb/cnb_metadata.go
```diff
@@ -55,8 +66,19 @@ func (r *RemoteMetadataRetriever) GetBuilderImage(repo registry.ImageRef) (Build
 		return BuilderImage{}, errors.Wrap(err, "failed to retrieve builder image SHA")
 	}
 
+	runImage, err := r.RemoteImageFactory.NewRemote(registry.NewNoAuthImageRef(metadata.Stack.RunImage.Image))
```
joaopapereira: Should this be NoAuthImageRef?

pkg/apis/build/v1alpha1/build_types.go
joaopapereira: This `ImageRef` name fells hard to understand what it is about, maybe `ImageToRebaseRef`?

pkg/apis/build/v1alpha1/build_types.go
joaopapereira: What do y'll think about adding the RunImage to the spec instead of inside the RebaseConfig?

pkg/cnb/cnb_metadata.go
ashwin-venkatesh: This is WIP where we are trying to get it to work first with public run images..will have to rethink how this works for private run images..we might have to make the runimage a type that can be an image ref that rides on things from the builder..

pkg/apis/build/v1alpha1/build_types.go
ashwin-venkatesh: Makes sense

ashwin-venkatesh: turns out we dont need it so :)

pkg/apis/build/v1alpha1/build_types.go
ashwin-venkatesh: Yeah..but would we duplicate that information on the build status as well? Would we explicitly pass this run image to the exporter?

joaopapereira: Hmm, In the end it is like the Builder itself. On each build you tell it with what RunImage/RunImageMirrors you want this build to be created. The status would be more a confirmation

pkg/cnb/cnb_metadata.go
ashwin-venkatesh: This is so that we can get the actual run image that is configured inside the builders `stack.toml` and get it's complete reference

pkg/apis/build/v1alpha1/build_types.go
ashwin-venkatesh: Well, kinda. With the builder, the build spec gets updated with the buildpacks that were used for the build and do not explicitly mentioned the SHA of the used builder AFAIK. With the run image it will seemingly be duplicated information.

pkg/apis/build/v1alpha1/build_types.go
ashwin-venkatesh: We dont need this field anymore :P

pkg/cnb/cnb_metadata.go
ashwin-venkatesh: This is to extract the RunImage that is specified inside the `stack.toml` of the builder and then try and resolve it to a SHA

pkg/apis/build/v1alpha1/build_types.go
ashwin-venkatesh: The only reason I am on the fence about this is because one the build is built, it will involve just duplicating the exact same info from the spec into the status. Maybe that isnt that big a deal, but it feels weird.

pkg/apis/build/v1alpha1/image_builds.go
```diff
@@ -106,6 +132,72 @@ func (im *Image) build(sourceResolver *SourceResolver, builder AbstractBuilder,
...
+			},
+			Labels: im.labels(map[string]string{
+				BuildNumberLabel: buildNumber,
+				ImageLabel:       im.Name,
+			}),
+			Annotations: map[string]string{
+				BuildReasonAnnotation: BuildReasonStack,
+			},
+		},
+		Spec: BuildSpec{
```
matthewmcnew: Is it problematic for the build needed logic if the last build does not have all the current source spec like last git url etc?

pkg/apis/build/v1alpha1/image_builds.go
```diff
@@ -106,6 +132,72 @@ func (im *Image) build(sourceResolver *SourceResolver, builder AbstractBuilder,
...
+	return RunImageConfig{
+		Ref:              ref,
+		BuilderNamespace: builder.GetObjectMeta().GetNamespace(),
+		PullSecretName:   secret,
+	}
+}
+
+type RunImageConfig struct {
+	Ref              string `json:"ref"`
+	BuilderNamespace string `json:"namespace"`
```
matthewmcnew: Why do we need a builder namespace here?

pkg/apis/build/v1alpha1/image_builds.go
ashwin-venkatesh: Oh yeah..we kinda just assumed everything is grabbed from the status and not the spec. We would need to put that information back. Though it "feels" weird for the rebase spec, I think it needs to be there so that the next build has it available to it.

pkg/apis/build/v1alpha1/image_builds.go
ashwin-venkatesh: If it does use the imagePullSecret for the builder, it needs to know the namespace of the builder as that is the namespace where the pull secret will lie

matthewmcnew: At this point we really should only support secrets in the same namespace as the build itself.

pkg/apis/build/v1alpha1/image_builds.go
matthewmcnew: Yeah, that's why I wonder if the build is the one that should `decide` to rebase instead of the image.

Gopkg.lock
```diff
@@ -1347,23 +1347,23 @@
...
   packages = ["."]
-  pruneopts = "NUT"
+  pruneopts = "T"
   revision = "3ca30a56d8a775276f9cdae009ba326fdc05af7f"
   version = "v0.4.0"
 
 [[projects]]
   branch = "master"
-  digest = "1:42674e29bf0cf4662d49bd9528e24b9ecc4895b32d0be281f9cf04d3a7671846"
+  digest = "1:a25f976798a636b58ef030f4d9820b8623e4cb042a5c3480520890f73d144c3d"
```
matthewmcnew: How did so much change here?

pkg/apis/build/v1alpha1/build_types.go
matthewmcnew: You pass not a fully qualified run image, and then it would not be duplicated no?

pkg/apis/build/v1alpha1/build_types.go
```diff
@@ -54,11 +54,18 @@ type BuildSpec struct {
 	CacheName      string                      `json:"cacheName"`
 	Env            []corev1.EnvVar             `json:"env"`
 	Resources      corev1.ResourceRequirements `json:"resources"`
+	Rebase         *RebaseConfig               `json:"rebase,omitempty"`
+}
+
+type RebaseConfig struct {
+	PreviousRunImage RunImageConfig `json:"previousRunImage"`
```
matthewmcnew: Everything that is an image reference should have the `image` key.

pkg/apis/build/v1alpha1/build_types.go
```diff
@@ -54,11 +54,18 @@ type BuildSpec struct {
...
+	LatestRunImage   RunImageConfig `json:"latestRunImage"`
```
matthewmcnew: Everything that is an image reference should have the image key.

pkg/apis/build/v1alpha1/build_types.go
ashwin-venkatesh: That is true..but if we aren;t passing the fully qualified run image, isn't that information duplicated within the builder? Unless we do want to move towards "optionally" treating the run image as something one can set independently on the build? As, I think about it, it does bring in the flexibility much like pack does, but could increase complexity for private run images

pkg/apis/build/v1alpha1/image_builds.go
ashwin-venkatesh: I go a little back and forth on this. The run image feels like a builder concern but I also see the argument of it being a build concern. It lies in that weird grey area, where it is entirely conceivable that the build gets to decide its run image, but when the run image is not explicit, we implicitly rely on the builder secrets to pull it

pkg/cnb/cnb_metadata.go
```diff
@@ -19,11 +20,21 @@ type BuildpackMetadata struct {
 }
 
 type BuilderImageMetadata struct {
-	Buildpacks []BuildpackMetadata `json:"buildpacks"`
+	Buildpacks []BuildpackMetadata    `json:"buildpacks"`
+	Stack      lcyclemd.StackMetadata `json:"stack"`
+}
+
+type Stack struct {
```
joaopapereira: Is this struct used anywhere?

pkg/apis/build/v1alpha1/build.go
```diff
@@ -96,3 +96,7 @@ func (b *Build) BuildEnvVars() []corev1.EnvVar {
 func (b *Build) ImagePullSecretsVolume() corev1.Volume {
 	return b.Spec.Source.Source().ImagePullSecretsVolume()
 }
+
+func (b *Build) Rebasable() bool {
+	return b.Annotations[BuildReasonConfig] == BuildReasonStack
```
joaopapereira: Where do we set this Annotation?

pkg/apis/build/v1alpha1/builder.go
```diff
@@ -16,12 +17,8 @@ func (b *Builder) BuildBuilderSpec() BuildBuilderSpec {
 	}
 }
 
-func (b *Builder) ImagePullSecrets() []string {
-	var secrets []string
-	for _, s := range b.Spec.ImagePullSecrets {
-		secrets = append(secrets, s.Name)
-	}
-	return secrets
+func (b *Builder) ImagePullSecrets() []v1.LocalObjectReference {
```
joaopapereira: Nice 👍

pkg/cnb/image_rebaser.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/pivotal/kpack/pkg/registry"
+)
+
+type ImageRebaser struct {
+	RemoteImageFactory RemoteImageUtilFactory
+	Rebaser            lifecycle.Rebaser
+}
+
+func (f *ImageRebaser) Rebase(build *v1alpha1.Build) (BuiltImage, error) {
+
```
joaopapereira:
```suggestion
```

pkg/reconciler/v1alpha1/build/build.go
```diff
@@ -86,25 +88,40 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 		return nil
 	}
 
-	pod, err := c.reconcileBuildPod(build)
-	if err != nil {
-		return err
-	}
+	if build.Rebasable() {
+		//todo how to handle errors
```
joaopapereira: What does this mean?

cmd/controller/main.go
```diff
@@ -86,10 +88,21 @@ func main() {
...
 	metadataRetriever := &cnb.RemoteMetadataRetriever{
-		RemoteImageFactory: &registry.ImageFactory{
-			KeychainFactory: secret.NewSecretKeychainFactory(k8sClient),
-		},
+		RemoteImageFactory: imageFactory,
+	}
+
+	rebaser := cnb.ImageRebaser{
+		RemoteImageFactory: imageUtilFactory,
+		Rebaser:            lifecycle.Rebaser{Logger: logging.New(os.Stderr)},
```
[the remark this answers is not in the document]
ashwin-venkatesh: Not sure what to do here. The lifecycle expects a lifecycle logger and not a sugared logger. This will write to the controller logs. I am not certain what we want. We should talk about what we wanna do in this situation.

pkg/cnb/cnb_metadata.go
ashwin-venkatesh: Good catch

pkg/apis/build/v1alpha1/build.go
ashwin-venkatesh: https://github.com/pivotal/kpack/pull/140/files#diff-93376212cc28d900cfd5f6ab34de23d0L26-R63

pkg/reconciler/v1alpha1/build/build.go
ashwin-venkatesh: great question. I think we want to figure out if we want to try to rebase again in case it fails, as the reconciler will try to perform the operation again. Do we want to skip the rebase or not is the question in this situation

cmd/controller/main.go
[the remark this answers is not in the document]
matthewmcnew: +1 to wrapping our zap logger.

pkg/cnb/image_rebaser_test.go
```diff
@@ -0,0 +1,143 @@
...
+import (
+	"io/ioutil"
+	"testing"
+
+	"github.com/buildpack/imgutil/fakes"
+	"github.com/buildpack/lifecycle"
+	"github.com/buildpack/lifecycle/logging"
+	"github.com/sclevine/spec"
+	"github.com/stretchr/testify/require"
+	"github.com/tj/assert"
```
matthewmcnew: Did you mean to use this assert?

pkg/apis/build/v1alpha1/build_types.go
```diff
@@ -54,11 +54,13 @@ type BuildSpec struct {
 	CacheName      string                      `json:"cacheName"`
 	Env            []corev1.EnvVar             `json:"env"`
 	Resources      corev1.ResourceRequirements `json:"resources"`
+	PreviousImage  string                      `json:"previousImage"`
```
matthewmcnew: A convention we are following is to always give an image the key image in an objects spec. 

Could we do something like:
```
lastBuild:
      image: 
```

pkg/apis/build/v1alpha1/builder_resource.go
```diff
@@ -1,14 +1,16 @@
...
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 type BuilderResource interface {
 	metav1.ObjectMetaAccessor
 	BuildBuilderSpec() BuildBuilderSpec
 	Image() string
-	ImagePullSecrets() []string
+	ImagePullSecrets() []v1.LocalObjectReference
```
matthewmcnew: 🏆

pkg/cnb/image_rebaser.go
```diff
@@ -0,0 +1,120 @@
...
+func (r remoteImageWrapper) Label(labelName string) (string, error) {
+	return r.remoteImgUtilImage.Label(labelName)
+}
+
+func (r remoteImageWrapper) Env(key string) (string, error) {
+	return r.remoteImgUtilImage.Env(key)
+}
+
+type runImageRef struct { //todo image ref helpers
```
matthewmcnew: I think we don't need this?

pkg/cnb/image_rebaser.go
```diff
@@ -0,0 +1,120 @@
...
+		Namespace:      build.Namespace,
+	})
+	if err != nil {
+		return BuiltImage{}, err
+	}
+
+	return readBuiltImage(remoteImageWrapper{rebasedImage})
+}
+
+type remoteImageWrapper struct {
```
matthewmcnew: This is really clever. I ❤ it.

pkg/cnb/image_rebaser_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+type fakeImageIdentifier struct {
+	identifier string
+}
+
+func (f *fakeImageIdentifier) String() string {
+	return f.identifier
+}
+
+type fakeImageRef struct {
```
matthewmcnew: Probably don't need this?

pkg/apis/build/v1alpha1/build_types.go
ashwin-venkatesh: I like the idea. I wasn't a big fan of calling it "PreviousImage" to begin with

pkg/cnb/image_rebaser_test.go
ashwin-venkatesh: good catch. to answer the question, hell no

pkg/cnb/image_rebaser_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+func testImageRebaser(t *testing.T, when spec.G, it spec.S) {
+	var (
+		imgRebaser             cnb.ImageRebaser
+		fakeRemoteImageFactory = &cnbfakes.FakeRemoteImageUtilFactory{}
+	)
+
+	when("#Rebase", func() {
+		additionalTags := []string{"other-tag"}
+		it("invokes Rebase on the previously built image", func() {
```
matthewmcnew: I like the decision to not mock out the lifecycle.

pkg/registry/ggcr_image.go
```diff
@@ -16,6 +16,7 @@ import (
 type GoContainerRegistryImage struct {
 	image    v1.Image
 	repoName string
+	keychain authn.Keychain
```
matthewmcnew: Probably no longer needed here?

pkg/cnb/image_rebaser.go
ashwin-venkatesh: The genius that is @djoyahoy

pkg/reconciler/v1alpha1/image/image_test.go
```diff
@@ -1413,6 +1420,7 @@ func builds(image *v1alpha1.Image, sourceResolver *v1alpha1.SourceResolver, coun
 			},
 			Status: v1alpha1.BuildStatus{
 				LatestImage: fmt.Sprintf("%s@sha256:build-%d", image.Spec.Tag, i),
+				RunImage:    "some/run@sha256acf123123", //maybe?
```
matthewmcnew: Probably take out the maybe, here?
Might make sense to use constants?

matthewmcnew: lol because I definitely wrote that maybe.

pkg/reconciler/v1alpha1/build/build.go
```diff
@@ -86,25 +88,40 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
 
-	if build.MetadataReady(pod) {
-		image, err := c.MetadataRetriever.GetBuiltImage(build)
+		image, err := c.ImageRebaser.Rebase(build)
 		if err != nil {
 			return err
```
matthewmcnew: Maybe use a NewPermanentError to prevent it from being re-enqued?

pkg/reconciler/v1alpha1/build/build.go
```diff
@@ -86,25 +88,40 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
-	build.Status.StepsCompleted = stepCompleted(pod)
-	build.Status.Conditions = conditionForPod(pod)
+			build.Status.BuildMetadata = buildMetadataFromBuiltImage(image)
+			build.Status.LatestImage = image.Identifier
+			build.Status.RunImage = image.RunImage
+		}
+
+		build.Status.PodName = pod.Name
+		build.Status.StepStates = stepStates(pod)
+		build.Status.StepsCompleted = stepCompleted(pod)
```
matthewmcnew: Does it make sense to have step states for rebase?

ashwin-venkatesh: Indentation makes it look weird, but the rebase flow skips those lines.

pkg/reconciler/v1alpha1/build/build.go
```diff
@@ -86,25 +88,40 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
-	pod, err := c.reconcileBuildPod(build)
-	if err != nil {
-		return err
-	}
+	if build.Rebasable() {
+		//todo how to handle errors
 
-	if build.MetadataReady(pod) {
-		image, err := c.MetadataRetriever.GetBuiltImage(build)
+		image, err := c.ImageRebaser.Rebase(build)
```
matthewmcnew: We need to put ConditionSucceeded & ConditionFailed so that image reconciler picks this up.

pkg/reconciler/v1alpha1/build/build.go
ashwin-venkatesh: 🙏

pkg/secret/k8s_keychain.go
```diff
@@ -37,6 +38,14 @@ func (f *k8sSecretKeychainFactory) KeychainForSecretRef(ref registry.SecretRef)
 	return authn.NewMultiKeychain(annotatedBasicAuthKeychain, k8sKeychain), nil
 }
 
+func toStringPullSecrets(secrets []v1.LocalObjectReference) []string {
```
matthewmcnew: So smooth. 👍

pkg/reconciler/v1alpha1/build/build.go
ashwin-venkatesh: Good catch

pkg/reconciler/v1alpha1/build/build.go
matthewmcnew: Yeah, I was more asking if we _should_ have step states on the rebase build?

It wouldn't need to be in this PR.

matthewmcnew:
As a future PR could we use previousImage as the input to regular builds?

ashwin-venkatesh:
> As a future PR could we use previousImage as the input to regular builds?

I like the idea..by that you mean lastBuild.image right?? 😛

pkg/reconciler/v1alpha1/build/build.go
ashwin-venkatesh: It makes sense if/when it happens in its own pod. That is something to think about though.

pkg/apis/build/v1alpha1/build_types.go
ashwin-venkatesh: Also, this will allow us to throw in more metadata here if that might help in follow up builds from an optimization perspective.

matthewmcnew:
> > As a future PR could we use previousImage as the input to regular builds?
> 
> I like the idea..by that you mean lastBuild.image right?? 😛

Haha. Yes.

PR #152, opened 2019-09-30
Add LastTransitionTime to conditions

joaopapereira:
When working with kpack it is hard to understand when was the last time a condition was updated.
This PR adds the time to the condition


ashwin-venkatesh:
Did we verify if this does not force multiple reconciles because of the Time.Now?

joaopapereira:
> Did we verify if this does not force multiple reconciles because of the Time.Now?

I had no time to verify it

joaopapereira:
I just tested and the date does not matter

PR #161, opened 2019-10-02
Use go git

matthewmcnew:
Use go-git to fetch git source builds

- combine source-init,creds-init,build-init
- Mount volumes as CNB user to avoid needing to chown
- Run Build Init as CNB user
- Move k8sSecretKeychainFactory -> dockercreds package


pkg/git/fetch.go
```diff
@@ -0,0 +1,65 @@
...
+		URLs: []string{gitURL},
+	})
+	if err != nil {
+		return errors.Wrap(err, "unable to create remote")
+	}
+
+	opts := &git.FetchOptions{
+		RefSpecs: []config.RefSpec{"refs/*:refs/*"},
+		Auth:     resolvedAuth,
+		Depth:    0,
```
matthewmcnew: Of note, For simplicity in this PR: I removed the depth=1 checkout from our current source-init. It would make sense in a future issue to optimize this checkout.

pkg/apis/build/v1alpha1/build.go
```diff
@@ -89,7 +89,7 @@ func (b *Build) Finished() bool {
 	return !b.Status.GetCondition(duckv1alpha1.ConditionSucceeded).IsUnknown()
 }
 
-func (b *Build) BuildEnvVars() []corev1.EnvVar {
+func (b *Build) SourceEnvVars() []corev1.EnvVar {
```
djoyahoy: This name is a little confusing, but I understand why you want to change it. Unfortunately, I don't have a suggestion to make.

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -116,70 +119,39 @@ func (b *Build) BuildPod(config BuildPodConfig, secrets []corev1.Secret, builder
...
-					ImagePullPolicy: corev1.PullIfNotPresent,
-					WorkingDir:      "/workspace",
-					VolumeMounts: []corev1.VolumeMount{
-						imagePullSecretsVolume,
-						sourceVolume,
-						homeVolume,
-					},
-				},
 				{
 					Name:  "prepare",
```
djoyahoy: Should we rename this step to `build-init` so that it matches up with the binary name?

pkg/git/git_keychain.go
```diff
@@ -0,0 +1,60 @@
...
+			if err != nil {
+				return nil, err
+			}
+			return &http.BasicAuth{
+				Username: basicAuth.Username,
+				Password: basicAuth.Password,
+			}, nil
+		}
+	}
+	return anynomousAuth, nil
```
djoyahoy: typo: should be anonymousAuth

pkg/git/k8s_git_keychain.go
```diff
@@ -15,6 +17,8 @@ type k8sGitKeychain struct {
 	secretManager secret.SecretManager
 }
 
+var anynomousAuth transport.AuthMethod = nil
```
djoyahoy: typo: anonymous

pkg/apis/build/v1alpha1/build.go
joaopapereira: BuildtimeEnvVariables

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -11,6 +11,8 @@ import (
 )
 
 const (
+	buildInitBinary = "/layers/org.cloudfoundry.go-mod/app-binary/build-init"
```
joaopapereira: Maybe we should add a comment here saying that when the go buildpack changes this should be changed as well

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -116,70 +119,39 @@ func (b *Build) BuildPod(config BuildPodConfig, secrets []corev1.Secret, builder
 					Resources:       b.Spec.Resources,
 				},
 			},
+			SecurityContext: &corev1.PodSecurityContext{
```
joaopapereira: Why do we need this change?

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -116,70 +119,39 @@ func (b *Build) BuildPod(config BuildPodConfig, secrets []corev1.Secret, builder
...
-						homeVolume,
-					},
-				},
 				{
 					Name:  "prepare",
 					Image: config.BuildInitImage,
 					SecurityContext: &corev1.SecurityContext{
-						RunAsUser:  &root,
-						RunAsGroup: &root,
+						RunAsUser:  &userAndGroup.Uid,
```
joaopapereira: We no longer need root to do this step?

pkg/dockercreds/k8s_keychain_test.go
```diff
@@ -1,4 +1,4 @@
-package secret_test
+package dockercreds
```
joaopapereira: Are we testing anything internal in this package? if not can we change the package to `dockercreds_test`?

pkg/dockercreds/parse_annoted_secrets_test.go
```diff
@@ -0,0 +1,66 @@
+package dockercreds
```
joaopapereira: Are we testing anything internal in this package? if not can we change the package to `dockercreds_test`?

pkg/dockercreds/parse_annoted_secrets_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"testing"
+
+	"github.com/sclevine/spec"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func TestParseAnnotatedSecrets(t *testing.T) {
+	spec.Run(t, "Test Parse Annotated Secrets", test)
```
joaopapereira:
```suggestion
	spec.Run(t, "Test Parse Annotated Secrets", testParseAnnotatedSecrets)
```

pkg/dockercreds/parse_annoted_secrets_test.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func test(t *testing.T, when spec.G, it spec.S) {
```
joaopapereira:
```suggestion
func testParseAnnotatedSecrets(t *testing.T, when spec.G, it spec.S) {
```

pkg/secret/volume_secret_reader.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+func ReadSecret(secretVolume, secretName string) (BasicAuth, error) { //test this
```
joaopapereira: Is this comment still true?

matthewmcnew: 🙂

pkg/dockercreds/parse_annoted_secrets_test.go
matthewmcnew: 😄

pkg/dockercreds/parse_annoted_secrets_test.go
matthewmcnew: 😄

pkg/git/git_keychain.go
matthewmcnew: If you spell it right is really anonymous?

pkg/apis/build/v1alpha1/build.go
matthewmcnew: I changed this because working on this kept confusing me. 
It is not really a Build env var because it is just the input to setup of the configured build source.

matthewmcnew: BuildtimeEnvVariables would imply it is the build.spec.env?

pkg/apis/build/v1alpha1/build_pod.go
matthewmcnew: This allows us to configure the owner of the volume mounts and prevents us from needing to `chown` the volumes in the prepare step.

pkg/apis/build/v1alpha1/build_pod.go
matthewmcnew: Yes, We can use the builder's user/group to run the build init. 
This allows the source and platform directories to be owned by the user that runs in the builder without a need for a `chown`.

pkg/dockercreds/parse_annoted_secrets_test.go
matthewmcnew: Yep let's move this to dockercreds_test package.

pkg/dockercreds/k8s_keychain_test.go
matthewmcnew: We are not currently exporting the k8sGitKeychain. I would like to keep this in the dockercreds package.

pkg/apis/build/v1alpha1/build.go
joaopapereira: But this is what these variables are correct?

matthewmcnew: No, These are the env vars for app source. 

Like GIT_URL & GIT_REVISON.

djoyahoy: Oh in that case this change makes sense. I thought these were build time env vars as well.

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -11,6 +11,8 @@ import (
 )
 
 const (
+	buildInitBinary = "/layers/org.cloudfoundry.go-mod/app-binary/build-init" // Can be changed to build-init in https://github.com/cloudfoundry/go-mod-cnb/issues/8
```
ashwin-venkatesh: Cant wait for the go buildpack to start putting the binaries on the path!!!!

PR #165, opened 2019-10-07
rebase comapares run images using the identifier only

### LINKED ISSUE #164: Kpack parses run images from registries with a port incorrectly
To reproduce, create a builder with a run image that lives in a registry with a port like: `registry.default.svc.cluster.local:5000/run`. 

The builds will record the wrong run image causing a stack build loop. 

The extra ':' confuses kpack.

ashwin-venkatesh:
closes #164

### DIFF
--- a/pkg/apis/build/v1alpha1/image_builds.go
+++ b/pkg/apis/build/v1alpha1/image_builds.go
@@ -24,21 +24,21 @@ const (
 	BuildReasonStack      = "STACK"
 )
 
-func (im *Image) buildNeeded(lastBuild *Build, sourceResolver *SourceResolver, builder BuilderResource) ([]string, bool) {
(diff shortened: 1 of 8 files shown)


pkg/apis/build/v1alpha1/image_builds.go
```diff
@@ -57,11 +57,23 @@ func (im *Image) buildNeeded(lastBuild *Build, sourceResolver *SourceResolver, b
 		reasons = append(reasons, BuildReasonBuildpack)
 	}
 
-	if lastBuild.Status.RunImage != "" && lastBuild.Status.RunImage != builder.RunImage() {
-		reasons = append(reasons, BuildReasonStack)
+	if lastBuild.Status.RunImage != "" {
```
matthewmcnew: It would be great to move much of this logic to the build itself in a future refactor. 

Something like: 

```
lastBuild.MatchesStack(builder)
```

PR #169, opened 2019-10-11
Webhook to add default service account to builds

### LINKED ISSUE #74: Default to `default` service account
When an image resource is created and `spec.ServiceAccount` is not specified it should use the [default service account in the namespace](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-the-default-service-account-to-access-the-api-server). 

This allow you to configure the secrets for a namespace and easily create images within that namespace.

joaopapereira:
Closes #74

### DIFF
--- a/cmd/build-defaults-webhook/main.go
+++ b/cmd/build-defaults-webhook/main.go
@@ -1,0 +1,37 @@
+package main
+
+import (
+	"log"
+	"net/http"
+	"path/filepath"
+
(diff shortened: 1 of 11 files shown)


matthewmcnew:
We should utilize the Defaultable & Validatable interfaces to setup the webhook server. Knative/pkg or kubebuilder can then help simplify the setup so that we don't need to roll our own mutate endpoints and tls configuration.

PR #171, opened 2019-10-14
Enable credential provider

joaopapereira:
This PR as 2 distinct commits:
1 - Will enable knative to display our errors on the builder and cluster builder reconcilers
2 - Enable extra registry credentials provider. The commit message provide more details

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/census-instrumentation/opencensus-proto v0.2.1 // indirect
 	github.com/evanphx/json-patch v4.5.0+incompatible // indirect
 	github.com/gogo/protobuf v1.3.0 // indirect
+	github.com/google/btree v1.0.0 // indirect
 	github.com/google/go-cmp v0.3.1
 	github.com/google/go-containerregistry v0.0.0-20190910142231-b02d448a3705
 	github.com/googleapis/gnostic v0.3.1 // indirect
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -52,7 +53,7 @@ require (
 	k8s.io/gengo v0.0.0-20190822140433-26a664648505 // indirect
 	k8s.io/klog v1.0.0 // indirect
 	k8s.io/kube-openapi v0.0.0-20190816220812-743ec37842bf // indirect
-	k8s.io/kubernetes v1.10.2 // indirect
+	k8s.io/kubernetes v1.10.2
```
matthewmcnew: Another indirect bites the dust.

pkg/dockercreds/k8svolume/provider.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (k VolumeSecretProvider) Enabled() bool {
+	_, ok := os.LookupEnv(SecretFilePathEnv)
+	return ok
+}
+
+func (k VolumeSecretProvider) Provide() credentialprovider.DockerConfig {
+	secretFolder := os.Getenv(SecretFilePathEnv)
+
+	dockerConfig, err := parseDockerConfigJson(filepath.Join(secretFolder, ".dockerconfigjson"))
```
matthewmcnew: Is there a reason not to use `dockercreds.ParseDockerPullSecrets`?

joaopapereira: I decided not to use `ParseDockerPullSecrets` because we talked about supporting docker config in json only and the return types of the functions do not match, so we would have to create an adaptation layer.
We can create a different PR to standardize our types and start using the structs provided by k8s instead of handrolling `DockerCreds` or `dockerConfigJson`

matthewmcnew: Ok. I don't know if would necessarily make sense to rely on k8s.DockerConfigEntry as it just has BasicAuth username/password keys.

PR #174, opened 2019-10-15
Use archtest to avoid unintentional dependencies

matthewmcnew:


pkg/cnb/image_rebaser_test.go
```diff
@@ -131,7 +131,7 @@ type newRemoteArgs struct {
 func (f *FakeRemoteImageUtilFactory) newRemote(imageName string, baseImage string, secretRef registry.SecretRef) (imgutil.Image, error) {
 	for i, args := range f.args {
 		remoteArgs := newRemoteArgs{ImageName: imageName, BaseImage: baseImage, SecretRef: secretRef}
-		if cmp.DeepEqual(args, remoteArgs)().Success() {
+		if diff := cmp.Diff(args, remoteArgs); diff == "" {
```
ashwin-venkatesh: You were right about `cmp` returning a string afterall

pkg/arch_test.go
```diff
@@ -0,0 +1,13 @@
+package pkg
+
+import (
+	"testing"
+
+	"github.com/matthewmcnew/archtest"
+)
+
+func TestDependencies(t *testing.T) {
```
ashwin-venkatesh: does this impact test times in a significant way?

matthewmcnew: It will only scan direct dependencies so the test time increase is manageable.

joaopapereira: Looks like the testing time, at least in actions, did not change

PR #175, opened 2019-10-15
Rebuild when stack changes

### LINKED ISSUE #162: Ensure that kpack does a Rebuild and not a Rebase when the run-image/builder stack has changed
When the  ONLY the run image and builder stack change, kpack should perform a rebuild and not a rebase.

ashwin-venkatesh:
- Created a struct for the full stack information (ID+Run Image)
- Add new reason for rebuild in case of a stack change

Closes #162


pkg/apis/build/v1alpha1/image_builds.go
```diff
@@ -17,11 +17,12 @@ const (
...
-	BuildReasonConfig     = "CONFIG"
-	BuildReasonCommit     = "COMMIT"
-	BuildReasonBuildpack  = "BUILDPACK"
-	BuildReasonStack      = "STACK"
+	BuildReasonAnnotation  = "image.build.pivotal.io/reason"
+	BuildReasonConfig      = "CONFIG"
+	BuildReasonCommit      = "COMMIT"
+	BuildReasonBuildpack   = "BUILDPACK"
+	BuildReasonStack       = "STACK"
+	BuildReasonStackUpdate = "STACK-UPDATE"
```
matthewmcnew: This should still be STACK for both reasons.

pkg/apis/build/v1alpha1/image_builds.go
```diff
@@ -57,8 +58,15 @@ func (im *Image) buildNeeded(lastBuild *Build, sourceResolver *SourceResolver, b
 		reasons = append(reasons, BuildReasonBuildpack)
 	}
 
-	if lastBuild.Status.RunImage != "" {
-		lastBuildRunImageRef, err := name.ParseReference(lastBuild.Status.RunImage)
+	if lastBuild.Status.Stack.ID != "" {
```
matthewmcnew: Ideally, buildNeeded should not change if the build is able to "make the optimization" decision.

ashwin-venkatesh: That is true. The run image SHAs will definitely be different.

PR #176, opened 2019-10-16
Refactor rebase to happen in a pod

djoyahoy:
- Add rebase image and rebase binary
- Remove rebase logic from build reconciler


matthewmcnew:
To capture the performance implications of this refactor: A non perfect study of times for kpack to rebase 50 images showed a non-negligible improvement of running this in a pod. 

**Cloud**
_Pod_
79 seconds
99 seconds
97 seconds

_Controller_
171 seconds
165 seconds
165 seconds

**Minikube**
_Pod_
64 seconds
61 seconds
61 seconds

_Controller_
107 seconds
108 seconds
107 seconds

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/census-instrumentation/opencensus-proto v0.2.1 // indirect
 	github.com/evanphx/json-patch v4.5.0+incompatible // indirect
 	github.com/gogo/protobuf v1.3.0 // indirect
+	github.com/google/btree v1.0.0 // indirect
```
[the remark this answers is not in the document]
ashwin-venkatesh: This always manages to make it back in

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -28,14 +29,18 @@ const (
...
+type BuildPodImages struct {
 	BuildInitImage string
 	NopImage       string
+	RebaseImage    string
 }
 
-type UserAndGroup struct {
-	Uid int64
-	Gid int64
+type BuildPodBuilderConfig struct {
```
ashwin-venkatesh: ❤️ it

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -28,14 +29,18 @@ const (
 	builderPullSecretsDirName = "builder-pull-secrets-dir"
 )
 
-type BuildPodConfig struct {
+type BuildPodImages struct {
```
ashwin-venkatesh: This is definitely a better name for the struct

pkg/cnb/cnb_metadata.go
```diff
@@ -12,7 +12,10 @@ import (
 	"github.com/pivotal/kpack/pkg/registry"
 )
 
-const BuilderMetadataLabel = "io.buildpacks.builder.metadata"
+const (
+	BuilderMetadataLabel = "io.buildpacks.builder.metadata"
+	StackLabel           = "io.buildpacks.stack.id"
```
ashwin-venkatesh: We should probably import this labels from the lifecycle so that we don't run into problems if it changes. I don't think it really will, but it seems like a safer way to use it

pkg/reconciler/v1alpha1/build/build.go
```diff
@@ -88,56 +86,26 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 		return nil
 	}
 
-	if build.Rebasable() {
-		image, err := c.ImageRebaser.Rebase(build, ctx)
+	pod, err := c.reconcileBuildPod(build)
```
ashwin-venkatesh: ❤️

pkg/cnb/cnb_metadata.go
ashwin-venkatesh: The lifecycle does not have a label for builder metadata tho but I we should use constants from the lifecycle where we can

go.mod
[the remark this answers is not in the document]
djoyahoy: Honestly, at this point I'm treating it as a dependency. I think go mod tidy is over-eager.

PR #178, opened 2019-10-17
Add rebase test to e2e tests

djoyahoy:
- Remove "when" block from e2e tests
- Test rebase by creating a new build with latest image set in the
  spec and the STACK annotation

### DIFF
--- a/test/execute_build_test.go
+++ b/test/execute_build_test.go
@@ -18,7 +18,9 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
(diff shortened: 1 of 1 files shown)


test/execute_build_test.go
```diff
@@ -249,11 +251,43 @@ func imageExists(name string) func() bool {
 		if err != nil {
 			return false
 		}
-
 		return true
 	}
 }
 
+func validateRebase(t *testing.T, clients *clients, imageName, testNamespace string) {
```
joaopapereira: Love it

test/execute_build_test.go
```diff
@@ -70,148 +71,149 @@ func testCreateImage(t *testing.T, when spec.G, it spec.S) {
 		for _, tag := range cfg.generatedImageNames {
 			deleteImageTag(t, tag)
 		}
+		deleteNamespace(t, clients, testNamespace)
```
ashwin-venkatesh: The only reason not to keep this is to ensure we can investigate failing acceptance tests. I hate when the tests fail because of the namespace problem, but I'd rather we dont delete the namespace in the after

djoyahoy: @matthewmcnew I agree. I'll implement a wait.

PR #179, opened 2019-10-17
rebuild images if the stack changes

ashwin-venkatesh:
- Add Stack on the Status of the image, build and the builder resource
- Move RunImage + StackId into struct called Stack


ashwin-venkatesh:
what just happened :|

PR #181, opened 2019-10-17
Build kpack's completion image

matthewmcnew:
- Remove dependency on knative nop
- Refactor hack/(apply/release)

### DIFF
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -10,3 +10,34 @@ function pack_build() {
     docker pull ${image}
     resolved_image_name=$(docker inspect ${image} --format '{{index .RepoDigests 0}}' )
 }
+
+function compile() {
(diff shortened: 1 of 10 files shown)


hack/common.sh
```diff
@@ -5,8 +5,39 @@ function pack_build() {
     target=$2
     builder="cloudfoundry/cnb:bionic"
 
-    pack build ${image} --builder ${builder} -e BP_GO_TARGETS=${target} --publish
+#    pack build ${image} --builder ${builder} -e BP_GO_TARGETS=${target} --publish
```
ashwin-venkatesh: We can delete this line right?

matthewmcnew: Whoops. Good catch 🙂

PR #182, opened 2019-10-18
Correct ClusterBuilder metadata retrieval keychain

ashwin-venkatesh:
- We need to create a Keychain and start to use it as a multi keychain
and read from the provider that we implemented. We might be able to get away with it as it is a corner case that shows up when the registry your images are in is the same GCP account as the cluster.
- Also ggcr is forking the providers and we should update our version of
ggcr when they finish this work so we do not have to depend on k8s.net/kubernetes
package
- Remove function NewRemoteWithDefaultKeychain that is no longer needed


pkg/reconciler/v1alpha1/clusterbuilder/clusterbuilder.go
```diff
@@ -21,8 +21,8 @@ import (
 )
 
 const (
-	ReconcilerName = "Builders"
-	Kind           = "Builder"
+	ReconcilerName = "ClusterBuilders"
```
joaopapereira: Very good catch :D

pkg/registry/image_factory.go
```diff
@@ -19,10 +19,6 @@ func (f *ImageFactory) NewRemote(image string, secretRef SecretRef) (RemoteImage
 	return NewGoContainerRegistryImage(image, keychain)
 }
 
-func (f *ImageFactory) NewRemoteWithDefaultKeychain(image string) (RemoteImage, error) {
```
matthewmcnew: Nice

matthewmcnew:
What problem specifically is this solving? It seems to just preventing the ClusterBuilder from falling back to the `default` service account when fetching metadata in a BuildPod. Did you run into an issue where this was causing problems?

ashwin-venkatesh:
> What problem specifically is this solving? It seems to just prevent the ClusterBuilder from falling back to the `default` service account when fetching metadata for a BuildPod. Did you run into an issue where this was causing problems?

We want to use credential providers to fetch the private images that are in the same registry as the cluster. We ran into problem where it would look from credentials in the service which would cause this to fail. Additionally, we wanted to be more clear about the intent of not using serviceAccount credentials for an image that definitely was not namespaced.

PR #184, opened 2019-10-18
Add validations for Builders and Builds

ashwin-venkatesh:


pkg/apis/build/v1alpha1/build_validation.go
```diff
@@ -1,31 +1,43 @@
...
+	if bbs.Image == "" {
+		return apis.ErrMissingField("name")
+	}
+
+	_, err := name.ParseReference(bbs.Image, name.WeakValidation)
+	if err != nil {
+		return apis.ErrInvalidValue(bbs.Image, "image")
+	}
+
+	switch bbs.Kind {
```
matthewmcnew: Do you need a Builder `kind` for a build?

pkg/apis/build/v1alpha1/builder_validation.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"context"
+
+	"knative.dev/pkg/apis"
+)
+
+func (b *Builder) SetDefaults(ctx context.Context) {
+	if b.Spec.UpdatePolicy == "" {
+		b.Spec.UpdatePolicy = Polling
```
matthewmcnew: This brings me so much joy.

pkg/apis/build/v1alpha1/build_validation.go
ashwin-venkatesh: We do now. It is used to determine the k8schain setup for pulling the builder before we create a buildpod

pkg/apis/build/v1alpha1/builder_validation.go
ashwin-venkatesh: There is probably one other validation I would like to do here, is verify that the builder image provided is a valid image reference.

pkg/apis/build/v1alpha1/build_validation.go
matthewmcnew: I am concerned that this makes it more difficult to use Builds on their own. 

It looks like the build will default to the namespace if the ClusterKind isn't specified. https://github.com/pivotal/kpack/blob/208f6940de04a30f7251407bec667fd8267bbb80/pkg/buildpod/generator.go#L62

Can we make Kind optional here or default it to 'Builder'?

pkg/dockercreds/volume_secret_keychain.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+const (
+	SecretFilePathEnv = "CREDENTIAL_PROVIDER_SECRET_PATH"
+)
+
+type VolumeSecretKeychain struct {
+}
+
+func (k *VolumeSecretKeychain) Resolve(r authn.Resource) (authn.Authenticator, error) {
```
matthewmcnew: Can we set this up so it doesn't always need to read the local file?

pkg/dockercreds/volume_secret_keychain_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/sclevine/spec"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/pivotal/kpack/pkg/dockercreds"
+)
+
+func TestProvider(t *testing.T) {
```
matthewmcnew: Rename this to testVolumeSecretKeychain?

PR #186, opened 2019-10-22
Build fails early if it does not have read access to the run image

ashwin-venkatesh:


pkg/dockercreds/access_checker_test.go
```diff
@@ -118,6 +118,45 @@ func testHasWriteAccess(t *testing.T, when spec.G, it spec.S) {
...
+
+			handler.HandleFunc("/v2/some/image/manifests/tag", func(writer http.ResponseWriter, request *http.Request) {
+				writer.WriteHeader(401)
+			})
+
+			canRead, err := HasReadAccess(testKeychain{}, tagName)
+			require.Error(t, err)
+			assert.False(t, canRead)
+		})
+		it("returns false when we cannot reach the server", func() {
```
djoyahoy:
```suggestion

		it("returns false when we cannot reach the server", func() {
```

PR #189, opened 2019-10-23
Update lifecycle to version 0.5.0

joaopapereira:


cmd/rebase/main.go
```diff
@@ -31,7 +33,12 @@ func init() {
 func main() {
 	flag.Parse()
 
-	logger := log.New(os.Stdout, "rebase:", log.Lshortfile)
+	logger := &log.Logger{
```
matthewmcnew: Can we just import Logger from "github.com/buildpack/lifecycle/cmd"?

PR #218, opened 2019-11-12
Fix matching on registries to be consistent across various annotations

ashwin-venkatesh:


pkg/dockercreds/k8sdockercreds/k8s_keychain.go
```diff
@@ -79,13 +79,10 @@ func (k *annotatedBasicAuthKeychain) Resolve(res authn.Resource) (authn.Authenti
 	sort.Slice(secrets, func(i, j int) bool { return secrets[i].Name < secrets[j].Name })
 
 	for _, s := range secrets {
-		if match(s, res) {
+		matcher := dockercreds.RegistryMatcher{Registry: s.Annotations[v1alpha1.DOCKERSecretAnnotationPrefix]}
```
matthewmcnew: Do we still have the `v1.SecretTypeBasicAuth` check on docker basic auth secrets?

ashwin-venkatesh: Good catch..will fix and merge

PR #220, opened 2019-11-15
Create experimental CustomBuilder resource

matthewmcnew:
- Add experimental.kpack.pivotal.io api group


pkg/reconciler/v1alpha1/custombuilder/custombuilder.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	customBuilder = customBuilder.DeepCopy()
+
+	builderRecord, creationError := c.reconcileCustomBuilder(customBuilder)
+	if creationError != nil {
+		customBuilder.ErrorCreate(creationError)
+
+		err := c.updateStatus(customBuilder)
+		if err != nil {
+			return nil
```
matthewmcnew: Return `err` here.

pkg/reconciler/v1alpha1/custombuilder/custombuilder.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+
+		return controller.NewPermanentError(creationError)
+	}
+
+	customBuilder.Status.ObservedGeneration = customBuilder.Generation
```
matthewmcnew: We should update the generation in the err case as well.

pkg/cnb/cnb_metadata.go
```diff
@@ -118,15 +108,12 @@ type BuiltImage struct {
 }
 
 func readBuiltImage(img registry.RemoteImage) (BuiltImage, error) {
-	var buildMetadataJSON string
```
matthewmcnew: Nice!

pkg/cnb/buildpackage_store.go
```diff
@@ -0,0 +1,135 @@
...
+			}
+
+			buildpackLayers = append(buildpackLayers, buildpackInfo.Layers...)
+		}
+
+	}
+	return buildpackLayers, nil
+}
+
+func kpackmountableLayer(ref name.Reference, digest, diffId string) v1.Layer {
```
matthewmcnew: This is all unused at the moment

cmd/controller/main.go
```diff
@@ -80,6 +81,7 @@ func main() {
 	builderInformer := informerFactory.Build().V1alpha1().Builders()
 	clusterBuilderInformer := informerFactory.Build().V1alpha1().ClusterBuilders()
 	sourceResolverInformer := informerFactory.Build().V1alpha1().SourceResolvers()
+	custromBuilderInformer := informerFactory.Experimental().V1alpha1().CustomBuilders()
```
matthewmcnew:
```suggestion
	customBuilderInformer := informerFactory.Experimental().V1alpha1().CustomBuilders()
```

pkg/apis/experimental/v1alpha1/custom_builder.go
```diff
@@ -0,0 +1,39 @@
...
+
+func (b *CustomBuilder) BuildBuilderSpec() v1alpha1.BuildBuilderSpec {
+	return v1alpha1.BuildBuilderSpec{
+		Image:            b.Status.LatestImage,
+	}
+}
+
+// FIXME : is this correct?
+func (b *CustomBuilder) Image() string {
+	return b.Spec.Tag
```
matthewmcnew: Hmm. This is only used in the metadata retriever. I wonder if we can make it so that the `BuilderResource` does not have this method. Instead the metadata retriever can use an interface that embeds and extends the BuilderResource?

pkg/cnb/builder_builder.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func newBuilderBuilder(keychain authn.Keychain, client Client, baseImage v1.Image) (*BuilderBuilder, error) {
+	baseMetadata := &BuilderImageMetadata{}
+	err := registry.GetLabel(baseImage, buildpackMetadataLabel, baseMetadata)
+	if err != nil {
+		return nil, err
+	}
+
+	runImageRef, err := name.ParseReference(baseMetadata.Stack.RunImage.Image, name.WeakValidation)
```
matthewmcnew: The parsing of the run image inside of the builder_builder feels a bit odd to me. All the other 'parts' of a builder get passed into the BuilderBuilder (layers and baseImage). I imagine in the future we will have even more methods that 'add' parts to the run image. `builderBuilder.AddLifecycle` or  `builderBuilder.AddStack` or `builderBuilder.SetRunImage`. 

And the actual resolved run image digest is not used by the builder_builder itself. It is just exposed via the stack. Could the `RemoteBuilderCreator` parse the fully qualified run image and use it to construct the `BuilderRecord`?

pkg/reconciler/v1alpha1/custombuilder/custombuilder.go
```diff
@@ -0,0 +1,106 @@
...
+	customBuilder.Status.BuilderStatus(builderRecord)
+	return c.updateStatus(customBuilder)
+}
+
+func (c *Reconciler) reconcileCustomBuilder(customBuilder *experimentalV1alpha1.CustomBuilder) (v1alpha1.BuilderRecord, error) {
+	keychain, err := c.KeychainFactory.KeychainForSecretRef(registry.SecretRef{
+		ServiceAccount: customBuilder.Spec.ServiceAccount,
+		Namespace:      customBuilder.Namespace,
```
matthewmcnew: Because we have had regressions here in the past. Could we test the secretRef we use?

pkg/cnb/create_builder_test.go
```diff
@@ -0,0 +1,394 @@
...
+	"testing"
+	"time"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/random"
+	"github.com/pkg/errors"
+	"github.com/sclevine/spec"
+	"github.com/stretchr/testify/require"
+	"github.com/tj/assert"
```
matthewmcnew: Whoops! Can we migrate that to testify?

pkg/cnb/create_builder.go
```diff
@@ -0,0 +1,74 @@
...
+	MakeStore(keychain authn.Keychain, storeImage string) (Store, error)
+}
+
+type Store interface {
+	FetchBuildpack(id, version string) (RemoteBuildpackInfo, error)
+}
+
+type RemoteBuilderCreator struct {
+	RemoteImageClient Client
+	StoreFactory      StoreFactory
```
matthewmcnew: Out of curiosity, what lead to migrating this to use an interface vs a func signature?

pkg/registry/client_test.go
```diff
@@ -0,0 +1,55 @@
...
+	spec.Run(t, "TestClient", testClient)
+}
+
+func testClient(t *testing.T, when spec.G, it spec.S) {
+	handler := http.NewServeMux()
+	server := httptest.NewServer(handler)
+	tagName := fmt.Sprintf("%s/some/image:tag", server.URL[7:])
+
+	when("Save", func() {
+		it("does not save images if exisiting image already exisits", func() {
```
matthewmcnew: Is it worth a simple save when image does not exist test?

samples/custom_builder.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  name: my-custom-builder
+  namespace: default
+spec:
+  serviceAccount: default
+  tag: sample/custom-builder
+  stack:
+    baseBuilderImage: cloudfoundry/cnb:bionic
+  store:
+    image: cloudfoundry/cnb:bionic
```
matthewmcnew: I wonder if we should not even have CustomBuilder samples until we stabilized the api and can point to released pack tooling.

djoyahoy: This was useful for smoke testing. We could move it to something like samples/experimental.

pkg/cnb/create_builder.go
djoyahoy: I just preferred giving it a name rather than leaving it as an anonymous function definition.

pkg/cnb/builder_builder.go
djoyahoy: Agreed. It felt weird to do it here. We can migrate it out.

pkg/apis/experimental/v1alpha1/custom_builder.go
djoyahoy: We can do this, curious why a custom builder doesn't need to have an Image conceptually though. Even if we don't use that method, it seems to fulfill that interface.

matthewmcnew: It certainly could implement the interface. (Probably return `b.Status.LatestImage`?)

I think the value in not including the `Image` method would be in being explicit about our intentions and not writing methods we don't need. A little interface-segregation principle.

djoyahoy: I buy that. I'm going to think a bit about the BuilderResource interface in general.

PR #223, opened 2019-11-15
omitempty possible empty fields

matthewmcnew:
-  This prevents kpack from serializing unnecessary fields creating
a less brittle api
- Allow build.spec.LastBuild to be nil
- Allow image.spec.build to be nil

### DIFF
--- a/pkg/apis/build/v1alpha1/build_test.go
+++ b/pkg/apis/build/v1alpha1/build_test.go
@@ -1,0 +1,47 @@
+package v1alpha1_test
+
+import (
+	"testing"
+
+	"github.com/pivotal/kpack/pkg/apis/build/v1alpha1"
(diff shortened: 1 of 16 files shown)


pkg/apis/build/v1alpha1/build_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"testing"
+
+	"github.com/pivotal/kpack/pkg/apis/build/v1alpha1"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestRebaseable(t *testing.T) {
```
ashwin-venkatesh: Any chance this can be split into 3 spec tests? I understand this does the same thing but it would be nice for consistency

PR #249, opened 2019-12-20
Don't refetch images from the store

matthewmcnew:

### DIFF
--- a/pkg/cnb/mountable_buildpack_layer_test.go
+++ b/pkg/cnb/mountable_buildpack_layer_test.go
@@ -1,0 +1,164 @@
+package cnb
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/random"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	"github.com/sclevine/spec"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/pivotal/kpack/pkg/apis/experimental/v1alpha1"
+)
+
+func TestMountableBuildpackLayer(t *testing.T) {
+	spec.Run(t, "Test Mountable Buildpack Layer", testMountableBuildpackLayer)
+}
+
+func testMountableBuildpackLayer(t *testing.T, when spec.G, it spec.S) {
+	handler := http.NewServeMux()
+	server := httptest.NewServer(handler)
+	tagName := fmt.Sprintf("%s/some/image:tag", server.URL[7:])
+
+	var registryCalls []string
+	var backingRandomLayer v1.Layer
+	var layer v1.Layer
+
+	it.Before(func() {
+		var err error
+
+		backingRandomLayer, err = random.Layer(1000, types.DockerLayer)
+		require.NoError(t, err)
+
+		digest, err := backingRandomLayer.Digest()
+		require.NoError(t, err)
+
+		diffID, err := backingRandomLayer.DiffID()
+		require.NoError(t, err)
+
+		size, err := backingRandomLayer.Size()
+		require.NoError(t, err)
+
+		handler.HandleFunc("/v2/", func(writer http.ResponseWriter, request *http.Request) {
+			registryCalls = append(registryCalls, request.RequestURI)
+
+			writer.WriteHeader(200)
+		})
+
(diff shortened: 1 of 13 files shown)


pkg/cnb/mountable_buildpack_layer_test.go
```diff
@@ -0,0 +1,165 @@
...
+			writer.WriteHeader(200)
+		})
+
+		digest, err := backingRandomLayer.Digest()
+		require.NoError(t, err)
+
+		handler.HandleFunc(fmt.Sprintf("/v2/some/image/blobs/%s", digest.String()), func(writer http.ResponseWriter, request *http.Request) {
+			registryCalls = append(registryCalls, request.RequestURI)
+
+			uncompressed, err := backingRandomLayer.Compressed()
```
ashwin-venkatesh: is this the correct assignment for `uncompressed`? should it be rename `compressed` or should the method call be backingRandomLayer.Uncompressed()?

pkg/cnb/mountable_buildpack_layer_test.go
```diff
@@ -0,0 +1,165 @@
...
+		backingRandomLayer, err = random.Layer(1000, types.DockerLayer)
+		require.NoError(t, err)
+
+		handler.HandleFunc("/v2/", func(writer http.ResponseWriter, request *http.Request) {
+			registryCalls = append(registryCalls, request.RequestURI)
+
+			writer.WriteHeader(200)
+		})
+
+		digest, err := backingRandomLayer.Digest()
```
ashwin-venkatesh: just an aesthetic preference but can be move the `handler.HandleFunc()` together and the assignment of digest, diffID and size together?

pkg/cnb/mountable_buildpack_layer_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+	it("MediaType()", func() {
+		mediaType, err := layer.MediaType()
+		require.NoError(t, err)
+
+		assert.Equal(t, types.DockerLayer, mediaType)
+		require.Empty(t, registryCalls)
+	})
+
+	it("Compressed()", func() {
```
ashwin-venkatesh: Can we add a test that makes the Compressed() and/or Uncompressed() calls twice or a call to Compressed() and then Uncompressed() to ensure that it makes the call to remote just once?

pkg/cnb/remote_store_reader.go
```diff
@@ -1,6 +1,8 @@
 package cnb
 
 import (
+	v1 "github.com/google/go-containerregistry/pkg/v1"
```
ashwin-venkatesh: ordering of imports

ashwin-venkatesh: import ordering

pkg/cnb/remote_store_reader_test.go
```diff
@@ -1,6 +1,7 @@
-package cnb_test
+package cnb
 
 import (
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
```
ashwin-venkatesh: import ordering

pkg/cnb/remote_store_reader_test.go
```diff
@@ -77,11 +96,19 @@ func testRemoteStoreReader(t *testing.T, when spec.G, it spec.S) {
 			})
 			require.NoError(t, err)
 
-			fakeClient.AddImage(buildpackageA, buildPackageA, "", expectedKeychain)
+			fakeClient.AddImage(buildpackageA, buildPackageA, expectedKeychain)
```
ashwin-venkatesh: this was not introduced by you, but can we change the reponame and the image name because this is really difficult to read, it being `buildpackageA` and `buildPackageA`

ashwin-venkatesh: Same thing with `buildpackageB` and `buildPackageB`

pkg/cnb/mountable_buildpack_layer_test.go
matthewmcnew: Good Catch. This should be `compressed`

pkg/cnb/mountable_buildpack_layer_test.go
matthewmcnew: 👍

pkg/cnb/mountable_buildpack_layer_test.go
matthewmcnew: The backing remote.Layer will actually call the registry multiple times on each call of `Compressed` or `Uncompressed`. (Doesn't store the layer contents in memory) 

The memoization only prevents multiple calls to `/v1/` (for auth) again.  

So, I am inclined to not bother testing it.

pkg/cnb/remote_store_reader.go
matthewmcnew: 🙏

matthewmcnew: 🙏

pkg/cnb/remote_store_reader_test.go
matthewmcnew: 🙏

pkg/cnb/remote_store_reader_test.go
matthewmcnew: Agreed

PR #256, opened 2019-12-31
Add stack CRD

djoyahoy:
- Use stack ref in custom builder specs

Signed-off-by: Danny Joyce <djoyce@pivotal.io>


samples/experimental/custom_builder.yaml
```diff
@@ -6,13 +6,9 @@ metadata:
 spec:
   serviceAccount: default
   tag: sample/custom-builder
-  stack:
-    baseBuilderImage: cloudfoundry/cnb:bionic
+  stack: cflinuxfs3-stack
   store: sample-store
   order:
-  - group:
```
matthewmcnew: Why drop the yarn group?

samples/experimental/stack.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: experimental.kpack.pivotal.io/v1alpha1
+kind: Stack
+metadata:
+  name: cflinuxfs3-stack
+spec:
+  id: "io.buildpacks.stacks.cflinuxfs3"
```
matthewmcnew: Up until this point all of our examples and tutorials have pointed to bionic. Could we keep that?

pkg/reconciler/v1alpha1/store/store_test.go
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 func TestStoreReconciler(t *testing.T) {
-	spec.Run(t, "Custom Cluster Builder Reconciler", testStoreReconciler)
+	spec.Run(t, "Store Reconciler", testStoreReconciler)
```
matthewmcnew: Good catch. 👍

pkg/apis/experimental/v1alpha1/stack_types.go
```diff
@@ -0,0 +1,50 @@
...
+
+type StackImage struct {
+	Image string `json:"image"`
+}
+
+type StackStatus struct {
+	duckv1alpha1.Status `json:",inline"`
+
+	BuildImageRef string `json:"buildImageRef"`
+	RunImageRef   string `json:"runImageRef"`
```
matthewmcnew: I am concerned that ref in k8s might imply another resource. On our builders we just call the resolved RunImage `status.stack.runImage`. However, for consistency, I wonder if we should have `status.runImage.latestImage` here and on our builder status.

pkg/apis/experimental/v1alpha1/stack_types.go
```diff
@@ -0,0 +1,50 @@
...
+	Id         string     `json:"id"`
+	BuildImage StackImage `json:"buildImage"`
+	RunImage   StackImage `json:"runImage"`
+}
+
+type StackImage struct {
+	Image string `json:"image"`
+}
+
+type StackStatus struct {
```
matthewmcnew: Not on this PR. But, help me remember to `omitempty` on these resources before we release.

pkg/cnb/builder_builder.go
```diff
@@ -83,12 +80,27 @@ func (bb *BuilderBuilder) writeableImage() (v1.Image, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	image, err := mutate.AppendLayers(bb.baseImage, append(lifecycleLayers, stackLayer)...)
+	if err != nil {
+		return nil, err
+	}
+
+	image, err = mutate.AppendLayers(image, append(layers, orderLayer)...)
```
matthewmcnew: Could we combine these, perhaps with a helper?

mutate.AppendLayers(image, append(layers, orderLayer)...)

hack/common.sh
```diff
@@ -39,5 +39,11 @@ function compile() {
...
   completion_image=${resolved_image_name}
 
-  ytt -f config/. -v controller_image=${controller_image} -v webhook_image=${webhook_image} -v build_init_image=${build_init_image} -v rebase_image=${rebase_image} -v completion_image=${completion_image} > $output
+  ytt -f config/. \
+    -v controller_image=${controller_image} \
+    -v webhook_image=${webhook_image} \
+    -v build_init_image=${build_init_image} \
+    -v rebase_image=${rebase_image} \
+    -v completion_image=${completion_image} \
+    -v lifecycle_image="gcr.io/cf-build-service-public/lifecycle-0.5.0" > $output
```
matthewmcnew: Where does this image come from? Can we avoid needing to build a lifecycle image to our toolchain by just pointing it at the lifecycle layer in an existing image. After all, it just needs to be a layer. 

Like this layer?`cloudfoundry/cnb@sha256:8f0b2d09ab4b38530a1630403967d11a601e56e02e79d3f56370d34fd071fe38`

matthewmcnew: We could use remote.Layer("cloudfoundry/cnb@sha256:8f0b2d09ab4b38530a1630403967d11a601e56e02e79d3f56370d34fd071fe38") to construct it.

pkg/reconciler/v1alpha1/custombuilder/custombuilder.go
```diff
@@ -107,6 +111,16 @@ func (c *Reconciler) reconcileCustomBuilder(customBuilder *expv1alpha1.CustomBui
 		return v1alpha1.BuilderRecord{}, err
 	}
 
+	stack, err := c.StackLister.Get(customBuilder.Spec.Stack)
+	if err != nil {
+		return v1alpha1.BuilderRecord{}, err
+	}
+
+	err = c.Tracker.Track(stack, customBuilder.NamespacedName())
```
matthewmcnew: Nice!

matthewmcnew: Can we add tests for these trackers?

pkg/cnb/builder_builder.go
```diff
@@ -83,12 +80,27 @@ func (bb *BuilderBuilder) writeableImage() (v1.Image, error) {
...
 		return nil, err
 	}
 
-	image, err := mutate.AppendLayers(bb.baseImage, append(layers, orderLayer)...)
+	lifecycleLayers, err := bb.lifecycleImage.Layers()
+	if err != nil {
+		return nil, err
+	}
+
+	image, err := mutate.AppendLayers(bb.baseImage, append(lifecycleLayers, stackLayer)...)
```
matthewmcnew: Do we need pack's "default dirs" layer? 

https://github.com/buildpacks/pack/blob/e4315be24c103e3c5722b08561bf13a55876cbbc/internal/builder/builder.go#L481

matthewmcnew:
This is great! A real game changer. I hope to read through the tests one more time.

samples/experimental/custom_builder.yaml
ashwin-venkatesh: we did this for a test. Should not have gotten checked in. Good catch.

pkg/apis/experimental/v1alpha1/stack_types.go
ashwin-venkatesh: Makes sense.

samples/experimental/custom_builder.yaml
djoyahoy: Well we dropped it because the sample store does not include it. We can either keep it dropped or add the yarn buildpackage to the sample store.

samples/experimental/stack.yaml
djoyahoy: Sure!

hack/common.sh
djoyahoy: I think this makes sense. I like the idea of not having to maintain a new image. I do worry that it is confusing to just pull a layer from another image, but the savings probably out-weigh the confusion factor.

pkg/cnb/builder_builder.go
djoyahoy: We probably do. Good call!

samples/experimental/custom_builder.yaml
ashwin-venkatesh: I dont think we have a buildpackage for yarn just yet, so this would lead to a builder that cannot get built, unless we use another builder as a store

hack/common.sh
ashwin-venkatesh: While looking into this somemore today, Danny and I realized that it probably makes more sense to use a lifecycle image instead of a lifecycle layer, as images are distributable artifacts in a way that layers aren't. Additionally, we can attached lifecycle metadata to the lifecycle image but that wouldnt be possible in the case of a layer. I agree that this involves additional work on our end to maintain that image, but I think there are advantages to this approach outweigh referencing a layer

samples/experimental/custom_builder.yaml
matthewmcnew: Should we update the sample store to include cloudfoundry/bionic?

pkg/cnb/builder_builder.go
```diff
@@ -4,43 +4,85 @@ import (
...
 	}
 
-	stackID, err := imagehelpers.GetStringLabel(baseImage, stackMetadataLabel)
+	configFile, err := baseImage.ConfigFile()
 	if err != nil {
 		return nil, err
 	}
 
+	var userId, groupId int
+	for _, envStr := range configFile.Config.Env {
```
matthewmcnew: Could we move this to a helper?

pkg/cnb/builder_builder.go
```diff
@@ -109,27 +151,65 @@ func (bb *BuilderBuilder) writeableImage() (v1.Image, error) {
...
-	return singeFileLayer(orderTomlPath, orderBuf.Bytes())
+	builderLayers = append(builderLayers, lifecycleLayers...)
+
+	stackLayer, err := bb.stackLayer()
+	if err != nil {
+		return nil, err
+	}
+	builderLayers = append(builderLayers, stackLayer)
+
+	orderLayer, err := bb.orderLayer()
```
matthewmcnew: Order layer should be the last layer in the image. (Most likely to change)

samples/experimental/stack.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: experimental.kpack.pivotal.io/v1alpha1
+kind: Stack
+metadata:
+  name: bionic-stack
+spec:
+  id: "io.buildpacks.stacks.bionic"
+  buildImage:
+    image: "cloudfoundry/build:base-cnb"
```
matthewmcnew: I wonder if we should put the digest here in the samples?

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -209,7 +209,7 @@ func (b *Build) BuildPod(config BuildPodImages, secrets []corev1.Secret, buildPo
 				{
 					Name:    "detect",
 					Image:   builderImage,
-					Command: []string{"/lifecycle/detector"},
+					Command: []string{"/cnb/lifecycle/detector"},
```
matthewmcnew: We should probably keep executing `/lifecycle` directly and follow the lead of pack and the [tekton catalog](https://github.com/tektoncd/catalog/blob/master/buildpacks/buildpacks-v3.yaml) if/when we move to /cnb/lifecycle

matthewmcnew: This will require us to add a symlink but, it will allow our builders to still be used with pack.

test/config.go
```diff
@@ -5,8 +5,13 @@ import (
 	"os"
 	"strconv"
 	"testing"
+	"time"
 )
 
+func init() {
+	rand.Seed(time.Now().Unix())
```
matthewmcnew: This is good! What drove this out?

matthewmcnew: Should we just move it before we call spec?

samples/experimental/custom_builder.yaml
ashwin-venkatesh: I do like the idea of surfacing a builder as a store, but am worried it might confuse people who are browsing through it. But it does allow us to bring `yarn` back 🤷‍♂

samples/experimental/stack.yaml
ashwin-venkatesh: instead of the tags for the images?

pkg/apis/experimental/v1alpha1/custom_builder_types.go
```diff
@@ -23,7 +23,7 @@ type CustomBuilder struct {
 
 type CustomBuilderSpec struct {
 	Tag   string       `json:"tag"`
-	Stack Stack        `json:"stack"`
+	Stack string       `json:"stack"`
```
ashwin-venkatesh: I wonder if we should rename these fields to `StackRef` and `StoreRef` as they actually refer to kubernetes objects of that type

pkg/apis/experimental/v1alpha1/custom_cluster_builder_validation_test.go
```diff
@@ -23,10 +23,8 @@ func testCustomBuilderValidation(t *testing.T, when spec.G, it spec.S) {
 		},
 		Spec: CustomClusterBuilderSpec{
 			CustomBuilderSpec: CustomBuilderSpec{
-				Tag: "some-registry.io/custom-builder",
-				Stack: Stack{
-					BaseBuilderImage: "some-registry.io/base-builder",
-				},
+				Tag:   "some-registry.io/custom-builder",
+				Stack: "some-stack-ref",
 				Store: "some-registry.io/store",
```
ashwin-venkatesh: This is not related to this PR but we should consider renaming some of these test fields to read to refs and not images in a registry to avoid confusion.

PR #273, opened 2020-01-16
use kbld to build images for kpack

ashwin-venkatesh:
Things left to do:
- ~~Figure out why ytt is not templating the image values into the final template~~
- How to make the image destinations configurable. Possibly make `images.yaml` a template that I ytt the destination into?
- Once this gets merged in, update pipelines to ensure things continue to get templated in accurately


tylerphelan:
closing b/c kpack images are now built with...kpack!

PR #274, opened 2020-01-16
Propagate annotations from image->build and build->buildpod

### LINKED ISSUE #32: [Feature] Selectively propagate annotations from parent to child resources
This is related to #8 .

When we get an update of the PBS custom resource via the K8s watch API, we need to extract some data to look up the corresponding object in our backend. Label value has too many restrictions so it's not suitable in this case.

> Valid label values must be 63 characters or less and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.

A good alternative is the annotation. To avoid conflicts with the system generated annotations, we may propagate the annotation values under some predefined (or configurable domain), such as:

```
Prefix: meta.build.pivotal.io/
meta.build.pivotal.io/key1 = blabla
meta.build.pivotal.io/some-other-key = blabla
```

ashwin-venkatesh:
Closes #32

### DIFF
--- a/pkg/apis/build/v1alpha1/build_pod.go
+++ b/pkg/apis/build/v1alpha1/build_pod.go
@@ -105,9 +105,10 @@ func (b *Build) BuildPod(config BuildPodImages, secrets []corev1.Secret, bc Buil
 		ObjectMeta: v1.ObjectMeta{
 			Name:      b.PodName(),
 			Namespace: b.Namespace,
-			Labels: b.labels(map[string]string{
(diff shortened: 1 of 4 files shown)


matthewmcnew:
We took inspiration from tekton (to keep it simple for now) and just propagate all annotations to build pods. 

As reference: https://github.com/tektoncd/pipeline/blob/ebb071e6b9c054040d70af309c584b274ef63cca/pkg/pod/pod.go#L193

PR #278, opened 2020-01-23
Cowardly refuse to attempt builds with a platform api > 0.2

matthewmcnew:
- Save build pod creation error in build status

### DIFF
--- a/pkg/reconciler/v1alpha1/build/build.go
+++ b/pkg/reconciler/v1alpha1/build/build.go
@@ -84,6 +84,19 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 	build = build.DeepCopy()
 	build.SetDefaults(ctx)
 
+	reconcileErr := c.reconcile(build)
+	if reconcileErr != nil {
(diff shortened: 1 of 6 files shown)


ashwin-venkatesh:
Additionally, it would be nice, if our builders that had a lifecycle with a Platform API of > 0.2, we dont mark them as ready, and keep condition false with an `unsupported version` error message. I am ok with that being a separate PR.

pkg/reconciler/v1alpha1/build/build.go
```diff
@@ -109,24 +122,22 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
-		}
-		return c.K8sClient.CoreV1().Pods(build.Namespace).Create(podConfig)
+	} else if !k8s_errors.IsNotFound(err) {
+		return pod, nil
 	}
 
-	return pod, nil
+	podConfig, err := c.PodGenerator.Generate(build)
+	if err != nil {
+		return nil, controller.NewPermanentError(err)
```
ashwin-venkatesh: ❤️

PR #282, opened 2020-01-31
Preemptively load azurecredentialhelperfix package for real

matthewmcnew:
- Previously it was only getting loaded in tests preventing it from being picked up in the actual controller


release-0.0.6-asc.yaml
```diff
@@ -0,0 +1,417 @@
+apiVersion: v1
```
ashwin-venkatesh: Was this made a part of the PR accidentally?

matthewmcnew: yes

PR #295, opened 2020-02-20
Ensure clientsets are configured just once for acceptance tests

ashwin-venkatesh:
The end to end tests are currently failing in the CI before the before block of the tests attempts to recreate the clients required by the tests. Use a sync.Once to ensure the before block does not fail. Also updated testify while we were at it.


matthewmcnew:
Might make sense to move these to execute before invoking spec.

ashwin-venkatesh:
> Might make sense to move these to execute before invoking spec.

I like this idea

PR #305, opened 2020-03-10
Ensure build and run image have the same stack as the stack.spec.id

### LINKED ISSUE #272: Ensure stack resource validates stack id across resource
Currently, the stack resource determines the `id` of the stack from the id field in the spec of the resource and does not verify if it matches the id specified in the `io.buildpacks.stack.id` of the build and the run image. Additionally, the build and run image can have different stack id's which should be a "broken" stack, but a stack still gets built. `kpack` should have stricter validations on stack inputs

ashwin-venkatesh:
- Make the stack resource not ready if there is a mismatch in the stack mentioned by the images and the spec
- Update openAPI module version

Closes #272

### DIFF
--- a/pkg/reconciler/v1alpha1/stack/stack.go
+++ b/pkg/reconciler/v1alpha1/stack/stack.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	ggcrv1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
@@ -19,11 +20,13 @@ import (
 	v1alpha1expListers "github.com/pivotal/kpack/pkg/client/listers/experimental/v1alpha1"
 	"github.com/pivotal/kpack/pkg/reconciler"
 	"github.com/pivotal/kpack/pkg/registry"
+	"github.com/pivotal/kpack/pkg/registry/imagehelpers"
 )
 
 const (
 	ReconcilerName = "Stacks"
 	Kind           = "Stack"
+	StackLabel     = "io.buildpacks.stack.id"
 )
 
 type ImageFetcher interface {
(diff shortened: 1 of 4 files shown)


pkg/reconciler/v1alpha1/stack/stack.go
```diff
@@ -123,17 +126,53 @@ func (c *Reconciler) getStackImages(stackSpec expv1alpha1.StackSpec) (string, st
...
+
+func getStackLabel(image ggcrv1.Image) (string, error) {
+	config, err := image.ConfigFile()
+	if err != nil {
+		return "", err
+	}
+	labels := config.Config.Labels
+	id, ok := labels[StackLabel]
+	if !ok {
+		return "", errors.New("invalid stack image")
```
ashwin-venkatesh: will improve this to be specific about which image is the invalid one

pkg/reconciler/v1alpha1/stack/stack.go
```diff
@@ -123,17 +126,53 @@ func (c *Reconciler) getStackImages(stackSpec expv1alpha1.StackSpec) (string, st
...
+	}
+
+	if (buildStack == stackId) && (runStack == stackId) {
+		return nil
+	} else {
+		return errors.New(fmt.Sprintf("Invalid stack images. Expected stack: %s, Build Image stack: %s, Run Image stack: %s", stackId, buildStack, runStack))
+	}
+}
+
+func getStackLabel(image ggcrv1.Image) (string, error) {
```
matthewmcnew: This could just use `imagehelpers.GetStringLabel`

pkg/reconciler/v1alpha1/stack/stack.go
```diff
@@ -123,17 +126,53 @@ func (c *Reconciler) getStackImages(stackSpec expv1alpha1.StackSpec) (string, st
...
+
+	runStack, err := getStackLabel(runImage)
+	if err != nil {
+		return err
+	}
+
+	if (buildStack == stackId) && (runStack == stackId) {
+		return nil
+	} else {
+		return errors.New(fmt.Sprintf("Invalid stack images. Expected stack: %s, Build Image stack: %s, Run Image stack: %s", stackId, buildStack, runStack))
```
matthewmcnew: Could you use `errors.Errorf` instead of `errors.New(fmt.Sprintf` ?

matthewmcnew: Also, super minor but, could we avoid capitalizing the error string? 
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

PR #308, opened 2020-03-11
Stack id validation

matthewmcnew:
- Don't create CustomBuilder's if stack is not ready. 
- No mixin validation, yet.

### DIFF
--- a/pkg/cnb/builder_builder.go
+++ b/pkg/cnb/builder_builder.go
@@ -35,48 +35,46 @@ const (
 
 var normalizedTime = time.Date(1980, time.January, 1, 0, 0, 1, 0, time.UTC)
 
-type BuilderBuilder struct {
+type builderBlder struct {
+	err error
+
 	baseImage         v1.Image
 	lifecycleImage    v1.Image
 	LifecycleMetadata LifecycleMetadata
-	stack             *expv1alpha1.Stack
+	stackId           string
 	order             []expv1alpha1.OrderEntry
 	buildpackLayers   map[expv1alpha1.BuildpackInfo]buildpackLayer
 	cnbUserId         int
 	cnbGroupId        int
 	kpackVersion      string
+	runImage          string
 }
 
-func newBuilderBuilder(baseImage v1.Image, lifecycleImage v1.Image, stack *expv1alpha1.Stack, kpackVersion string) (*BuilderBuilder, error) {
+func newBuilderBldr(lifecycleImage v1.Image, kpackVersion string) (*builderBlder, error) {
 	lifecycleMd := LifecycleMetadata{}
 	err := imagehelpers.GetLabel(lifecycleImage, lifecycleMetadataLabel, &lifecycleMd)
 	if err != nil {
 		return nil, err
 	}
 
-	userId, err := parseCNBID(baseImage, cnbUserId)
-	if err != nil {
-		return nil, err
-	}
-
-	groupId, err := parseCNBID(baseImage, cnbGroupId)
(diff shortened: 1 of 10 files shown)


pkg/cnb/builder_builder.go
```diff
@@ -172,7 +171,21 @@ func (bb *BuilderBuilder) writeableImage() (v1.Image, error) {
 	})
 }
 
-func (bb *BuilderBuilder) lifecycleLayer() (v1.Layer, error) {
+func (bb *builderBlder) validateBuilder() {
+	for _, bp := range bb.buildpackLayers {
+		err := bp.BuildpackLayerInfo.supports(bb.stackId)
+		if err != nil {
+			bb.err = errors.Wrapf(err, "validating buildpack: %s", bp.BuildpackInfo)
+			return
```
ashwin-venkatesh: i wonder if we should go through every buildpack in the spec and list all the once that are unsupported

pkg/cnb/create_builder.go
```diff
@@ -34,11 +34,17 @@ func (r *RemoteBuilderCreator) CreateBuilder(keychain authn.Keychain, buildpackR
...
-	builderBuilder, err := newBuilderBuilder(buildImage, lifecycleImage, stack, r.KpackVersion)
+	builderBldr, err := newBuilderBldr(lifecycleImage, r.KpackVersion)
 	if err != nil {
 		return v1alpha1.BuilderRecord{}, err
 	}
 
+	builderBldr.AddStack(
+		stack.Spec.Id,
+		buildImage,
+		[]string{},
```
ashwin-venkatesh: one day, we shall have mixins :)

pkg/reconciler/v1alpha1/custombuilder/custombuilder.go
```diff
@@ -121,6 +123,10 @@ func (c *Reconciler) reconcileCustomBuilder(customBuilder *expv1alpha1.CustomBui
 		return v1alpha1.BuilderRecord{}, err
 	}
 
+	if !stack.Status.GetCondition(corev1alpha1.ConditionReady).IsTrue() {
```
ashwin-venkatesh: ❤️

pkg/cnb/builder_builder.go
matthewmcnew: That would require us to combine the errors. Which would be great :)

pkg/cnb/create_builder.go
matthewmcnew: Maybe, we could remove this field until we support it. I was just playing with what it looks like.

pkg/cnb/builder_builder.go
```diff
@@ -35,48 +35,46 @@ const (
 
 var normalizedTime = time.Date(1980, time.January, 1, 0, 0, 1, 0, time.UTC)
 
-type BuilderBuilder struct {
+type builderBlder struct {
+	err error
```
matthewmcnew: Error handling strategy inspired by Dave Cheney 

https://dave.cheney.net/2019/01/27/eliminate-error-handling-by-eliminating-errors

pkg/cnb/create_builder.go
ashwin-venkatesh: It threw me off a little bit when I saw it getting passed in and had to do a double take. I'm alright with keeping it around if we throw a tiny comment about why we are passing in an empty slice 🙂

matthewmcnew: Mixins are the next thing to work on :)

PR #309, opened 2020-03-11
Select buildpacks for builders by comparing semvers

djoyahoy:
When there are duplicates of a buildpack but one of those buildpacks does not have a semver compatible version, then the builder will not roll forward and the status will have a corresponding error message.

### DIFF
--- a/pkg/cnb/store_buildpack_repository.go
+++ b/pkg/cnb/store_buildpack_repository.go
@@ -3,6 +3,7 @@ package cnb
 import (
 	"sort"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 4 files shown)


pkg/cnb/store_buildpack_repository.go
```diff
@@ -93,9 +94,14 @@ func (s *StoreBuildpackRepository) layersForOrder(order v1alpha1.Order) ([]build
 	return buildpackLayers, nil
 }
 
-func highestVersion(matchingBuildpacks []v1alpha1.StoreBuildpack) v1alpha1.StoreBuildpack {
+func highestVersion(matchingBuildpacks []v1alpha1.StoreBuildpack) (v1alpha1.StoreBuildpack, error) {
+	for _, bp := range matchingBuildpacks {
+		if _, err := semver.NewVersion(bp.Version); err != nil {
+			return v1alpha1.StoreBuildpack{}, errors.Errorf("cannot find buildpack '%s' with latest version due to invalid semver '%s'", bp.Id, bp.Version)
```
matthewmcnew: Really smart. I like the verify it is valid Semver and then sort with `semver.MustParse`

matthewmcnew: Makes this really simple.

Issue #254, opened 2019-12-30
CustomBuilder's should not create builders with unsupported stacks


matthewmcnew:
CustomBuilder's should not create builders with buildpacks that don't support the builder's stack.

ashwin-venkatesh:
If this [spec](https://github.com/buildpacks/rfcs/blob/bionic-mixins/text/0000-bionic-mixins.md) gets merged, we might have to ensure we check mixin information for the buildpack as well. Additionally, we will need the stack images to surface their mixin information via metadata possibly.

PR #312, opened 2020-03-12
Mixin validation

matthewmcnew:
Resolves https://github.com/pivotal/kpack/issues/254


pkg/cnb/remote_stack_reader.go
```diff
@@ -0,0 +1,154 @@
...
+	if !hasLabel || err != nil {
+		return mixins, err
+	}
+
+	err = imagehelpers.GetLabel(image, MixinsLabel, &mixins)
+	return mixins, err
+}
+
+func mixins(build, run []string) ([]string, error) {
+	invalid, buildStage, buildCommon := classifyMixins(build)
```
djoyahoy: `invalid` swaps places when classifying for build and run. That is a bit confusing. Can we pass in a flag to classifyMixins so that it spits out either build or run but not both?

matthewmcnew: Good catch. I like that idea. I tried passing in the mixin prefixes. 

Let me know what you think.

PR #323, opened 2020-03-17
Only restrict webhook's rbac secret resourceName on get

matthewmcnew:
- resourceNames cannot be restricted on create
- Remove redundant mutatingwebhookconfigurations rbac
- Convert RoleBinding on ClusterResources to ClusterRoleBinding

### DIFF
--- a/config/webhookrole.yaml
+++ b/config/webhookrole.yaml
@@ -19,17 +19,12 @@ rules:
   - webhook-certs
   verbs:
   - get
-  - create
-  - update
 - apiGroups:
-  - "admissionregistration.k8s.io"
+  - ""
   resources:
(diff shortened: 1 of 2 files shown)


config/webhookrole.yaml
```diff
@@ -61,7 +56,7 @@ rules:
   - update
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: RoleBinding
+kind: ClusterRoleBinding
 metadata:
   name: kpack-webhook-certs-mutatingwebhookconfiguration-admin-binding
   namespace: kpack
```
ashwin-venkatesh: we should drop the namespace from here

matthewmcnew: 👍

PR #327, opened 2020-03-18
Better error when run image is private

ashwin-venkatesh:


pkg/dockercreds/access_checker.go
```diff
@@ -71,7 +71,11 @@ func HasReadAccess(keychain authn.Keychain, tag string) (bool, error) {
 	}
 	_, err = remote.Get(ref, remote.WithAuthFromKeychain(keychain), remote.WithTransport(http.DefaultTransport))
 	if err != nil {
-		return false, errors.Wrapf(err, "connect to registry store '%s'", tag)
+		if _, ok := err.(*transport.Error); ok {
+			return false, nil
+		}
+
+		return false, errors.Wrapf(err, "validating write access to: %s", tag)
```
ashwin-venkatesh:
```suggestion
		return false, errors.Wrapf(err, "validating read access to: %s", tag)
```

matthewmcnew: 👍

PR #331, opened 2020-03-19
Update acr fix to use new gggcr.k8schain import paths

matthewmcnew:
- The azure credential provider needs to be imported first to initialize the pflag.


matthewmcnew:
Because https://github.com/google/go-containerregistry/pull/652 was merged we could investigate moving all our flags to use the pflags library and avoid this "fix".

PR #359, opened 2020-04-22
Update id for yarn

mgibson1121:
My ccb wasn't building successfully due to yarn ID not being available in the store.  I'd imagine at some point we will have to swap out the IDs for paketo, so maybe just do that en-masse.  If this PR is accepted I'd be happy to make the change in other samples.

Update: Saw Matt McNew's previous PR but it didn't cover samples

### DIFF
--- a/samples/experimental/custom_cluster_builder.yaml
+++ b/samples/experimental/custom_cluster_builder.yaml
@@ -12,7 +12,7 @@ spec:
   order:
   - group:
     - id: org.cloudfoundry.node-engine
-    - id: org.cloudfoundry.yarn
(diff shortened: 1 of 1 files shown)


matthewmcnew:
Good Catch! https://github.com/pivotal/kpack/pull/358 updated to paketo but, yarn should still be: `paketo-buildpacks/yarn-install`.

Could you rebase and update your pr?

mgibson1121:
Closing this PR because #358 the CCB sample to reference Java Paketo CNBs, which is probably more applicable than the Node ones anyway.

Issue #369, opened 2020-04-29
Support CNB Service Bindings


matthewmcnew:
kpack should support [CNB service bindings](https://github.com/buildpacks/rfcs/blob/master/text/0012-service-binding.md), so that images can be properly configured for services at build time. 

This should be enabled by allowing images and builds to be configured with a list of bindings:
```
name: image
...
spec:
  ...
  bindings:
  - name: my-binding
    metadataRef:
      name: provider-binding-metadata
    secretRef:
      name: provider-binding-secret
```

MetadataRefs should reference a config map and secrets should reference a secret.

scothis:
While I think this is something kpack needs to support, it also loosens the security posture of kpack. Since the build/image creator can specify a secret name that is mounted into the build, this exposes an enumeration attack against secrets that are in the same namespace as the build's pod. Consumers of kpack that host multi-tenant builds in the same namespace should stop doing that, or risk exposure of secrets.

This is no different than allowing the user to create a Pod. To date, kpack has limited the surface area of the PodSpec that is configurable.

matthewmcnew:
That  is really good point @scothis. We have gone to great lengths to prevent user provided code from executing with access to registry credentials and this would allow any user to be able to read those credentials.

scothis:
We settled on only allowing secrets to be bound that are not attached to any service account. This will protect secrets like K8s API service accounts, registry and source repository credentials, while still allowing access to other secrets.

PR #371, opened 2020-04-30
Add CNB Service Bindings to Image and Build resources

scothis:
The CNB Service Bindings spec provides a structured way to expose
service metadata and credentials into a build. For example, the build
can use the binding information to install a database driver, configure
an APM agent for the built image, or discover any other capability
outside of the built container.

Each binding reference has a name, a reference to a k8s ConfigMap
holding service metadata, and optionally a k8s Secret holding service
credentials. Builds generally should not have access to the credentials
so that the values are not backed into the built image, but instead
bound fresh at runtime.

For images, the bindings are located at `.spec.build.bindings`:

```
kind: Image
...
spec:
  ...
  build:
    bindings:
    - name: my-binding
      metadataRef:
        name: provider-binding-metadata
      secretRef:
        name: provider-binding-secret
```

For builds, the bindings are located at `.spec.bindings`:

```
kind: Build
...
spec:
  ...
  bindings:
  - name: my-binding
    metadataRef:
      name: provider-binding-metadata
    secretRef:
      name: provider-binding-secret
```

In the build the bindings appear as a directory tree under the path
specified by the `$CNB_BINDINGS` env var. Each binding is a directory,
which in turn has a `metadata` directory with files and if credentials
are exposed a `secret` directory with more files. The environment
variable and volume mounts are available during the detect and build
steps. Currently, the value of `$CNB_BINDINGS` is `/var/bindings` but
should not be presumed to be stable or consistent with the value at
runtime.

There is no guarantee the same services, or types of services, will be
bound at runtime.

Refs #369


pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -84,6 +84,10 @@ var (
 		MountPath: "/builderPullSecrets",
 		ReadOnly:  true,
 	}
+	bindingsEnv = corev1.EnvVar{
+		Name:  "CNB_BINDINGS",
+		Value: "/var/bindings",
```
matthewmcnew, line 89: It seems that [buildpacks spec](https://github.com/buildpacks/spec/blob/master/extensions/bindings.md#detect-and-build-phases) expects that this in `/platform/bindings`.

matthewmcnew:
This is so good @scothis!  

I'd like to take to a moment to think through the security implications of allowing arbitrary secrets to be mounted in the build pod as you aptly noted in https://github.com/pivotal/kpack/issues/369.

pkg/apis/build/v1alpha1/build_pod.go
scothis: Good catch. I was paying attention to the launch phase which is variable based.

scothis: Moved the mount path to be under `/platform/bindings`, I need to test it to make sure the nested volume mounts don't cause issues

scothis:
Used the validator to disable the secretRef on the binding until we have a better understanding of how to protect again secret enumeration attacks, or direct capturing of secrets if they know the name.

scothis:
Based on our conversation yesterday, I switch up the secret handling to forbid binding secrets that are attached to any service account in the namespace. This should offer a reasonable compromise to protect registry, and api server credentials while still allowing bindings to work as expected.

pkg/buildpod/generator_test.go
```diff
@@ -184,6 +185,31 @@ func testGenerator(t *testing.T, when spec.G, it spec.S) {
...
+				K8sClient:       fakeK8sClient,
+				KeychainFactory: keychainFactory,
+				ImageFetcher:    imageFetcher,
+			}
+
+			var build = &testBuildPodable{
+				namespace: namespace,
+				bindings: []v1alpha1.Binding{
+					{
+						Name:        "naughty",
```
ashwin-venkatesh, line 202: lolol

samples/image_with_service_bindings.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: build.pivotal.io/v1alpha1
```
ashwin-venkatesh, line 1: Love the samples!!!

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -110,6 +110,51 @@ func (b *Build) BuildPod(config BuildPodImages, secrets []corev1.Secret, bc Buil
 		SubPath:   b.Spec.Source.SubPath, // empty string is a nop
 	}
 
+	bindingVolumes := []corev1.Volume{}
```
ashwin-venkatesh, line 113: can this be moved into a private method `setupBindings()`?

scothis: done

PR #372, opened 2020-05-01
Add buildpack buildpackage info to the store

djoyahoy:


pkg/cnb/remote_store_reader.go
```diff
@@ -22,9 +22,15 @@ func (r *RemoteStoreReader) Read(storeImages []v1alpha1.StoreImage) ([]v1alpha1.
...
+		storeImageCopy := storeImage
 		g.Go(func() error {
-			image, _, err := r.RegistryClient.Fetch(r.Keychain, storeImage.Image)
+			image, _, err := r.RegistryClient.Fetch(r.Keychain, storeImageCopy.Image)
+			if err != nil {
+				return err
+			}
+
+			bpMetadata := BuildpackageMetadata{}
+			err = imagehelpers.GetLabel(image, buildpackageMetadataLabel, &bpMetadata)
```
matthewmcnew, line 33: What happens if this is builder image without the label?

I know the capi team is still planning on doing that.

djoyahoy: I did not realize we'd have to handle that case. I suppose we could leave the buildpackage info blank. Thats the best I can think of, considering those are non-conformant buildpackages, in my opinion.

matthewmcnew: It might be good idea to move to only supporting buildpackages.

matthewmcnew: Actually, they might just have been doing that just to get started because at the time there were not really any real buildpackages.

matthewmcnew: I think we can safely ignore that case.

PR #375, opened 2020-05-08
Improve error message for prepare step

sukhil-suresh:
- wrap unhandled server errors returned by container registry with a more reasonable error message
- removed redundant code from test file

Before:
```
[prepare] GET https://dev.registry.pivotal.io/service/token?scope=repository%3Agarbage%3Apush%2Cpull&service=harbor-registry: unsupported status code 500
```
After:
```
[prepare] Error validating write permission to dev.registry.pivotal.io/garbage. GET https://dev.registry.pivotal.io/service/token?scope=repository%3Agarbage%3Apush%2Cpull&service=harbor-registry: unsupported status code 500
```

Signed-off-by: Sukhil Suresh <ssuresh@pivotal.io>

### DIFF
--- a/pkg/dockercreds/access_checker.go
+++ b/pkg/dockercreds/access_checker.go
@@ -39,6 +39,7 @@ func HasWriteAccess(keychain authn.Keychain, tag string) (bool, error) {
 			}
 		}
 
+		err = errors.Errorf("Error validating write permission to %s. %s", tag, err.Error())
 		return false, errors.WithStack(err)
 	}
 
--- a/pkg/dockercreds/access_checker.go
+++ b/pkg/dockercreds/access_checker.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
@@ -27,19 +28,7 @@ func VerifyWriteAccess(keychain authn.Keychain, tag string) error {
 	scopes := []string{ref.Scope(transport.PushScope)}
 	tr, err := transport.New(ref.Context().Registry, auth, http.DefaultTransport, scopes)
 	if err != nil {
-		if transportError, ok := err.(*transport.Error); ok {
-			for _, diagnosticError := range transportError.Errors {
-				if diagnosticError.Code == transport.UnauthorizedErrorCode {
-					return errors.Wrap(err, "Unauthorized")
-				}
-			}
-
-			if transportError.StatusCode == 401 {
-				return errors.Wrap(err, "Unauthorized")
-			}
-		}
-
-		return errors.WithStack(err)
+		return diagnoseIfTransportError(err)
 	}
 
 	client := &http.Client{Transport: tr}
@@ -53,12 +42,12 @@ func VerifyWriteAccess(keychain authn.Keychain, tag string) error {
 	// Make the request to initiate the blob upload.
 	resp, err := client.Post(u.String(), "application/json", nil)
 	if err != nil {
-		return errors.WithStack(err)
+		return diagnoseIfTransportError(err)
 	}
 	defer resp.Body.Close()
 
 	if err = transport.CheckError(resp, http.StatusCreated, http.StatusAccepted); err != nil {
-		return errors.WithStack(err)
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


pkg/dockercreds/access_checker_test.go
```diff
@@ -111,10 +105,20 @@ func testAccessChecker(t *testing.T, when spec.G, it spec.S) {
...
+			assert.False(t, hasAccess)
+		})
+
+		it("wraps unhandled server errors with a reasonable error message", func() {
+			handler.HandleFunc("/v2/", func(writer http.ResponseWriter, request *http.Request) {
+				writer.WriteHeader(500)
+			})
 
 			hasAccess, err := HasWriteAccess(testKeychain{}, tagName)
 			require.Error(t, err)
```
ashwin-venkatesh, line 119: I think there is a matcher called `EqualError` that allows one to also assert on the error message in the same require, but that is a minor nit.

pkg/dockercreds/access_checker.go
```diff
@@ -39,6 +39,7 @@ func HasWriteAccess(keychain authn.Keychain, tag string) (bool, error) {
 			}
 		}
 
+		err = errors.Errorf("Error validating write permission to %s. %s", tag, err.Error())
```
matthewmcnew, line 42: This should be an error wrap. 

`errors.Wrapf(err, "message %s", thing)`

matthewmcnew:
Right now we have two mechanisms to print errors to the user. Either the `HasWriteAccess` returns false or we get an unexpected error and print that out. 

I wonder if `HasWriteAccess` should just become `VerifyWriteAccess` that only returns an error and print that error out if it is not nil.

ashwin-venkatesh:
> Right now we have two mechanisms to print errors to the user. Either the `HasWriteAccess` returns false or we get an unexpected error and print that out.
> 
> I wonder if `HasWriteAccess` should just become `VerifyWriteAccess` that only returns an error and print that error out if it is not nil.

im into this!!

sukhil-suresh:
> Right now we have two mechanisms to print errors to the user. Either the `HasWriteAccess` returns false or we get an unexpected error and print that out.
> 
> I wonder if `HasWriteAccess` should just become `VerifyWriteAccess` that only returns an error and print that error out if it is not nil.

Aye. that makes sense. Apply the same approach to `HasReadAccess` check too?

matthewmcnew:
`VerifyReadAccess` sounds like a great idea, @sukhil-suresh !

matthewmcnew:
Awesome!

What does the updated output look like for 401 and 500 cases?

sukhil-suresh:
> Awesome!
> 
> What does the updated output look like for 401 and 500 cases?

I captured the output for 500 in the [commit message](https://github.com/pivotal/kpack/pull/375/commits/e9ba53855f927631d03c8f1b7d24eb727dae9622). Also below:
```
[prepare] Error verifying write access to "dev.registry.pivotal.io/garbage": GET https://dev.registry.pivotal.io/service/token?scope=repository%3Agarbage%3Apush%2Cpull&service=harbor-registry: unsupported status code 500
```
Did not keep a copy of the output for a 401.. But, the message should be prefixed identically `[prepare] Error verifying write access to "dev.registry.pivotal.io/garbage"`

Let me grab the output from an actual run and will update soon.

sukhil-suresh:
Checked against dockerhub with invalid creds
```
```

sukhil-suresh:
also the `dev.registry.pivotal.io/garbage/garbage`, which previously failed with the message `[prepare] invalid credentials to build to dev.registry.pivotal.io/garbage/garbage`

now fails with the consistent prefix and a lot more verbose error message. The additional error message was previously lost behind the `false` boolean, nil error return 

```
[prepare] Error verifying write access to "dev.registry.pivotal.io/garbage/garbage": POST https://dev.registry.pivotal.io/v2/garbage/garbage/blobs/uploads/: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:garbage/garbage Type:repository] map[Action:push Class: Name:garbage/garbage Type:repository]]
```

matthewmcnew:
I think we can merge.

I know in the past we explicitly made the decision to not show the entire verbose error message on common 401s because it is a bit confusing message with information they don't need to understand. Do you think that extra information is valuable in the 401 case?

sukhil-suresh:
> I know in the past we explicitly made the decision to not show the entire verbose error message on common 401s because it is a bit confusing message with information they don't need to understand. Do you think that extra information is valuable in the 401 case?

Sorry, was not aware of the preference to limit the verbose error message!
The original error was wrapped and bubbled up with the `Unauthorized` message, as part of dropping the boolean return. 

From what, I understand it's a case of seeing
```
[prepare] Error verifying write access to "index.docker.io/sukhilsuresh/spring-petclinic": Unauthorized
```
vs
```
[prepare] Error verifying write access to "index.docker.io/sukhilsuresh/spring-petclinic": Unauthorized: GET https://auth.docker.io/token?scope=repository%3Asukhilsuresh%2Fspring-petclinic%3Apush%2Cpull&service=registry.docker.io: unsupported status code 401`
```

To be honest, I like seeing as much detail as possible :) 
Not strongly opinionated about it, so I am very open to making the error less verbose. Let me know, very minor change - drop the wrapping of error @ https://github.com/pivotal/kpack/blob/issue_357/pkg/dockercreds/access_checker.go#L33-L38

sukhil-suresh:
Made error messages less verbose for known errors

How errors look now:

- 500
```
[prepare] Error verifying write access to "dev.registry.pivotal.io/garbage": GET https://dev.registry.pivotal.io/service/token?scope=repository%3Agarbage%3Apush%2Cpull&service=harbor-registry: unsupported status code 500
```

- 401
```
[prepare] Error verifying write access to "index.docker.io/sukhilsuresh/spring-petclinic": UNAUTHORIZED
```

- harbor special case (diagnosed as unauthorized)
```
[prepare] Error verifying write access to "dev.registry.pivotal.io/garbage/garbage": UNAUTHORIZED
```

sukhil-suresh:
> harbor special case (diagnosed as unauthorized)
[prepare] Error verifying write access to "dev.registry.pivotal.io/garbage/garbage": UNAUTHORIZED

I know it's merged, but noticed when testing (again) that the above statement is inaccurate.  The harbor special case `dev.registry.pivotal.io/garbage/garbage` behaves same as the prior comment. 
```
[prepare] Error verifying write access to "dev.registry.pivotal.io/garbage/garbage": POST https://dev.registry.pivotal.io/v2/garbage/garbage/blobs/uploads/: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:garbage/garbage Type:repository] map[Action:push Class: Name:garbage/garbage Type:repository]]
```
To be fair, this is bit of an outlier case/error, but still... :(

The 401s and 500 work as expected.

Issue #362, opened 2020-04-23
Images report Ready=True incorrectly on git revision updates. 


matthewmcnew:
When an Image's spec.git.revision gets an update kpack will temporarily update the status to report Ready=True instead of Ready=Unknown. This rogue status update will confuse tooling that is watching for an image update to be properly reconciled. This is because it it is not possible to differentiate this update from a true Ready=True update when no build is needed. 

*Why this currently happens:* An image spec.source update will update the image's source resolver and wait for the source resolver to update. If the source resolver is not ready the image will temporarily not schedule a build but, still report a Read=True state.

PR #377, opened 2020-05-11
Correctly Report Ready=Unknown on Images

matthewmcnew:
- Resolves https://github.com/pivotal/kpack/issues/362
- Remove unnecessary package `reconciler/v1alpha1/` nesting
- Remove `BuildApplier` indirection as that logic can live in the reconciler package. (And was really confusing)

This PR begins to refactor reconciliation/operational logic out of the `pkg/apis/*` packages. 
- Much of this logic is not part of the public api for those resources. 
- This logic could get awkward with multiple versions of a resource
- There is less clarity of domain boundaries if many types are in `pkg/apis` package. 

Open Question: Does it still make sense to have type creation methods like `image.Build()` or `build.pod()`


djoyahoy:
@matthewmcnew as for things like `Image.Build` and `build.pod`, it "feels" like they shouldn't be in the api package. I'll have to think harder as to why 🤔.

PR #397, opened 2020-05-27
Default image cache size

djoyahoy:


cmd/webhook/main.go
```diff
@@ -54,14 +78,31 @@ func defaultingAdmissionController(ctx context.Context, cmw configmap.Watcher) *
 		types,
 		// A function that infuses the context passed to Validate/SetDefaults with custom metadata.
 		func(ctx context.Context) context.Context {
+			storageClasses, err := storageClassLister.List(labels.NewSelector())
```
matthewmcnew, line 81: I really like the simplicity of using the lister here.

cmd/webhook/main.go
```diff
@@ -29,22 +35,40 @@ var types = map[schema.GroupVersionKind]resourcesemantics.GenericCRD{
...
 }
 
+var (
+	storageClassLister v1.StorageClassLister
+)
+
 func main() {
+	restConfig := sharedmain.ParseAndGetConfigOrDie()
+
+	k8sClient, err := kubernetes.NewForConfig(restConfig)
```
matthewmcnew, line 45: This is bizarre but, could we utilize knative pkg's injection here so that we have just one way of setting up informers?

PR #398, opened 2020-05-27
Apply wait

djoyahoy:
- ImageWaiter is intended to be imported and used in kpack tooling
- Update Build logs to support tailing a build name
- Allow build logs to exit on build completion

PR #405, opened 2020-06-03
Remove eye-catcher to match pack build output

matthewmcnew:
- pack was recently updated to no longer display the eye catcher prefix on log lines. (https://github.com/buildpacks/pack/pull/653)
- this will match pack log output.

### DIFF
--- a/pkg/logs/build_logs.go
+++ b/pkg/logs/build_logs.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


matthewmcnew:
Previously: 
Screen Shot 2020-06-03 at 3 44 03 PM

With this update:
Screen Shot 2020-06-03 at 3 43 43 PM

matthewmcnew:
> Just heads up both this and #398 have both of these commits, besides that looks good

Yep, #398 should be merged first. I just wanted a separate discussion on this one.

PR #420, opened 2020-06-23
Source resolver s3

mtatheonly:
Adds support to pull code from a S3-like object storage

### DIFF
--- a/cmd/build-init/main.go
+++ b/cmd/build-init/main.go
@@ -17,6 +17,7 @@ import (
 	"github.com/pivotal/kpack/pkg/flaghelpers"
 	"github.com/pivotal/kpack/pkg/git"
 	"github.com/pivotal/kpack/pkg/registry"
+	"github.com/pivotal/kpack/pkg/s3"
 )
 
 var (
@@ -24,10 +25,17 @@ var (
 	imageTag        = flag.String("imageTag", os.Getenv("IMAGE_TAG"), "tag of image that will get created by the lifecycle")
 	runImage        = flag.String("runImage", os.Getenv("RUN_IMAGE"), "run image that the build the image on")
 
-	gitURL        = flag.String("git-url", os.Getenv("GIT_URL"), "The url of the Git repository to initialize.")
-	gitRevision   = flag.String("git-revision", os.Getenv("GIT_REVISION"), "The Git revision to make the repository HEAD.")
-	blobURL       = flag.String("blob-url", os.Getenv("BLOB_URL"), "The url of the source code blob.")
-	registryImage = flag.String("registry-image", os.Getenv("REGISTRY_IMAGE"), "The registry location of the source code image.")
+	gitURL           = flag.String("git-url", os.Getenv("GIT_URL"), "The url of the Git repository to initialize.")
+	gitRevision      = flag.String("git-revision", os.Getenv("GIT_REVISION"), "The Git revision to make the repository HEAD.")
+	blobURL          = flag.String("blob-url", os.Getenv("BLOB_URL"), "The url of the source code blob.")
+	registryImage    = flag.String("registry-image", os.Getenv("REGISTRY_IMAGE"), "The registry location of the source code image.")
+	s3Url            = flag.String("s3-url", os.Getenv("S3_URL"), "The url of the S3 server")
+	s3Bucket         = flag.String("s3-bucket", os.Getenv("S3_BUCKET"), "The bucket that has the source code")
+	s3File           = flag.String("s3-file", os.Getenv("S3_FILE"), "The source code file")
+	s3AccessKey      = flag.String("s3-access-key", os.Getenv("S3_ACCESS_KEY"), "The S3 access key")
+	s3SecretKey      = flag.String("s3-secret-key", os.Getenv("S3_SECRET_KEY"), "The S3 secret key")
+	s3ForcePathStyle = flag.String("s3-force-path-style", os.Getenv("S3_FORCE_PATH_STYLE"), "The S3 secret key")
+	s3Region         = flag.String("s3-region", os.Getenv("S3_REGION"), "The S3 secret key")
 
 	basicGitCredentials     flaghelpers.CredentialsFlags
 	sshGitCredentials       flaghelpers.CredentialsFlags
@@ -150,8 +158,13 @@ func fetchSource(logger *log.Logger, serviceAccountCreds dockercreds.DockerCreds
 			Keychain: authn.NewMultiKeychain(imagePullSecrets, serviceAccountCreds),
 		}
 		return fetcher.Fetch(appDir, *registryImage)
+	case *s3Url != "":
+		fetcher := s3.Fetcher{
+			Logger: logger,
+		}
+		return fetcher.Fetch(appDir, *s3Url, *s3AccessKey, *s3SecretKey, *s3Bucket, *s3File, *s3ForcePathStyle, *s3Region)
 	default:
-		return errors.New("no git url, blob url, or registry image provided")
+		return errors.New("no git url, blob url, s3 url, or registry image provided")
 	}
 }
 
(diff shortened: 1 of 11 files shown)


cmd/build-init/main.go
```diff
@@ -17,17 +17,25 @@ import (
...
+	gitURL           = flag.String("git-url", os.Getenv("GIT_URL"), "The url of the Git repository to initialize.")
+	gitRevision      = flag.String("git-revision", os.Getenv("GIT_REVISION"), "The Git revision to make the repository HEAD.")
+	blobURL          = flag.String("blob-url", os.Getenv("BLOB_URL"), "The url of the source code blob.")
+	registryImage    = flag.String("registry-image", os.Getenv("REGISTRY_IMAGE"), "The registry location of the source code image.")
+	s3Url            = flag.String("s3-url", os.Getenv("S3_URL"), "The url of the S3 server")
+	s3Bucket         = flag.String("s3-bucket", os.Getenv("S3_BUCKET"), "The bucket that has the source code")
+	s3File           = flag.String("s3-file", os.Getenv("S3_FILE"), "The source code file")
+	s3AccessKey      = flag.String("s3-access-key", os.Getenv("S3_ACCESS_KEY"), "The S3 access key")
+	s3SecretKey      = flag.String("s3-secret-key", os.Getenv("S3_SECRET_KEY"), "The S3 secret key")
+	s3ForcePathStyle = flag.String("s3-force-path-style", os.Getenv("S3_FORCE_PATH_STYLE"), "The S3 secret key")
```
tylerphelan, line 37: `The S3 secret key` -> `Use the S3 path-style (V1) request URI`

cmd/build-init/main.go
```diff
@@ -17,17 +17,25 @@ import (
...
+	s3Region         = flag.String("s3-region", os.Getenv("S3_REGION"), "The S3 secret key")
```
tylerphelan, line 38: `The S3 secret key` -> `The S3 region`

pkg/apis/build/v1alpha1/image_validation_test.go
```diff
@@ -6,8 +6,8 @@ import (
 
 	"github.com/sclevine/spec"
 	"github.com/stretchr/testify/assert"
-	"k8s.io/api/core/v1"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
tylerphelan, line 10: use `corev1 "k8s.io/api/core/v1"`

pkg/s3/fetch.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	if err != nil {
+		return err
+	}
+	f.Logger.Printf("Successfully downloaded %s/%s/%s in path %q", blob.Host+blob.Path, s3Bucket, s3File, dir)
+	return nil
+}
+
+func extractTar(reader io.Reader, dir string) error {
```
tylerphelan, line 107: maybe time to extract these 3 methods: `extractTar`, `extractZip`, `classifyFile`

cmd/build-init/main.go
tylerphelan, line 37: Maybe `flag.Bool`

pkg/s3/fetch.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	forcePath, err := strconv.ParseBool(s3ForcePathStyle)
+	if err != nil {
+		forcePath = false
+	}
+
+	region := s3Region
+	if region == "" {
+		region = "us-west-2"
```
tylerphelan, line 44: should we default this server-side? unclear to me why this specific default.

tylerphelan:
Thank you for this PR! We'll look at this and get back soon.

pkg/s3/fetch.go
mtatheonly: Changed to a stub string. This field is required by amazon-sdk

tylerphelan:
@mtatheonly 

Could you explain a little about your use case for this PR:
- How would you use this S3 source type? Would the other source types solve your use case?
- Do you consider this feature a must-have to use `kpack`?
- Anything else you'd like to share about how you're using `kpack`

We opened an issue to track our team's discussions about this feature: https://github.com/pivotal/kpack/issues/429

Our initial thought about this:

The s3 creds (access key and secret key) should be accessed via Secrets in this resolver (meaning the user would have to create this Secret with s3 creds before creating an Image with this source type). There is some precedent for this in the `pkg/git`  package [here](https://github.com/pivotal/kpack/tree/master/pkg/git).

mtatheonly:
Hello @tylerphelan,

I'm helping other team at work and they are evolving a image build service. Today they use s2i and other tools to build images and I already used buildpacks with Heroku and saw a chance to use CNCF buildpacks. Once we have a wish of using k8s clusters to build the images, I thought that kpack would be a great option. 

> How would you use this S3 source type? Would the other source types solve your use case?

I thought of uploading source code to a bucket on a s3-based object storage and using the uploaded file to build the images. I know that we can generate direct links (blob source) to bucket objects, but that link has an expiration and we have to keep that source for audit reasons. A part of projects resides in git repositories (so git source would suffice) but we have ongoing projects using other VCS, so S3 source type would be a great feature for us.

> Do you consider this feature a must-have to use kpack?

As I said, it  would be great to have this feature, but not blocking.

> The s3 creds (access key and secret key) should be accessed via Secrets in this resolver (meaning the user would have to create this Secret with s3 creds before creating an Image with this source type). 

Agreed. I was already thinking about that would be safer to keep the access key and secret key within secrets. I'll try to make appropriate changes in PR.

matthewmcnew:
Hi @mtatheonly!

> I thought of uploading source code to a bucket on a s3-based object storage and using the uploaded file to build the images. I know that we can generate direct links (blob source) to bucket objects, but that link has an expiration and we have to keep that source for audit reasons. A part of projects resides in git repositories (so git source would suffice) but we have ongoing projects using other VCS, so S3 source type would be a great feature for us.

The need to provide source to kpack from other source types make a lot of sense. We usually recommend actually utilizing the [registry source](https://github.com/pivotal/kpack/blob/master/docs/image.md#source-configuration) as alternative to git. This involves uploading the application source itself to the registry. (By doing this we can even leverage some performance benefits the registry can provide). 

I know it is a bit tricky to upload application source to a registry but, going forward we are looking to improve the kpack tooling to make it easier.  Would you be interested in exploring that as an alternative to s3?

mtatheonly:
> @mtatheonly
> 
> Could you explain a little about your use case for this PR:
> 
>     * How would you use this S3 source type? Would the other source types solve your use case?
> 
>     * Do you consider this feature a must-have to use `kpack`?
> 
>     * Anything else you'd like to share about how you're using `kpack`
> 
> 
> We opened an issue to track our team's discussions about this feature: #429
> 
> Our initial thought about this:
> 
> The s3 creds (access key and secret key) should be accessed via Secrets in this resolver (meaning the user would have to create this Secret with s3 creds before creating an Image with this source type). There is some precedent for this in the `pkg/git` package [here](https://github.com/pivotal/kpack/tree/master/pkg/git).

Just pushed the code to configure accesskey/secretkey with k8s secrets =)

mtatheonly:
> Hi @mtatheonly!
> 
> > I thought of uploading source code to a bucket on a s3-based object storage and using the uploaded file to build the images. I know that we can generate direct links (blob source) to bucket objects, but that link has an expiration and we have to keep that source for audit reasons. A part of projects resides in git repositories (so git source would suffice) but we have ongoing projects using other VCS, so S3 source type would be a great feature for us.
> 
> The need to provide source to kpack from other source types make a lot of sense. We usually recommend actually utilizing the [registry source](https://github.com/pivotal/kpack/blob/master/docs/image.md#source-configuration) as alternative to git. This involves uploading the application source itself to the registry. (By doing this we can even leverage some performance benefits the registry can provide).
> 
> I know it is a bit tricky to upload application source to a registry but, going forward we are looking to improve the kpack tooling to make it easier. Would you be interested in exploring that as an alternative to s3?

Sorry for the delay =|

I think it is kind of weird to create a image with source in order to create an application image.

Issue #399, opened 2020-05-28
Expose most recent build reason in Image status


matthewmcnew:
This would make it possible for integrations to know if the "BuildReason" for the most recently built image and respond accordingly. 

Requested by the cloudfoundry capi team. 
cc @piyalibanerjee

matthewmcnew:
Closing with https://github.com/pivotal/kpack/pull/435.

PR #435, opened 2020-07-10
Add LatestBuildReason to ImageStatus

piyalibanerjee:
Related issue for more context: https://github.com/pivotal/kpack/issues/399  

Authored-by: Piyali Banerjee <pbanerjee@pivotal.io>

### DIFF
--- a/pkg/reconciler/image/reconcile_build.go
+++ b/pkg/reconciler/image/reconcile_build.go
@@ -35,6 +35,7 @@ func (c *Reconciler) reconcileBuild(image *v1alpha1.Image, latestBuild *v1alpha1
 			BuildCounter:               nextBuildNumber,
 			BuildCacheName:             buildCacheName,
 			LatestBuildRef:             build.BuildRef(),
+			LatestBuildReason:          build.BuildReason(),
 			LatestImage:                image.LatestForImage(latestBuild),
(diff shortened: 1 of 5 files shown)


pkg/reconciler/image/reconcile_build.go
```diff
@@ -35,6 +35,7 @@ func (c *Reconciler) reconcileBuild(image *v1alpha1.Image, latestBuild *v1alpha1
 			BuildCounter:               nextBuildNumber,
 			BuildCacheName:             buildCacheName,
 			LatestBuildRef:             build.BuildRef(),
+			LatestBuildReason:          build.BuildReason(),
```
matthewmcnew, line 38: I think we need the LatestBuildReason populated when a build is not scheduled. (ConditionUnknown and ConditionFalse cases)

piyalibanerjee: Got it -- I'll add it to the ConditionFalse case (looks like ConditionUnknown is just a fallthrough). Thanks!

piyalibanerjee: @matthewmcnew - I added the LatestBuildReason to the ConditionFalse case in the newest commit.

matthewmcnew:
@piyalibanerjee This is great! Thanks!

PR #439, opened 2020-07-14
**POTENTIAL BREAKING CHANGES** RFC: Rename the CustomBuilder Resource. Remove the Builder Resource.

matthewmcnew:
[Readable](https://github.com/matthewmcnew/kpack/blob/rfc-remove-builder/rfcs/0000-remove-builders.md)

### DIFF
--- a/rfcs/0000-remove-builders.md
+++ b/rfcs/0000-remove-builders.md
@@ -1,0 +1,34 @@
+Today, kpack has a complicated hierarchy of builder resources (Builder & ClusterBuilder & CustomBuilder & CustomClusterBuilder). Navigating this hierarchy is hard to explain and has consistently been confusing in early feedback. This is complicated by the recommended kpack experience being "CustomBuilders" when it appears that the obvious default is "Builder" or "ClusterBuilder".
+
+The existing "Builder" relies on a polling mechanism to monitor an external Builder for buildpack and stack updates. External polling mechanisms in core kpack types have been a source of confusion that contradicts the pure declarative philosophy outlined in the [kpack monitors rfc](https://github.com/pivotal/kpack/pull/433).
+
+The "Builder's" limited control makes it difficult for users to actualize and understand kpack's functionality and make it a poor onboarding experience for new kpack community members. This discourages users from utilizing the extensible (Custom)Builder resource. Whereas onboarding with the explicit Stack, Store, Builder resources should make it easier to understand kpack's re-build logic. If we removed the existing Builder resource to invest in simplified CustomBuilder tooling and documentation we could focus on providing the best-unified experience for users and remove the toil that Builders create.
+
+Finally, for users to get the most value out of kpack they will eventually need to utilize the (Custom)Builder toolchain to leverage the granular control of buildpacks and stack images. For users to do that the cloud-native buildpacks community will need to have a robust ecosystem of buildpackages and stack images. We can encourage this community migration by discouraging or removing our existing Builder support.
+
+## Outcomes:
+- Reduce kpack onboarding confusion that results from the need to choose the right builder kind.
+- Encourage kpack users to utilize CustomBuilders which will provide the best kpack onboarding experience.
+- Simplify a future kpack cli from needing to use the terms `custom-cluster-builder` and `cluster-builder`.
+- Reduce the maintenance overhead of maintaining both sets of Builder resources. 
+
+## Actions to take:
+* Rename the CustomBuilder to Builder and ClusterCustomBuilder to ClusterBuilder and remove the Builder/ClusterBuilder resource.
+* Provide a kpack cli with functionality that initializes a set of ~~(Custom)~~Builders, Stacks, Stores to make onboarding and setup as simple as the previous builder experience.
+* We will consider recreating Builders in the eventual kpack monitors project.   
+
+## Complexity:
+There is already [active](https://github.com/pivotal/kpack/pull/437) [PRs](https://github.com/pivotal/kpack/pull/434) with major API renames. This is an ideal release window to include another major breaking api change.     
+
+ ## Prior Art:
+N/A 
+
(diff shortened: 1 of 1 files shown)


rfcs/0000-remove-builders.md
```diff
@@ -0,0 +1,34 @@
...
+
+ ## Prior Art:
+N/A 
+
+## Alternatives:
+
+The Builder could be renamed to something like "PreBuilt" Builder. This will allow users to continue utilizing external Builders. However, this approach does not resolve the underlying problems with the Builder concept and may continue to encourage users to use the PreBuiltBuilder.
+
+
+## Risks:
```
tylerphelan, line 31: This has been brought up by @djoyahoy – one risk is that the `Builder` is a [Cloud Native Buildpacks concept](https://buildpacks.io/docs/operator-guide/create-a-builder/) (however not a spec concept) and is an artifact that can be created and used by the Cloud Native Buildpacks `pack` cli project.

This means the builders created by `pack` can be used in `kpack` and makes builders a natural transition path from using `pack` cli to using `kpack`. Removing the `Builder` resource could make this transition tougher.

rfcs/0000-remove-builders.md
```diff
@@ -0,0 +1,34 @@
+Today, kpack has a complicated hierarchy of builder resources (Builder & ClusterBuilder & CustomBuilder & CustomClusterBuilder). Navigating this hierarchy is hard to explain and has consistently been confusing in early feedback. This is complicated by the recommended kpack experience being "CustomBuilders" when it appears that the obvious default is "Builder" or "ClusterBuilder".
```
tylerphelan, line 1: I agree that the overhead and confusion of having two resources is a significant concern.

thephw:
I remember circling back a bunch of times re-reading the docs before settling on what is presently `CustomBuilder`. I am a little curious about the use-case for the `ClusterCustomBuilder`. I sort of went back and forth on which to use, but opted for a `CustomBuilder` given I had no use case to expose that functionality cluster-wide. 

If we can't ponder a valid use case for the ClusterCustomerBuilder it could be another decision point to axe to simplify initial adoption.

rfcs/0000-remove-builders.md
sukhil-suresh: If we do switch to renaming `CustomBuilder` > `Builder`, it would end up with the term `Builder` meaning 2 different things (as pointed out by Tyler above) in the CNB context and kpack context. And, considering kpack is a CNB platform implementation, this may not be ideal?

matthewmcnew: > This means the builders created by pack can be used in kpack and makes builders a natural transition path from using pack cli to using kpack. Removing the Builder resource could make this transition tougher.

I am hopeful that our kpack cli will be able to help make the transition from pack to kpack smoother and easier for new users. 

I also would conjecture that it is more important to optimize for onboarding with (Custom)Builders in kpack than optimizing for onboarding in general.

matthewmcnew: >If we do switch to renaming CustomBuilder > Builder, it would end up with the term Builder meaning two different things (as pointed out by @tylerphelan) in the CNB context and kpack context. And, considering kpack is a CNB platform implementation, this may not be ideal?

I'll need to think about this a bit more but, it is not quite two different things because the (Custom)Builder concept in kpack is broadly equivalent to `pack create-builder` and kpack (Custom)Builder image are usable as a builder with pack. So, it is more that pack created builders are not directly useable with kpack.

matthewmcnew:
That is a good question @thephw. I think the thought is that a Custom(Cluster)Builders are useable in a multi-namespace/multi-tentant environment where a cluster administer would like to create and manage a builder for multiple different namespaces. 

It seems like our docs should recommend starting with namespaced resources. Do you think you would ever use kpack in a multi-namespace environment?

thephw:
@matthewmcnew We really use the build artifacts (built images) across namespaces, but not the whole kpack build process itself. We have the build process isolated since buildpacks need source access and we build images for multiple tenants of the service. In our usage, namespaces and other k8s security policies help avoid having as many attack vectors for multi-tenant builds.

matthewmcnew:
@thephw Doest that mean all kpack images are in one namespace?

sampeinado:
I like this idea a lot, but I would like to add on that I think we should remove the separation between: Builder and Clusterbuilder. It feels very clunky to me. 
Why do they need to be separate at all? Couldn't the scope of the builder be part of its configuration?

sampeinado:
This may be implicitly part of the RFC and I'm missing it, but I understood the intention of the static "Builder" to be intended partially as a convenience for new users who just wanted to learn the tool and weren't particularly picky about how their app got built (i.e. didn't need to customize or optimize the build process). How does this change affect this use case? Will there still be a default builder to get started with?

matthewmcnew:
> I like this idea a lot, but I would like to also raise the possibility that we remove ClusterBuilder as an independent thing. Couldn't the scope of the builder (namespace or cluster) be part of its configuration, a property of a Builder, instead of making it a separate entity with its own name?

@sampeinado A resource has a scope (Namespace/Cluster), so it is not possible to for this to be configured in the resource's spec.

matthewmcnew:
>This may be implicitly part of the RFC and I'm missing it, but I understood the intention of the static "Builder" to partly be a convenience for new users who just wanted to learn the tool and weren't particularly picky about how their app got built (i.e. didn't need to customize or optimize the build process). How does this change affect this use case? Will there still be a default builder to get started with?

@sampeinado kpack will provide new tooling to make it easy to easily get setup. This is likely possible with an import functionality on the kpack cli that can automatically setup the necessary resources.

matthewmcnew:
Merging. This will be released in the 0.1.0 release of kpack. Please submit a future RFC if you would like to take kpack in a different direction.

PR #440, opened 2020-07-14
Error when blob url cannot be fetched

### LINKED ISSUE #426: Prepare step should error when a source blob is inaccessible or invalid
## What Happens
Currently, when an inaccessible or invalid blob url is provided to the prepare step the prepare step will complete without error and the build will fail when detect runs against an empty workspace. 

This is because when the prepare step fetches a provided source blob url it will attempt to unpack the blob even if a non 200 status code is returned. This is additionally problematic because the fetch will [ignore unexpected file types](https://github.com/pivotal/kpack/blob/e075180cb39b2898928973d0469fba9fc2172c4e/pkg/blob/fetch.go#L48) and exit with a success. 

## Reproduce
* Create an image with a `spec.source.blob.url` of something that does not exist like: `https://storage.googleapis.com/INVALID.jar`
* See this errored build: 

```
===> PREPARE
Loading secret for "****" from secret "***" at location "/var/build-secrets/***"
Successfully downloaded storage.googleapis.com/INVALID.jar in path "/workspace"
===> DETECT
ERROR: No buildpack groups passed detection.
ERROR: Please check that you are running against the correct path.
ERROR: failed to detect: no buildpacks participating
```

tylerphelan:
Resolves #426

### DIFF
--- a/pkg/blob/fetch.go
+++ b/pkg/blob/fetch.go
@@ -11,6 +11,8 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+
+	"github.com/pkg/errors"
 )
 
 type Fetcher struct {
@@ -28,6 +30,10 @@ func (f *Fetcher) Fetch(dir string, blobURL string) error {
 		return err
 	}
 	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return errors.Errorf("failed to get blob %s", blobURL)
+	}
 
 	file, err := ioutil.TempFile("", "")
 	if err != nil {
--- a/pkg/blob/fetch.go
+++ b/pkg/blob/fetch.go
@@ -15,6 +15,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+var unexpectedBlobTypeError error = errors.New("unexpected blob file type, must be one of .zip, .tar.gz, .tar, .jar")
+
 type Fetcher struct {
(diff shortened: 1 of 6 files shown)


matthewmcnew:
Part of what made this bug so mysterious is that we silently exit if an unexpected fileType appears. Could we update this PR to also exit when an unexpected fileType is provided even if it returns a 200 status code?

pkg/blob/fetch.go
```diff
@@ -65,7 +73,18 @@ func (f *Fetcher) Fetch(dir string, blobURL string) error {
 		if err != nil {
 			return err
 		}
+
+		if !isTar(r){
```
tylerphelan, line 77: i like keeping `isTar` outside of `extractTar` because
- `isTar` requires a `Seek(0,0)` to reset the reader but if we use a `ReadSeeker` as the arg to `extractTar` we lose the ability to use `gzip.NewReader(r)` as the reader which is nice to have.
- I think `isTar` is meant as a last check validation that the file is the type we expect, which I think is a different responsibility than validating that the input to `extractTar` is a tar file. we're able to cleanly handle the `unexpectedBlobTypeError` in this way.

tylerphelan: @matthewmcnew

matthewmcnew: That makes sense.

PR #451, opened 2020-07-22
**Breaking Change** Move resources to kpack.io api group

matthewmcnew:
- All "experimental" resources moved to kpack.io group
- Removed the "Builder" & "CustomBuilder" resources
- Renamed (Cluster)CustomBuilder to (Cluster)Builder


tylerphelan:
@matthewmcnew Amazing rename, I know this can be really tricky. See some suggestions in this PR off your branch: https://github.com/pivotal/kpack/pull/452

matthewmcnew:
@tylerphelan I attempted to incorporate your updates from this PR: https://github.com/pivotal/kpack/pull/452

matthewmcnew:
@tylerphelan Good catch! Updated.

matthewmcnew:
Image Validation tests are failing. Will fix shortly.

config/builder.yaml
```diff
@@ -1,19 +1,19 @@
...
   names:
     kind: Builder
     singular: builder
     plural: builders
     shortNames:
-    - cnbbuilder
-    - cnbbuilders
     - bldr
-    - bldrs
+    - custombuilder
```
djoyahoy, line 14: These seem unnecessary. While we have to opportunity can we just get rid of the short names?

config/clusterbuilder.yaml
```diff
@@ -1,16 +1,18 @@
...
+  group: kpack.io
   version: v1alpha1
   names:
     kind: ClusterBuilder
     singular: clusterbuilder
     plural: clusterbuilders
     shortNames:
-    - clstbldr
+    - ccb
+    - customclusterbuilders
```
djoyahoy, line 14: Would love to remove these as well, or have the only short names be `cb` and `cbs`.

pkg/apis/build/v1alpha1/builder_validation_test.go
```diff
@@ -6,118 +6,111 @@ import (
...
 )
 
-func TestBuilderValidation(t *testing.T) {
-	spec.Run(t, "Builder Validation", testBuilderValidation)
+func TestClusterBuilderValidation(t *testing.T) {
+	spec.Run(t, "Custom Cluster Builder Validation", testClusterBuilderValidation)
 }
 
-func testBuilderValidation(t *testing.T, when spec.G, it spec.S) {
+func testClusterBuilderValidation(t *testing.T, when spec.G, it spec.S) {
```
djoyahoy, line 18: This seems to be testing a regular builder and not a cluster builder?

pkg/apis/build/v1alpha1/cluster_builder_validation_test.go
```diff
@@ -11,18 +11,18 @@ import (
 	"knative.dev/pkg/apis"
 )
 
-func TestCustomBuilderValidation(t *testing.T) {
-	spec.Run(t, "Custom Builder Validation", testCustomBuilderValidation)
+func TestBuilderValidation(t *testing.T) {
```
djoyahoy, line 14: This should prolly say `TestClusterBuilderValidation`.

config/builder.yaml
matthewmcnew: Hmm. I left those in so if users have the habit of typing `kubectl get custombuilder` it still works. However, maybe we should just rip the band-aide off.

djoyahoy: I'm in favor of ripping off the band-aid. This change is already breaking, I figured it would be a good time to pile on some more breaking changes.

matthewmcnew: Cool cool. I think I agree. Will update.

PR #492, opened 2020-08-28
Reason Message Diff RFC

sampeinado:
[Readable](https://github.com/sampeinado/kpack/blob/master/rfcs/0001-reason-message-diff.md)


rfcs/0001-reason-message-diff.md
```diff
@@ -0,0 +1,53 @@
+### Context:
+
+When kpack produces new builds today, a Reason is given which may be one of five values: COMMIT, TRIGGER, CONFIG, STACK, BUILDPACK
```
matthewmcnew, line 3: It is worth noting that a build may be multiple of those values. As in a build could have `COMMIT,CONFIG`. In extreme cases it may even have them all `COMMIT,CONFIG,STACK,BUILDPACK`

rfcs/0001-reason-message-diff.md
```diff
@@ -0,0 +1,53 @@
...
+
+How does the user compare two non-adjacent builds? ex. builds 17 and 20
+
+Users still have to do a significant amount of digging work to make this information actionable. For example, one developer told me he has to look through the Paketo buildpacks commits and/or diff the dpackage lists of the stacks to see what changed in order to decide whether he should cut a new release based on the new build. 
+
+
+
+### Alternatives:
+
+- Create a new diff command to encapsulate this functionality
```
matthewmcnew, line 52: Could you explain the drawback vs some of these other approaches?

rfcs/0001-reason-message-diff.md
```diff
@@ -0,0 +1,53 @@
...
+**For Reason "COMMIT"**
+- Ex. Reason: COMMIT - Received a new commit with ID [truncated-SH] from branch [branch-name]. 
+
+**For Reason "CONFIG"**
+- Ex. Reason: CONFIG - The [image-name] image config was updated. See diff below.
+[output of diffing library on the two image config files]
+
+**For Reason "BUILDPACK"**
+- Ex. Reason: BUILDER - The buildpacks used by your app have changed. See diff below. 
+[output of diffing library on the two Bill of Materials' for the builds]
```
matthewmcnew, line 30: I don't think Bill of Materials is appropriate for explaining a Build Reason of Buildpack. (As the Bill of Materials is not available until the build successfully completes and may have changes unrelated to the buildpack changes)

rfcs/0001-reason-message-diff.md
```diff
@@ -0,0 +1,53 @@
+### Context:
+
+When kpack produces new builds today, a Reason is given which may be one of five values: COMMIT, TRIGGER, CONFIG, STACK, BUILDPACK
+
+These one-word values are helpful for providing a basic level of context on why a new build has been created, but they don't provide enough information to inform users' decisions about what to do with the resulting build. They may be asking themselves: "Does this new build have some update that affects my app? Should I run my regression tests on it? Should I promote it to prod?"
```
matthewmcnew, line 5: It seems like we are targeting two different use cases here. (Why is a build kicked off & what is different about a resulting image). 

The "Why is a build kicked off" question is one that seems like it could be solved by expanding the BuildReason. However, the "what is different about a resulting image" seems like it could be better solved disconnected from the BuildReason.

matthewmcnew: > However, the "what is different about a resulting image" seems like it could be better solved disconnected from the BuildReason.

A specific situation I am thinking about for this is when a new COMMIT build leads to the change in a dependency version provided by a buildpack such as the JVM.

djoyahoy: I'm very interested in a `kp image diff <build-A> <build-B>` command and I'm wondering if that is a useful place for this feature. Do we think users have a desire to diff any two builds for an image? The default could simply be the two most recent builds.

rfcs/0001-reason-message-diff.md
djoyahoy: I am really interested in a diff command (https://github.com/pivotal/kpack/pull/492#discussion_r479462612).

rfcs/0001-reason-message-diff.md
```diff
@@ -0,0 +1,53 @@
...
+### Goals:
+
+Propose an additional attribute, “Reason-Message” be added to the build object, which can be printed in the context of the existing “Reason” attribute. Contextual information can be provided in the Reason-Message and a diff may also be printed as appropriate. This “Reason-Message” would be printed for these two commands:
+
+$ kp build status image-name
+$ kp image status image-name
+
+The format in which it would be printed:
+
+Reason: [Reason-Code] - [Reason-Message]
```
matthewmcnew, line 18: I know that the exact diff we choose is dependent on the implementation but, I would be curious to mock up what it could look like for complicated Reasons (like multiple reasons).

rfcs/0001-reason-message-diff.md
mgibson1121: I think this interface could be an elegant way to handle both build diff and more details on the reason, but there is a risk here around edge cases either bloating the interface or causing it to break or not be useful in certain cases.  To me the next step is to figure out if it is sensible to handle both of these use cases together.  If no, then maybe focus on the status first and submit a different RFC that presents an interface to handle build diffs.

mgibson1121: @djoyahoy I think one of the reasons why I find the UX of this RFC attractive is that the user wouldn't have to specify the builds they want to compare against.  The metadata about what changed from the last build would be available on every build.  Preserving that UX I think would be ideal rather than juggling build numbers.

rfcs/0001-reason-message-diff.md
```diff
@@ -0,0 +1,53 @@
...
+[output of diffing library on the two Bill of Materials' for the builds]
+
+**For Reason "STACK"**
+- Ex. Reason: STACK - Stack-name changed from dockerhub/sha1 to dockerhub/sha2
+
+**For Reason "TRIGGER"**
+- Ex. Reason: TRIGGER - A new build was manually triggered by [user] at [time]
+
+### Prior Art:
+We have been thinking about different ways to surface the [Bill of Materials](https://github.com/buildpacks/pack/issues/378) for a build and how to show the delta between builds for a few months now. This is an evolution of that work. Although the `pack` cli renders the contents of a single BoM alone with the `inspect-image` command, I believe we don't need to provide a BoM alone, because the purpose of kpack is to iteratively build images, and therefore build BoMs should be placed in context as much as possible.
```
mgibson1121, line 39: +1 on trying to figure out a way to curate the metadata instead of just dumping it.

rfcs/0001-reason-message-diff.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Complexity/Risks:
+How does the output render when there are two Reasons for a new build?
```
mgibson1121, line 42: Couldn't they be stacked on top of each other ordered by their priority?  The priority logic already exists because we use it to determine `REASON` primary in the `kp build list` table

rfcs/0001-reason-message-diff.md
```diff
@@ -0,0 +1,53 @@
...
+
+How does the user compare two non-adjacent builds? ex. builds 17 and 20
```
mgibson1121, line 44: I'm curious about how common this use case is.  I've been assuming that the more common use case is to determine the diff between subsequent builds.  This is something that we could implement along the lines of Danny's comment above `kp image diff <build-A> <build-B>` as a follow along feature.  Or if it equally as desired as subsequent builds, it could become the primary interface for diffing builds.

mgibson1121:
Nice job @sampeinado!  I think the critical think you pushed forward is forcing us to consider if we want to separate "build diff" functionality as its own interface or modify `kp build status` to include it.

rfcs/0001-reason-message-diff.md
sampeinado: Yes definitely, i tried to reference that in the complexity section

sampeinado:
I updated the RFC to include the feedback on this thread and from two users I spoke with. 

Adding colored mockups for context:
Frame 1 (4)
Frame 2 (2)
Frame 3 (2)
Frame 4 (1)

rfcs/0001-reason-message-diff.md
```diff
@@ -0,0 +1,292 @@
...
+
+**For Reason "TRIGGER"**
+
+- Print the name of the user and the time () it was triggered as detail for the reason
+- No diff
+
+------------------------------
+
+$kp build status petclinic
+Image:      index.docker.io/username/instance@sha256:fsdgh39478th0g9tuisodfhgvns938e54iunwuiehfasf9wgiauwsfx
```
matthewmcnew, line 60: Could you put these in ``` to properly format them in the markdown readable view?

matthewmcnew (approved):

rfcs/0001-reason-message-diff.md
```diff
@@ -2,52 +2,291 @@
...
-$ kp image status image-name
 
-The format in which it would be printed:
+See mockups below. 
+
+### Prior Art:
+We have been thinking about different ways to surface the [Bill of Materials](https://github.com/buildpacks/pack/issues/378) for a build and how to show the delta between builds for a few months now. This is an evolution of that work. Although the `pack` cli renders the contents of a single BoM alone with the `inspect-image` command, I believe we don't should provide the BoM in this case, because this feature is intended to answer questions around WHY a build kicked off and whats different about the causes of the build, not the contents. It's important to address cause and contents separately because the dependencies involved are not 1:1 related.  A true diff of the contents would be better done using a different feature or standalone tool. 
+
+### Complexity/Risks:
+How does the user compare two non-adjacent builds? ex. builds 17 and 20
```
mgibson1121, line 23: Is this still a risk if you're de-scoping the build diff?

rfcs/0001-reason-message-diff.md
```diff
@@ -2,52 +2,291 @@
...
-Reason: [Reason-Code] - [Reason-Message]
+- Print the name of the user and the time () it was triggered as detail for the reason
+- No diff
+
+------------------------------
+
+$kp build status petclinic
+Image:      index.docker.io/username/instance@sha256:fsdgh39478th0g9tuisodfhgvns938e54iunwuiehfasf9wgiauwsfx
+Status:     SUCCESS
+Reason:     TRIGGER - A new build was manually triggered by [user] at [time].
```
mgibson1121, line 62: Definitely like the idea of identifying the user.

rfcs/0001-reason-message-diff.md
```diff
@@ -2,52 +2,291 @@
...
 
 **For Reason "CONFIG"**
-- Ex. Reason: CONFIG - The [image-name] image config was updated. See diff below.
-[output of diffing library on the two image config files]
 
-**For Reason "BUILDPACK"**
-- Ex. Reason: BUILDER - The buildpacks used by your app have changed. See diff below. 
-[output of diffing library on the two Bill of Materials' for the builds]
+- Print a diff beneath the reason section
+- Print the three lines preceding where the diff begins in order to show context for the change.
```
mgibson1121, line 110: Could it just be simpler to print the config for build A with the `-` symbol next to the fields that have changed and then print the config for build B with the `+` symbol next to the fields that are different from build A?

rfcs/0001-reason-message-diff.md
```diff
@@ -2,52 +2,291 @@
...
-We have been thinking about different ways to surface the [Bill of Materials](https://github.com/buildpacks/pack/issues/378) for a build and how to show the delta between builds for a few months now. This is an evolution of that work. Although the `pack` cli renders the contents of a single BoM alone with the `inspect-image` command, I believe we don't need to provide a BoM alone, because the purpose of kpack is to iteratively build images, and therefore build BoMs should be placed in context as much as possible.
+------------------------------
 
-### Complexity/Risks:
-How does the output render when there are two Reasons for a new build?
+$kp build status petclinic
+Image:      index.docker.io/username/instance@sha256:fsdgh39478th0g9tuisodfhgvns938e54iunwuiehfasf9wgiauwsfx
+Status:     SUCCESS
+Reason:     STACK
+            - RunImage: index.docker.io/paketobuildpacks/run@sha256:34978twgsui4789et0wh9384erhsg0e798riuhgs0er7iugshe98riugeh99rgaser892
```
mgibson1121, line 156: Out of scope here, but makes me think about how valuable it would be to have a stack homepage link so the user could review what changed about the stacks rather than just see the diff in the digest.

mgibson1121 (commented):
Nice work!  Left a few comments

mgibson1121 (approved):

rfcs/0001-reason-message-diff.md
tylerphelan: I agree I think it should be the whole config because if multiple config values change it would be hard to satisfy this constraint:

> Print the three lines preceding where the diff begins in order to show context for the change. 

I think we should just print the whole config but I think it should have a line-by-line diff so that changes are clear, not both configs listed entirely:

```
  build:
    env:
    - name: BP_JAVA_VERSION
-     value: 11.*
+     value: 14.*
- resources: {}
+ resources:
+   limits:
+     cpu: 1M
- CacheSize: 1M
+ CacheSize: 12M
  unchanged: value
  rest:
    of:
      config: foo
```

rfcs/0001-reason-message-diff.md
```diff
@@ -0,0 +1,306 @@
...
+```
+$ kp build status image-name
+```
+
+See mockups below. 
+
+### Prior Art:
+We have been thinking about different ways to surface the [Bill of Materials](https://github.com/buildpacks/pack/issues/378) for a build and how to show the delta between builds for a few months now. This is an evolution of that work. Although the `pack` cli renders the contents of a single BoM alone with the `inspect-image` command, I believe we don't should provide the BoM in this case, because this feature is intended to answer questions around WHY a build kicked off and whats different about the causes of the build, not the contents. It's important to address cause and contents separately because the dependencies involved are not 1:1 related.  A true diff of the contents would be better done using a different feature or standalone tool. 
+
+### Complexity/Risks:
```
tylerphelan, line 24: Now that the build status output is getting bigger I think we could hide these build reason details behind a `--verbose` flag. The first build will always have a massive diff.

rfcs/0001-reason-message-diff.md
sampeinado: i think the latter

rfcs/0001-reason-message-diff.md
sampeinado: did you do it for me? thank you !

rfcs/0001-reason-message-diff.md
sampeinado: yea totally! 
we should do a brainstorm with the core deps team about it

djoyahoy (approved):
Looks great!

rfcs/0001-reason-message-diff.md
kvedurmu: +1. Would be super cool if users could see the dpkg output difference

rfcs/0001-reason-message-diff.md
```diff
@@ -0,0 +1,322 @@
...
+
+- Print a diff as part of the buildpacks table 
+
+------------------------------
+
+```
+$kp build status petclinic
+Image:      index.docker.io/username/instance@sha256:fsdgh39478th0g9tuisodfhgvns938e54iunwuiehfasf9wgiauwsfx
+Status:     SUCCESS
+Reason:     BUILDPACK
```
kvedurmu, line 209: Will this buildpack diff account for both language family as well as implementation buildpack updates? Currently, kp doesn't show the tree-like buildpack output (similar to `pack inspect-builder`), which makes it really hard to figure out specific buildpacks/dependencies that were updated.

Also, any thoughts on the diff showing the actual dependencies that changed? This would let users determine whether a specific buildpack update is due to upstream CVE, new runtime, etc. (dependency update) or a feature/bug fix in the buildpack itself.

matthewmcnew: > Will this buildpack diff account for both language family as well as implementation buildpack updates? Currently, kp doesn't show the tree-like buildpack output (similar to pack inspect-builder), which makes it really hard to figure out specific buildpacks/dependencies that were updated.

The rebuild logic is agnostic to the meta buildpack hierarchy. It simplistically will schedule a rebuild when the builder no longer has an implementation buildpack that was previously used. So, this output should likely represent that simplicity.  However, `kp builder status` should be updated to show the tree like output.

matthewmcnew: > Also, any thoughts on the diff showing the actual dependencies that changed? This would let users determine whether a specific buildpack update is due to upstream CVE, new runtime, etc. (dependency update) or a feature/bug fix in the buildpack itself.

At this point this diff view largely shows *why* a build was scheduled as a window into the kpack declarative logic. It seems like some other diff-like functionality would be better positioned to show relevant dependency changes. 

However, surfacing this info about what dependency changes happen and why they might matter (CVE, bug fix, etc) somewhere would be very interesting and would require some more buildpack metadata.

kvedurmu: Thanks, that makes sense! Happy to chat about what changes we'd need to make in the buildpacks to support this.

rfcs/0001-reason-message-diff.md
```diff
@@ -0,0 +1,325 @@
...
+- Print the stack-name after the reason in order to give context 
+
+------------------------------
+```
+$kp build status petclinic
+Image:      index.docker.io/username/instance@sha256:fsdgh39478th0g9tuisodfhgvns938e54iunwuiehfasf9wgiauwsfx
+Status:     SUCCESS
+Reason:     STACK
+            - RunImage: index.docker.io/paketobuildpacks/run@sha256:34978twgsui4789et0wh9384erhsg0e798riuhgs0er7iugshe98riugeh99rgaser892
+            + RunImage: index.docker.io/paketobuildpacks/run@sha256:fh34578tg29wb645rwhs87e5hrw0e98jtfnws9e7w378qe49tr3g39n5gwe9u5hge549w
```
piyalibanerjee, line 174: Hello from CAPI! Will this diff be included in the Image CR's status as well? It would be really great to include this stack diff in the Image's status as well to help with the making the progress of stack updates observable. It would also be a valuable thing to add to the logs for the cf-api-controllers (which handles deploying rebased images to the cluster).

matthewmcnew: This information will likely just be metadata placed on build when the builds are scheduled.

matthewmcnew: > It would be really great to include this stack diff in the Image's status as well to help with the making the progress of stack updates observable

Could you explain your usecase a bit more here? Where would this information be surfaced?

matthewmcnew: > It would also be a valuable thing to add to the logs for the cf-api-controllers (which handles deploying rebased images to the cluster).

Is this cf-api-controller's logging?

PR #494, opened 2020-09-01
Set memory and cpu requests and limit values for all containers

paulcwarren:
The CF-for-K8s release integration team, who consume kpack, are currently working on a scaling and Quality of Service (QoS) set of stories.

We are targeting 1.0 to be configured out-of-the-box as a "developer" edition aimed at those users who want to kick the tires.  As part of this, we would like to set limits on mem/cpu.

Since a "developer" edition may not be preferred by everyone, we want each component to be configurable to scale both horizontally (replicas) and vertically (mem/cpu).  This will also allow users to deliver a Guaranteed QoS when required (although we are recommending that all of our pods and containers use the Burstable QoS) As part of this we would like to ask you to do several things:

1. consider which of your pods/containers you would like to expose for scaling properties for.
2. expose said configuration properties.
3. sets mem and cpu values for all containers in order to provide as much meta-data to k8s as possible so that its scheduler can do as good a job as possible. This PR is an initial attempt at setting these values, although we know you are much more likely to have insight into your components mem/cpu requirements than our guess.  

If you have any questions or concerns, please let us know! Thanks! 

[#174462927](https://www.pivotaltracker.com/story/show/174462927)

Co-Authored-By: Angela Chin <achin@pivotal.io>

### DIFF
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -84,3 +84,10 @@ spec:
             configMapKeyRef:
               name: lifecycle-image
               key: image
+        resources:
+          requests:
+            cpu: 10m
+            memory: 50Mi
+          limits:
+            cpu: 100m
+            memory: 200Mi
--- a/config/webhook.yaml
+++ b/config/webhook.yaml
@@ -70,3 +70,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        resources:
+          requests:
+            cpu: 10m
+            memory: 50Mi
+          limits:
+            cpu: 100m
+            memory: 200Mi


matthewmcnew (approved):
I load tested this configuration today and monitored kpack's usage.

tylerphelan (approved):

paulcwarren:
Looks like this was approved 7 days ago.  Is there anything else we can do to help get this merged.  Making all pods Burstable is a 1.0 goal for cf-for-k8s and we are approaching out 1.0 release.  Thanks for your attention.

tylerphelan:
@paulcwarren sorry about the delay! merged manually.

PR #495, opened 2020-09-02
Fix header in Stacks doc

making:
Not about Stores 😅


sukhil-suresh (approved):
Thanks for correcting it, @making !

matthewmcnew (approved):

matthewmcnew:
Good catch @making ! Thanks for the fix!

PR #496, opened 2020-09-08
add a non-tailing version of fetching logs for an image

LukeWinikates:
👋 Hi there. Fellow 'Pivotal' team here.

We're looking to fetch kpack logs and get just the logs available now for the containers currently running. None of the existing `Tail*` functions supported that use case, but we were able to figure out how to accomplish what we needed by adding more boolean flags.

We'd like to see if we can get this merged so that we can eliminate our fork as a dependency. This shouldn't change the behavior of any of the existing functions in the package.

It does start to clutter up the function signatures with boolean flags.

Curious to hear any feedback about the `GetImageLogs` function name - `Tail` doesn't seem appropriate, since we're not trying to tail the logs in this case.

Thanks in advance,
@LukeWinikates  / @ssama-pivotal

### DIFF
--- a/pkg/logs/build_logs.go
+++ b/pkg/logs/build_logs.go
@@ -32,12 +32,19 @@ func (c *BuildLogsClient) Tail(ctx context.Context, writer io.Writer, image, bui
 	return c.tailPods(ctx, writer, namespace, metav1.ListOptions{
 		Watch:         true,
 		LabelSelector: fmt.Sprintf("%s=%s,%s=%s", v1alpha1.ImageLabel, image, v1alpha1.BuildNumberLabel, build),
-	}, true)
+	}, true, true)
 }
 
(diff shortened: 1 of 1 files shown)


tylerphelan (approved):
I like it!

LukeWinikates:
I just pushed an updated version of this commit - the changes in the original version ended up not being sufficient, as `watchReadyContainers` was also waiting for all of the pods to finish their work. The commit I just pushed reflects what we're currently using for Iris. If this code or some version of this feature could be merged, that would be great for the Iris team as we'd be able to eliminate our fork. Thanks for considering!

matthewmcnew (approved):

matthewmcnew:
@LukeWinikates Could you rebase and we will merge?

LukeWinikates:
@matthewmcnew done!

PR #497, opened 2020-09-08
Fix generated openapi spec

djoyahoy:
- Use string type for lastTransitionTime in the Condition schema as it does custom json encoding/decoding
- Fix cluster object template for patches

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ bazel-*
 coverage.txt
 
 vendor
+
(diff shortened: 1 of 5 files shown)


djoyahoy:
Should generating the openAPI spec be documented, and if so, where?

tylerphelan (approved):

matthewmcnew (approved):

PR #500, opened 2020-09-17
Notary V1 Integration RFC

djoyahoy:
[Readable](https://github.com/pivotal/kpack/blob/notary-rfc/rfcs/0001-notary-integration.md)


rfcs/0001-notary-integration.md
```diff
@@ -0,0 +1,50 @@
...
+    rootPrivateKeyRef: <secret-name>
+    rootPublicKeyRef: <secret-name> (may not be necessary)
+    targetPrivateKeyRef: <secret-name>
+    targetPublicKeyRef: <secret-name> (may not be necessary)
+  ```
+2. Add a new container after export that uses the notary go client to sign the image. The presence or absence of the notary section in the image config will determine if the notary signing container is executed.
+
+There are a few different options to consider for the root and target keys:
+
+**Key generation (2 Options)**
```
matthewmcnew, line 22: How will we pick an option? Will we decide before the rfc gets merged or when implementing?

rfcs/0001-notary-integration.md
```diff
@@ -0,0 +1,39 @@
+# Problem
+kpack does not handle image signing.
```
matthewmcnew, line 2: I know it is a bit cumbersome but, can we have a bit more detail on why users might want image signing?

rfcs/0001-notary-integration.md
```diff
@@ -0,0 +1,39 @@
...
+
+# Outcome
+kpack integrates with notary to sign images that it builds.
+
+# Actions to take
+There are a few concrete steps to make this work in kpack:
+1. Add a new section to the image config that allows users to configure notary settings:
+  ```
+  notary:
```
matthewmcnew, line 11: This is `spec.notary` I assume? What would be the pros/cons of nesting this under v1 to better prepare us for future versions of notary?

rfcs/0001-notary-integration.md
```diff
@@ -0,0 +1,39 @@
...
+2. Add a new container after export that uses the notary go client to sign the image. The presence or absence of the notary section in the image config will determine if the notary signing container is executed.
+
+## Key Generation and Storage
+kpack will auto generate the public/private key pair and create new secrets given the secret names in the image config. Auto generation will only happen on the first build of an image. Subsequent builds will use the keys from the secrets that were created. Additionally, kpack will not store the public key, letting the notary server store them instead.
+
+# Complexity
+The actual integration with notary is of low complexity. All of the necessary code is available in the open source docker CLI; particularly the files `cli/command/image/trust.go` and `cli/trust/trust.go`.
+
+# Prior Art
+* The docker CLI implements this feature already.
```
matthewmcnew, line 25: Do we know of any other build solutions that provide notary image signing? Tekton? Cloud Build? Acr task?

rfcs/0001-notary-integration.md
```diff
@@ -0,0 +1,39 @@
...
+
+# Alternatives
+* Implement notary signing directly in the lifecyle.
+* Do not integrate with notary.
+* A solution that signs images outside of kpack entirely (aka. as another step in CI/CD).
```
matthewmcnew, line 30: This is the most obvious alternative for me. What would the consequence be of that? Lose confidence the image was signed by kpack?

rfcs/0001-notary-integration.md
```diff
@@ -0,0 +1,39 @@
...
+# Actions to take
+There are a few concrete steps to make this work in kpack:
+1. Add a new section to the image config that allows users to configure notary settings:
+  ```
+  notary:
+    host:                <notary server url>
+    rootPrivateKeyRef:   <secret-name>
+    targetPrivateKeyRef: <secret-name>
+  ```
+2. Add a new container after export that uses the notary go client to sign the image. The presence or absence of the notary section in the image config will determine if the notary signing container is executed.
```
matthewmcnew, line 16: Does kpack sign both images the tag and the additional timestamp/build-number tag?

rfcs/0001-notary-integration.md
```diff
@@ -0,0 +1,39 @@
...
+* The docker CLI implements this feature already.
+
+# Alternatives
+* Implement notary signing directly in the lifecyle.
+* Do not integrate with notary.
+* A solution that signs images outside of kpack entirely (aka. as another step in CI/CD).
+* kpack could let users provide the public/private key pair instead of generating them.
+* kpack could store both the private key and the public key.
+
+# Risks
```
matthewmcnew, line 34: Could we mention the possible risk about overwriting the tag on each new build?

rfcs/0001-notary-integration.md
djoyahoy: This impl would sign both the image tag and the build tag.

djoyahoy:
Most likely going to create a POC to spike out the implementation before deciding what to do with this RFC.

rfcs/0001-notary-integration.md
ImJasonH: Hi! 👋  I stumbled across this issue, and I can answer some of this question. I work on Tekton and Cloud Build. 😄 

Cloud Build can publish verifiable build information to [Google Container Analysis](https://cloud.google.com/container-registry/docs/container-analysis) (based on [Grafeas](https://github.com/grafeas/grafeas)).

Tekton doesn't publish signatures to Notary, but someone _could_ write a Task to publish signatures to Notary, and there's an experimental project to layer on supply chain security in Tekton natively, called [Chains](https://github.com/tektoncd/chains).

djoyahoy:
The following branch has a rough proof of concept of how kpack might integrate with notary: https://github.com/pivotal/kpack/tree/notary-poc

Some key takeaways:

- Users could supply their own secret refs with private keys so that kpack wouldn't need to generate them
- The image digest could be accessed by the image-signer by mounting the layers and reading the report.toml. The image size could be calculated by the image-singer or added to the report.toml.

Next Steps:

- Investigate authentication with a notary server and how to thread those credentials into kpack.

matthewmcnew:
> Users could supply their own secret refs with private keys so that kpack wouldn't need to generate them.
What format would these keys be in?

djoyahoy:
> > Users could supply their own secret refs with private keys so that kpack wouldn't need to generate them.
> 
> What format would these keys be in?

I believe it's just a reguler PEM format. Users can generate keys with the notary CLI or openssl.

rfcs/0001-notary-integration.md
```diff
@@ -0,0 +1,69 @@
...
+        secretRef: <secret-name>
+  ```
+2. Add a new container after export that uses the notary go client to sign the image. The presence or absence of the notary section in the image config will determine if the notary signing container should be executed.
+3.
+    * The singing container would use `report.toml` to access the image digest. The image size would need to be added to `report.toml` or computed by the signer container.
+      The `report.toml` must exist in its own volume as export does quite a bit of mangling to the layers volume.
+    * Alternatively, the image digest and size could be computed by pulling the image in the signer container.
+
+## Key Generation and Storage
+kpack will not generate the targets public or private key pair used to sign images.
```
tomkennedy513, line 26: does this still require the root key as well, or were you able to drop it?

tomkennedy513 (approved):

rfcs/0001-notary-integration.md
djoyahoy: The root key should not be necessary.

rfcs/0001-notary-integration.md
djoyahoy: I think the loss of confidence and the fact that each CI/CD will probably require a different implementation (perhaps with manual intervention).

rfcs/0001-notary-integration.md
djoyahoy: To me thats a known quantity with Notary, so not necessarily a risk.

rfcs/0001-notary-integration.md
```diff
@@ -0,0 +1,69 @@
...
+## Enable Image Singing
+To use Notary image signing:
+1. Add a new section to the image config that allows users to configure notary settings:
+  ```
+  notary:
+    v1:
+        url:       <notary server url>
+        secretRef: <secret-name>
+  ```
+2. Add a new container after export that uses the notary go client to sign the image. The presence or absence of the notary section in the image config will determine if the notary signing container should be executed.
```
matthewmcnew, line 19: At least something to consider is if this should actually just happen in the `completion` container

rfcs/0001-notary-integration.md
```diff
@@ -0,0 +1,69 @@
...
+
+notary key rotate example.registry.io/my-app snapshot -r
+
+notary publish example.registry.io/my-app
+```
+
+Users must then upload the encrypted target private key and private key password as a generic k8s secret.
+The secret data must be formatted as follows:
+```
+kubectl create secret generic notary-secret --from-file=notary-password.txt --from-file=~/.notary/private/<targets-hash>.key
```
matthewmcnew, line 40: Do the keys matter?

tylerphelan (approved):
This will be very cool!

rfcs/0001-notary-integration.md
djoyahoy: Thats a really nice idea 🤔

rfcs/0001-notary-integration.md
djoyahoy: Yes, the user will have to upload the targets key. The notary CLI indicates which key this is when you setup the repo and it can be found in `~/.notary/private`. Thats some unfortunate leg work for the user, but anyone who is familiar with Notary should have some understanding of how to get that key. If they use docker, the idea is exactly the same, the key will just be in `~/.docker/trust`.

rfcs/0001-notary-integration.md
```diff
@@ -0,0 +1,69 @@
...
+# Alternatives
+* Implement Notary signing directly in the lifecyle.
+* Do not integrate with Notary.
+* A solution that signs images outside of kpack entirely (aka. as another step in CI/CD).
+* Use gpg keys much like Tekton [Chains](https://github.com/tektoncd/chains).
+
+# Risks
+* Notary is only one of the image singing tools available.
+* Notary v2 is on the way.
+* This scheme does not mention the notary CA certificate if there is one.
```
mgibson1121, line 68: Seems likely that Notary would be provisioned with some kind of CA cert.  Are you omitting a solution for this from your reasoning about ultimate kpack implementation?  Or just omitting as part of this scheme to simplify creation of a PoC?

rfcs/0001-notary-integration.md
```diff
@@ -0,0 +1,69 @@
...
+The secret data must be formatted as follows:
+```
+kubectl create secret generic notary-secret --from-file=notary-password.txt --from-file=~/.notary/private/<targets-hash>.key
+```
+
+That secret will be provided via the image config as described above.
+
+## Authenticating with the Notary Server
+If the Notary server requires authentication, users must create and attach a secret with basic auth Notary credentials to the service account used by the image/build config.
+The secret must be properly annotated with the key `kpack.io/notary.v1` and the Notary server host name.
```
matthewmcnew, line 47: These are `kubernetes.io/basic-auth` secrets?

mgibson1121 (approved):

rfcs/0001-notary-integration.md
djoyahoy: Yup, that is correct.

rfcs/0001-notary-integration.md
djoyahoy: I think it's outside the scope of this RFC. Users will have to figure out how to bake the CA cert into the completion image. This is something that they already have to solve with builders and registry CA certs for example.

matthewmcnew (approved):

PR #502, opened 2020-09-18
Explicitly specify namespace for tutorial

harmjanblok:
Thanks for the tutorial. Some steps depend however on the namespace used in the current context. I've added an explicit definition of these namespaces.

Furthermore the `Builder` seems to require the `serviceAccount: tutorial-service-account` instead.


tylerphelan (approved):

tylerphelan:
thank you!

Issue #486, opened 2020-08-26
Blob Source types do not preserve file permissions 


matthewmcnew:
When a blob source tar/zip is extracted the files within the blob are [not created](https://github.com/pivotal/kpack/blob/b1f6db50c2eb300bf95c899125e2a055ef5c466e/pkg/blob/fetch.go#L126) with the same permissions. This is most notable with build system wrappers like mvnw or gradlew which loose executable permissions. 

This was discovered by @ArieShout in https://github.com/pivotal/kpack/issues/445#issuecomment-675254976.

PR #503, opened 2020-09-21
Preserve file mode when untarring blob files

djoyahoy:
#486


pkg/blob/fetch.go
```diff
@@ -118,14 +118,18 @@ func extractTar(reader io.Reader, dir string) error {
 				return err
 			}
 		case tar.TypeReg:
-			outFile, err := os.Create(filePath)
+			outFile, err := os.OpenFile(filePath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, header.FileInfo().Mode())
 			if err != nil {
 				return err
 			}
-			defer outFile.Close()
```
tylerphelan, line 125: why no longer defer?

matthewmcnew (approved):

pkg/blob/fetch.go
djoyahoy: The defer was in a for loop and was getting complaints from the IDE. I suppose the defer gets lost with each iteration, or something equally bad happens when you defer in a for loop. So I just manually close the file.

tylerphelan: I can live with that!

tylerphelan (approved):

PR #506, opened 2020-09-23
Update builder config in tutorial

ameyer-pivotal:
Fix service account in builder config


ameyer-pivotal:
Already done in another PR

PR #507, opened 2020-09-24
Allow source images to contain zip, tar, and tar.gz files

### LINKED ISSUE #491: Allow registry source to handle an image that contains a nonexploded artifact
**Why** 
If a user already has a prebuilt artifact like a war/jar the paketo java buildpacks expect the jar/war to be exploded into the application workspace. Today, If a jar/zip/war/tar is provided as a blob source it will extract the artifact before executing the build. However, a registry source expects artifacts to already be exploded and repackaged in a layer by the tooling that creates the source image. If a user wants to leverage the registry without extracting and repackaging an artifact it would be useful for kpack to extract the artifact from a registry image layer before the build executes. 

**Questions**
* Should kpack identify that the image has an artifact that needs to be expanded with a label as an identifier?

djoyahoy:
* Refactor archive extraction code.
* Add a registry client to the registry source fetcher.
* Add tests to registry source fetcher.

Closes #491


pkg/openapi/openapi_generated.go
```diff
@@ -2455,7 +2455,7 @@ func schema_pkg_apis_core_v1alpha1_Condition(ref common.ReferenceCallback) commo
 					"lastTransitionTime": {
 						SchemaProps: spec.SchemaProps{
 							Description: "LastTransitionTime is the last time the condition transitioned from one status to another. We use VolatileTime in place of metav1.Time to exclude this from creating equality.Semantic differences (all other things held constant).",
-							Type: []string{"string"}, Format: "",
+							Type:        []string{"string"}, Format: "",
```
matthewmcnew, line 2458: What caused this?

pkg/reconciler/build/build_test.go
```diff
@@ -419,9 +419,9 @@ func testBuildReconciler(t *testing.T, when spec.G, it spec.S) {
 										ObservedGeneration: originalGeneration,
 										Conditions: corev1alpha1.Conditions{
 											{
-												Type:   corev1alpha1.ConditionSucceeded,
-												Status: corev1.ConditionUnknown,
-												Reason: "ImagePullBackOff",
+												Type:    corev1alpha1.ConditionSucceeded,
```
matthewmcnew, line 422: go fmt?

pkg/openapi/openapi_generated.go
djoyahoy: go fmt I believe

pkg/registry/fetch.go
```diff
@@ -1,90 +1,200 @@
...
+	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
 
 	"github.com/google/go-containerregistry/pkg/authn"
-	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
-	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"gopkg.in/errgo.v2/fmt/errors"
```
matthewmcnew, line 11: Is this import needed?

matthewmcnew (approved):

matthewmcnew:
Do you have a sample image to test with with one of the other source types?

djoyahoy:
> Do you have a sample image to test with with one of the other source types?

I don't have anything on hand. But, it should be easy enough with a docker file that looks like this:
```
FROM scratch
COPY ./<some-zip-file> /test.zip
LABEL source.contenttype.kpack.io=zip
```

djoyahoy:
@matthewmcnew for the above Dockerfile, the archive in the source image will need to have all of it's components extracted to the root level or provide a subPath. So if I were to use a tar for example, I would probably do something like `tar cvf my-app.tar -C my-app/ .`. Alternatively, in the image config, I'd provide a subpath of `my-app`.

matthewmcnew:
> @matthewmcnew for the above Dockerfile, the archive in the source image will need to have all of it's components extracted to the root level or provide a subPath. So if I were to use a tar for example, I would probably do something like tar cvf my-app.tar -C my-app/ .. Alternatively, in the image config, I'd provide a subpath of my-app.

I don't quite follow. Isn't the tar extracted to /workspace?

djoyahoy:
> I don't quite follow. Isn't the tar extracted to /workspace?

Yes that is correct. What I'm trying to convey is that whoever makes that tar has to be aware of how it's structured so they can provide the right `subPath` to the image config.

tylerphelan (approved):
Very nice!

PR #508, opened 2020-09-25
Add yaml tags for builder order types

tylerphelan:
This enables use of gopkg.in/yaml package for builder order config files

Ex. https://github.com/vmware-tanzu/kpack-cli/blob/361eae2b180a85aee480b5b6bff0901db42bffd0/docs/kp_import.md


tylerphelan:
we've decided to go in another direction

PR #516, opened 2020-10-06
Make image.cacheSize an immutable field

tylerphelan:
- Extend ImmutableField method for optional error details
  Allows for comparison of non-stringable types

### DIFF
--- a/pkg/apis/build/v1alpha1/image_validation.go
+++ b/pkg/apis/build/v1alpha1/image_validation.go
@@ -2,6 +2,7 @@ package v1alpha1
 
 import (
 	"context"
+	"fmt"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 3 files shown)


pkg/apis/build/v1alpha1/image_validation.go
```diff
@@ -26,46 +27,61 @@ func init() {
...
+}
+
+func (is *ImageSpec) validateCacheSize(ctx context.Context) *apis.FieldError {
+	if apis.IsInUpdate(ctx) {
+		original := apis.GetBaseline(ctx).(*Image)
+		return validate.ImmutableField(
+			*original.Spec.CacheSize,
+			*is.CacheSize,
+			"cacheSize",
+			fmt.Sprintf("got: %v, want: %v", is.CacheSize, original.Spec.CacheSize),
```
matthewmcnew, line 80: How is this different from the ImmutableField details?

matthewmcnew (approved):

pkg/apis/build/v1alpha1/image_validation.go
tylerphelan: @matthewmcnew 

`k8s.io/apimachinery/pkg/api/resource.Quantity` only has pointer receiver implementation so since we need to use the value for comparison the `%v` format would be `{{6 9} {<nil>} 6G DecimalSI}`

We also could have compared `k8s.io/apimachinery/pkg/api/resource.Quantity.Value()` or `k8s.io/apimachinery/pkg/api/resource.Quantity.AsDec()` but then the error would read ex: `got: 6000000000, want: 5000000000` which I think is less readable.

I think this is an example of an interface that could plausibly show up again for other immutable fields. Also, it's optional (although a little ugly with `strings.Join(errDetails, "\n")`)

sukhil-suresh (approved):

PR #518, opened 2020-10-07
Validate the image cacheSize has not decreased

tylerphelan:
- Fix receivers

### DIFF
--- a/pkg/apis/build/v1alpha1/image_validation.go
+++ b/pkg/apis/build/v1alpha1/image_validation.go
@@ -2,6 +2,7 @@ package v1alpha1
 
 import (
 	"context"
+	"fmt"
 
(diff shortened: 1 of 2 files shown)


sukhil-suresh (commented):
wondering what happens if user has specified a cache size, but the cluster has no DefaultStorageClass set?

matthewmcnew (approved):

matthewmcnew:
> I might have missed, but want to confirm if we handle the scenario where user has specified a cache size, but the cluster has no DefaultStorageClass set?

We don't currently do that inside the webhook. That would probably be good validation to have.

tomkennedy513 (approved):

sukhil-suresh (approved):
:flash:

tomkennedy513 (approved):

PR #522, opened 2020-10-12
Build contents diff RFC

sampeinado:
[Readable](https://github.com/sampeinado/kpack/blob/master/rfcs/0003-build-contents-diff.md)


rfcs/0003-build-contents-diff.md
```diff
@@ -0,0 +1,160 @@
...
+ 
+- As a DevOps engineer or App Developer, when I am told there is a problem with a specific build (new build failing in CI or old build failing in production), I want to know what’s different about the problem build than the last one that worked. 
+
+- As an App Developer, I want to see the difference between the commit I built locally with pack yesterday and the build that kpack built for me today based on my commit. 
+ 
+Implications: 
+- User will need to compare their most recent build to the older build in production
+i.e. builds being compared are non-adjacent 
+- Because the builds are non-adjacent, many things are likely to have changed in the contents and there will have been many “Reasons” that caused the intervening builds. Therefore, it doesn’t make a lot of sense to print the “Reason” as part of the diff
+- I assume that it is easy / possible for users to access the build number associated with a container in production (perhaps there is another build identifier we could use?)
```
mgibson1121, line 29: Given that the use case this feature seems to address is comparing a given build against one in production, seems like we should support providing an image by its tag or digest to compare against a particular build.  I know we talked about this a couple of weeks ago, but may be worth discussing further in a working group setting.

rfcs/0003-build-contents-diff.md
```diff
@@ -0,0 +1,160 @@
+### Build Contents Diff RFC 
+
+## Context
+In order to make decisions about what to do with any given build generated by TBS, users need to compare a new build’s contents with the contents of other builds.
```
matthewmcnew, line 4: This is on the kpack repo not TBS :)

rfcs/0003-build-contents-diff.md
```diff
@@ -0,0 +1,160 @@
...
+
+
+## Mockups 
+Notes on WIP mockup
+- don’t really love the all caps headers (maybe bold or other color?)
+- should there be some kind of header info to set the stage for the diff? 
+ 
+------------------
+``` 
+$kp build diff petclinic 37 13
```
matthewmcnew, line 74: As this information is largely the result of a built image, it seems like we should disconnect it from the "Build Reason" Headers. This would enable us to instead focus on displaying the unique information in the the bill of materials.

matthewmcnew: This would also enable us to `diff` against non kpack built images theoretically.

rfcs/0003-build-contents-diff.md
matthewmcnew: Given that it is probably likely that the image to be diffed against may no longer be available or easily retrievable as a build number it seems important to allow the diffing to be performed against a specific image via tag or digest.

matthewmcnew: This is especially true if we want to enable the pack usecase listed above.

rfcs/0003-build-contents-diff.md
```diff
@@ -0,0 +1,160 @@
...
+COMMIT is the same 
+CONFIG is the same
+BUILDPACK is different because it prints the Bill of Materials which represents the results of the buildpacks used instead of only the buildpack/s that caused the new build to kick off
+STACK is the same now, but will be different once we have more stack metadata such as the dpackage list (and someday a CVE list) that we can print
+ 
+Competitor examples: 
+https://github.com/GoogleContainerTools/container-diff
+
+## Complexity/Risks
+- Would this be helpful for diagnosing failed builds or builds that don't start? If so, how can we surface this info without a BoM being generated?
```
matthewmcnew, line 44: It seems that if a build fails the build-reason should sufficiently provide information to diagnose the failure.

matthewmcnew:
My big existential question on this RFC is if this functionality makes sense as part the `kp` cli. Up to this point all the functionality in `kp` has been about interacting with kpack via the kubernetes api. However, this functionality would involve reading and parsing cnb image metadata outside of kpack interaction. 

That might be more generally useful as part as another standalone tool or via pack. This diffing is probably useful to the broader cnb ecosystem as well.

mgibson1121:
I'd advocate for its inclusion in the `kp` CLI.  A common complaint about the developer experience of using kpack is the need to use other tools in conjunction with `kp`.  Given that this workflow would be very helpful to developers when debugging problems, it makes sense to provide this to them as an available tool without having to look elsewhere.

rfcs/0003-build-contents-diff.md
sampeinado: Great call out, thank you

rfcs/0003-build-contents-diff.md
sampeinado: thank you thank you

rfcs/0003-build-contents-diff.md
```diff
@@ -0,0 +1,160 @@
...
+  resources:
+    requests:
+-     cpu: "0.25"
++     cpu: "0.5"
+      memory: "128M"
+    limits:
+      cpu: "0.5"
+      memory: "256M"
+ 
+BUILDPACK
```
sclevine, line 109: Would it make sense to use something like "BILL OF MATERIALS" instead of "BULIDPACK" here? The buildpack itself isn't included in the image

rfcs/0003-build-contents-diff.md
sampeinado: ok i hear you, let me try that out

rfcs/0003-build-contents-diff.md
```diff
@@ -0,0 +1,152 @@
...
+petclinic         2020-08-04 10:12:46        index.docker.io/username/instance@sha256:afdsniuubadsiugbn3784htw70835qioresgkdjfvba3493w8uiebgd
+petclinic         2020-09-08 12:08:36        index.docker.io/username/instance@sha256:fsdgh39478th0g9tuisodfhgvns938e54iunwuiehfasf9wgiauwsfx
+
+------------------------------
+
+COMMIT CHANGES
+- Revision: 234890hpe9vjsr8gse98rghser
++ Revision: 43t789wghges87h540eq8378ge
+ 
+CONFIG CHANGES
```
matthewmcnew, line 63: As this build diff supports diffing an arbitrary build (that might not even be built with kpack) it won't necessarily have access to the "config" details of the build. I would drop this section.

rfcs/0003-build-contents-diff.md
```diff
@@ -0,0 +1,152 @@
...
+
+STACK CHANGES
+- RunImage: index.docker.io/paketobuildpacks/run@sha256:34978twgsui4789et0wh9384erhsg0e798riuhgs0er7iugshe98riugeh99rgaser892
++ RunImage: index.docker.io/paketobuildpacks/run@sha256:fh34578tg29wb645rwhs87e5hrw0e98jtfnws9e7w378qe49tr3g39n5gwe9u5hge549w
+
+Mixins:
+- libgts-0.7-4
++ libgts-0.7-5
+```
+
```
matthewmcnew, line 103: Could we add a processes section? This matches the output of `pack inspect-image` and is useful information if there are any changes.

rfcs/0003-build-contents-diff.md
```diff
@@ -0,0 +1,152 @@
...
+``` 
+$kp build diff petclinic index.docker.io/username/instance@sha256:fsdgh39478th0g9tuisodfhgvns938e54iunwuiehfasf9wgiauwsfx
+
+IMAGE NAME        BUILD TIME                 LATEST IMAGE 
+petclinic         2020-08-04 10:12:46        index.docker.io/username/instance@sha256:afdsniuubadsiugbn3784htw70835qioresgkdjfvba3493w8uiebgd
+petclinic         2020-09-08 12:08:36        index.docker.io/username/instance@sha256:fsdgh39478th0g9tuisodfhgvns938e54iunwuiehfasf9wgiauwsfx
+
+------------------------------
+
+COMMIT CHANGES
```
matthewmcnew, line 59: The [source metadata ](https://github.com/buildpacks/spec/blob/313078611d8a7925cb69a241df58e7d749d7f364/platform.md#iobuildpacksprojectmetadata-json)available on CNB built images is available on `io.buildpacks.project.metadata` (Which may include source metadata).  It might make sense to model the output from that information.

tylerphelan:
I think we should include this functionality in `kp` (working with any CNB-created image tag) regardless of if `pack` includes it because I don't think it would hurt `kp`s usability by having this feature.

I also think we should have a `-o json/yaml` flag for this for programatic uses of this feature.

jromero:
What's the status of this RFC? I can see the value of this feature and would throw my +1 on making it a `kp` feature in that, with my current understanding, it would leverage build information available via kpack (k8s) and not just solely what is available on the images. For example, config/resource changes and build times aren't things that are on the image as far as I'm aware.

tomkennedy513:
Closing due to staleness and that this seems better in the [kpack cli repo](https://github.com/buildpacks-community/kpack-cli)

PR #523, opened 2020-10-12
Image Filter RFC

mgibson1121:
[Readable](https://github.com/pivotal/kpack/blob/mgibson1121-image-filter-rfc/rfcs/0003-image-filter.md)

Moved over from the kpack CLI repo based on the decision to publish all RFCs that could affect kpack in the kpack repo.

V4.  Summary of changes:

* Used kubectl inspired `kp image list --filter key=value --filter key=value` command structure
* Updated help text and mockups to reflect this change


rfcs/0003-image-filter.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Mockups
+
+Below I will cover specific scenarios made possible by a filtering feature. All the following scenarios are derived from this proposed modification to the `kp image list` command:
+
+```
+kp image list --status <string1, string 2 ...> --latest-reason <string1, string2 ...>
+```
+
+Rather than adding an additional subcommand to the `kp image list` path like `filter` or `query`, I propose adding an additional `LATEST REASON` column to the existing output of the `kp image list` table. This would also allow for future design extension. If we wanted to add additional filtering capabilities, we would just create new attribute flags with values as arguments.
```
djoyahoy, line 52: I believe that using an individual flag for each type of filter could lead to an explosion of possible filter flags. Additionally, using a filter or query syntax is pretty common in CLIs (thats what kubectl does) and allows for more extensibility later on (via query syntax improvements).

tylerphelan: I kinda like the ui simplicity of individual flags, but if the possible filtering parameters is too high I think it becomes not worth the simplicity. We could also potentially support both...

rfcs/0003-image-filter.md
```diff
@@ -0,0 +1,176 @@
...
+------------------------------
+
+**Print Image Configs That Are Not Ready**
+
+```
+$kp image list --status not-ready -A
+
+NAME               READY     LATEST REASON    LATEST IMAGE     NAMESPACE    
+
+mg-test-image      False     CONFIG                            mg-test
```
tylerphelan, line 91: How are these sorted? Options: creation time, alphabetically by name.

rfcs/0003-image-filter.md
```diff
@@ -0,0 +1,176 @@
...
+------------------------------
+
+**Image Configs Are Not Ready Because Adding A New Buildpack To The Default Store Caused a Breaking Change**
+
+```   
+$kp image list --clusterbuilder default -A
+
+NAME             READY      LATEST REASON     LATEST IMAGE     NAMESPACE                                                                                         
+
+mg-test-image1   False      BUILDPACK                          matt-fun-zone
```
tylerphelan, line 158: We currently sort images by creation time whereas `kubectl` sorts by namespace first then alphabetically. Do we want to match that? Or maybe do namespace first then creation time

Without namespace sorting it might be harder to parse.

rfcs/0003-image-filter.md
tomkennedy513: I like the idea of just a filter flag like kubectl. I feel like it's pretty clean and easy to use from a user standpoint.

rfcs/0003-image-filter.md
mgibson1121: @tylerphelan @sukhil-suresh @tomkennedy513 I agree with the suggestion, but I'm wondering if it makes more sense for that detail to be addressed in one of the stories that implements the RFC.  I'm concerned about being too prescriptive at the RFC level, imo stories should be broken off of it not serve as the story for implementation.

rfcs/0003-image-filter.md
mgibson1121: One thing that confuses me about this suggestion is that it seems like we'd be trading the complexity of flags for the complexity of arguments.  The user still has to specify the values they want to filter.  So if we did something like `kp image filter/query` we'd have to support a list of arguments to be used.  Is that less technically complex than supporting a bunch of flags?  I also slightly prefer the UX of the current implementation because it can be reconciled with `kp image list`.  The user is effectively trying to filter this list, not generate their own list based on inputs (which is effectively what `kp image filter` would be.  @tomkennedy513 @tylerphelan @djoyahoy lmk your thoughts.

mgibson1121:
@sukhil-suresh For some reason there isn't an option visible to reply to your comment (https://github.com/pivotal/kpack/pull/523#issuecomment-710300827).  So posting here.  I appreciate you raising this scenario.  It caused me to reflect a lot on a case I hadn't considered.  However, I propose we move forward with the RFC based on my reasoning below.  

The Operator's primary concerns are that apps are using the correct dependencies and are in a ready state.  The former concern can be addressed by manipulation of the store, stack, and builder/clusterbuilder resources.  As you mention, this can be verified by querying images that use a certain builder.  Given this feature, let us assume that the Operator is satisfied with his concern that images are using the correct dependencies during the build process.  Given this assumption, we can then say that the Operator's main concern during a stack/store update is to confirm the ready status of a group of images.  His concern is heightened because such an operation will trigger more rebuilds in a small period of time than otherwise would occur.  

In this case, being able to return a list of images that are both not ready and use a particular builder would be extremely helpful compared to the operations we currently provide.  Because it would surface images that are in a not ready state (for whatever reason) in the aftermath of an event that triggered many images to rebuild, thereby giving the Operator an indication of where to dig deeper and resolve problems.  Further, I feel the `latest-reason` field also provides value by allowing the operator to quickly understand if an event that triggered many rebuilds went off successfully (even if that operation only reflects a state of the world precisely at the time it was run).

sukhil-suresh:
> In this case, being able to return a list of images that are both not ready and use a particular builder would be extremely helpful compared to the operations we currently provide.

>  Further, I feel the latest-reason field also provides value by allowing the operator to quickly understand... 

Yes, I definitely agree with the above statements

sukhil-suresh (approved):

rfcs/0003-image-filter.md
tomkennedy513: I don't know if it's less technically complex to maintain, but I think having a filter query flag vs individual flags would accomplish the same thing. I am a little confused by the comment that users are not trying to generate their own list mainly because my understanding of the filter query method would still return all columns (i.e not creating some new data view) so if a user ran something like `kp image list --lastest-reason "config"` vs `kp image list --filter latest-reason=config` they would get the same result. I prefer the filter flag because it is how `kubectl` does it and since users are likely using both `kp` and `kubectl` together, I figure it would be slightly less cognitive load if the  filtering mechanism was identical

sampeinado: I agree with @djoyahoy and @tomkennedy513 that we should follow the kubectl convention and use a filter flag. We should keep in mind Jakob's Law of the Internet User Experience: "Users spend most of their time on other sites." This means that users prefer your tool to work the same way as all the other tools they already know.

sampeinado: As far as I can tell, the flag for filtering a `kubectl get` command is `--selector` (`-l`) or `--field-selector` but I can't quite tell what the difference between them is

sampeinado: @mgibson1121 i dont think anyone is proposing a new subcommand of `kp image filter` or something like that. the proposal would be `kp image list --selector reason==config". Am I right @djoyahoy @tomkennedy513 ?

mgibson1121: This strikes me as a different case than Kubectl because it just runs matching on keys and values for labels (these labels are frequently managed by end-users of kubectl).  So this functionality strikes me as a generic implementation so that they can make something generically useful for lots of users.  In this RFC we're trying to find an interface to allow for filtering of an opinionated set of criteria.   Based on this feedback and my above response, I propose the following syntax as a path forward: `kp image list --latest-reason=value1, value2`.  We would provide help text for values that are supported.

tomkennedy513: after talking about this with Matt, I also feel like this is the best path forward

tomkennedy513 (approved):

rfcs/0003-image-filter.md
matthewmcnew: @tomkennedy513 Could you add a little info on your thought process/conversation there?

matthewmcnew (approved):

rfcs/0003-image-filter.md
mgibson1121: We're going with this structure.  I will update the mockups to reflect this decision. 

`kp image list --filter reason=stack, config --field-selector clusterbuilder=value`

djoyahoy (approved):

PR #527, opened 2020-10-19
surface errors from image wait

tomkennedy513:
- only display error message if it exists to prevent dangling colon in message

### DIFF
--- a/pkg/logs/wait_for_image.go
+++ b/pkg/logs/wait_for_image.go
@@ -56,7 +56,11 @@ func (w *imageWaiter) resultOfImageWait(ctx context.Context, writer io.Writer, g
 		return w.waitBuild(ctx, writer, image.Namespace, image.Status.LatestBuildRef)
 	}
 
-	if image.Status.GetCondition(corev1alpha1.ConditionReady).IsFalse() {
+	if condition := image.Status.GetCondition(corev1alpha1.ConditionReady); condition.IsFalse() {
+		if condition.Message != "" {
+			return "", errors.Wrap(errors.New(condition.Message), fmt.Sprintf("update to image %s failed", image.Name))
+		}
+
 		return "", errors.Errorf("update to image %s failed", image.Name)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/logs/wait_for_image.go
```diff
@@ -138,8 +142,12 @@ func (w *imageWaiter) waitBuild(ctx context.Context, writer io.Writer, namespace
 		return "", errors.New("unexpected object received")
 	}
 
-	if build.Status.GetCondition(corev1alpha1.ConditionSucceeded).IsFalse() {
-		return "", errors.Errorf("update to image failed")
+	if condition:= build.Status.GetCondition(corev1alpha1.ConditionSucceeded); condition.IsFalse() {
+		if condition.Message != "" {
```
sukhil-suresh, line 146: wonder if a single `return "", errors.Wrap(errors.New(condition.Message), "update to image failed")` without the `condition.Message != ""`  check is sufficient? not sure.

tomkennedy513: it will try to wrap still so you get "update to image failed: " with the dangling colon

sukhil-suresh: got it. thanks for clarifying

pkg/logs/wait_for_image.go
```diff
@@ -56,7 +56,11 @@ func (w *imageWaiter) resultOfImageWait(ctx context.Context, writer io.Writer, g
 		return w.waitBuild(ctx, writer, image.Namespace, image.Status.LatestBuildRef)
 	}
 
-	if image.Status.GetCondition(corev1alpha1.ConditionReady).IsFalse() {
+	if condition := image.Status.GetCondition(corev1alpha1.ConditionReady); condition.IsFalse() {
+		if condition.Message != "" {
+			return "", errors.Wrap(errors.New(condition.Message), fmt.Sprintf("update to image %s failed", image.Name))
```
matthewmcnew, line 61: Could we use a single `errors.Errorf` here?

matthewmcnew (approved):

sukhil-suresh (approved):

pkg/logs/wait_for_image.go
tomkennedy513: ya I can change that

djoyahoy (approved):

PR #534, opened 2020-10-23
Windows images rfc

tylerphelan:
[Readable](https://github.com/pivotal/kpack/blob/windows-containers/rfcs/0000-windows-containers.md)


rfcs/0000-windows-containers.md
```diff
@@ -0,0 +1,76 @@
+## Problem:
+
+kpack does not support building containers that can run on windows
+
+## Background
+
+kpack has an [rfc](https://github.com/pivotal/kpack/pull/534) to add windows support
```
tomkennedy513, line 7: isn't this just linking to this pull request

tomkennedy513 (approved):

rfcs/0000-windows-containers.md
tylerphelan: 😂

rfcs/0000-windows-containers.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Prior Art:
+
+[pack cli](https://github.com/buildpacks/pack) has support for windows containers
+
+## Alternatives:
+
+
+
+## Risks:
```
micahyoung, line 69: Maybe worth including a risk about how Windows nodes can only create/run containers from images that match their OS version (ex: Windows Server LTSC 2019 can only run images that are `ltsc2019` or equivalent version. [docs]https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/version-compatibility?tabs=windows-server-2019%2Cwindows-10-20H2) and handling failures from mismatched OS versions may need additional work or documentation. This also could end up being a non-issue if just letting things fail and propagating k8s errors back to users provides sufficient feedback.

Additionally, I feel it's definitely worth keeping out-of-scope any support for more than a single Windows OS version per cluster.

More details about supported Windows OS versions in k8s [docs](https://kubernetes.io/docs/setup/production-environment/windows/intro-windows-in-kubernetes/#windows-os-version-support).

matthewmcnew, line 69: Also, might we worth calling out the need for a windows build-init/completion image for each supported windows os version.

rfcs/0000-windows-containers.md
```diff
@@ -0,0 +1,73 @@
...
+
+### kpack should support running builds on windows workers
+
+**Actions to Take**
+
+- kpack-controller will accept new environment variables for windows images needed to run builds. These will have the same functionality as on linux and will be used for running builds on windows workers.
+    - `WINDOWS_BUILD_INIT_IMAGE` - a windows build init image
+    - `WINDOWS_COMPLETION_IMAGE` - a windows completion image
+    - `WINDOWS_REBASE_IMAGE` - a windows rebase image
+    - These will be optional to remove the dependency on windows images
```
matthewmcnew, line 46: Is there a reason for these to be optional or could we always ship with them in our release.yaml? They will never get used/pulled if windows is not used.

rfcs/0000-windows-containers.md
```diff
@@ -0,0 +1,73 @@
...
+**Actions to Take**
+
+- kpack-controller will use a new environment variable `WINDOWS_LIFECYCLE_IMAGE` to reference the windows lifecycle image
+- When a Builder or ClusterBuilder is created using a windows stack and windows buildpackages, kpack-controller must create a windows builder using the windows lifecycle image.
+    - Image os is determined by the image [os property](https://github.com/opencontainers/image-spec/blob/master/config.md#:~:text=os%20string).
+
+Notes:
+
+- Builders with mixed os images must error gracefully
+- Builders should be labeled with `os.kpack.io: windows` for os disambiguation independent of image metadata
```
matthewmcnew, line 36: Why do we need to label kpack produced builder images with our own os label?

tylerphelan: I think this removes the need to talk to the registry to determine the os

rfcs/0000-windows-containers.md
tylerphelan: I don't see the need for them.

And I would be concerned with distribution rights maybe?

matthewmcnew: @micahyoung Can correct me if I am wrong but, we wouldn't actually ship the windows layers (as they will be foreign layers) so, I don't there are distribution concerns.

rfcs/0000-windows-containers.md
matthewmcnew: At what point in the kpack lifecycle?

rfcs/0000-windows-containers.md
```diff
@@ -0,0 +1,119 @@
...
+
+### kpack should support running builds on windows workers
+
+**Actions to Take**
+
+- kpack-controller will accept new environment variables for windows images needed to run builds. These will have the same functionality as on linux and will be used for running builds on windows workers.
+    - `WINDOWS_BUILD_INIT_IMAGE` - a windows build init image
```
matthewmcnew, line 43: This is creating a combinatorial explosion. Perhaps, we should just have one windows image that can build-init/complete/rebase?

rfcs/0000-windows-containers.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Prior Art:
+
+[pack cli](https://github.com/buildpacks/pack) has support for windows containers
+
+## Risks:
+
+**Windows images can only run and be created on workers with the same OS version**
+
+Windows images can only run on windows nodes with the same OS. ex: Windows Server LTSC 2019 can only run images that are ltsc2019 or equivalent version [docs](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/version-compatibility?tabs=windows-server-2019%2Cwindows-10-20H2#:~:text=Supports%20process%20isolation)
```
matthewmcnew, line 69: Is there a specific windows version we are targeting and will we need windows build-init/rebase/completion images for each targeted windows version?

matthewmcnew:
I think this RFC should also discuss how stack/store resources handle windows.  As it is unclear to me if a store should support both windows/linux buildpackages. For the sake of clarify, I am inclined to say no.

rfcs/0000-windows-containers.md
micahyoung: @matthewmcnew Well, all non-distributable baselayer blobs would live on `mcr.microsoft.com` and be referenced as a `url` in the image manifest layer, but any layers you create yourself above that would need to live as blobs in your registry alongside your the manifest. 

Here's the bottom two layers from `golang:1.15-nanoserver-1809` manifest for example: `layers[0]`'s blob would come from `mcr.microsoft.com` but `layers[1]`'s blob would come from the manifest'ss registry (`index.docker.io` in this case):
```
  "layers": [
    {
      "mediaType": "application/vnd.docker.image.rootfs.foreign.diff.tar.gzip",
      "size": 101205155,
      "digest": "sha256:aab6118ce69c93410df7fa15842a6e3b3c7ff20b639c779b5d5f78e7653eaa07",
      "urls": [
        "https://mcr.microsoft.com/v2/windows/nanoserver/blobs/sha256:aab6118ce69c93410df7fa15842a6e3b3c7ff20b639c779b5d5f78e7653eaa07"
      ]
    },
    {
      "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
      "size": 911,
      "digest": "sha256:b5bc5109798f9a64116d8a878ce7b6065052706b8c0b20c2a60b08b6e159c98e"
    },
    ...
  ]
```

rfcs/0000-windows-containers.md
tylerphelan: creating builds time

rfcs/0000-windows-containers.md
tylerphelan: Given the following:
- `Windows Server 2019 (LTSC) aka version 1809 is the only Windows operating system supported currently on all k8s api versions`
- `Once they support Windows containers with Hyper-V isolation in Kubernetes, the limitation and compatibility rules will change.`

I think we should only support one windows version (LTSC 2019) and kpack should only be able to support one version at a time

rfcs/0000-windows-containers.md
tylerphelan: I think that's viable

I would prefer the windows behavior mirror the linux for consistency and maintainability

rfcs/0000-windows-containers.md
matthewmcnew: I think think that would prevent distribution concerns of publishing windows images as part of the kpack release.

rfcs/0000-windows-containers.md
matthewmcnew: Is this label a k8s label or a registry label? We already fetch the registry on each build pod creation.

rfcs/0000-windows-containers.md
matthewmcnew: It might make sense to do this for linux as well.

rfcs/0000-windows-containers.md
tylerphelan: k8s

in all of the cases I've added a `os.kpack.io: windows`, if we already read the image then it's not necessary

matthewmcnew: We could add an `os` key to the builder/image/stack/(store?) status but, It does not seem necessary and probably could be skipped in the rfc.

rfcs/0000-windows-containers.md
micahyoung: I'd recommend starting with support for Windows Hosts and Images at version LTSC2019. It's currently the only version you can use as a [k8s node](https://kubernetes.io/docs/setup/production-environment/windows/intro-windows-in-kubernetes/#windows-os-version-support) that *also* has long-term support from Microsoft ([supported until 2024](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/base-image-lifecycle#:~:text=17763,01/09/2024), whereas other Semi-Annual Channel (SAC) host/image versions only get 18 months of support).

Handling non-matching host/image OS versions gets fuzzier after that... LTSC2019 nodes cannot *currently* create k8s containers from newer images [but may eventually](https://kubernetes.io/docs/setup/production-environment/windows/intro-windows-in-kubernetes/#:~:text=Once%20we%20support%20Windows%20containers%20with%20Hyper). Microsoft does have strong binary ABI compatibility which should ensure binaries created in an LTSC2019 container would run on a 2004 host/image.

With that said, it's probably worth validating that anyone would be running non-LTSC2019 versions in production clusters, I suspect there may not be a strong use case for supporting the SAC versions.

rfcs/0000-windows-containers.md
micahyoung: If you all are considering ARM support too, that may add yet another dimension. Docker hub images rely heavily on [image indexes](https://github.com/opencontainers/image-spec/blob/master/image-index.md) (aka manifest lists) currently, for which k8s appears to have [patterns and support](https://kubernetes.io/docs/concepts/containers/images/#multi-architecture-images-with-manifests). Also, ggcr lets you filter indexes with [`remote.WithPlatform()`](https://github.com/google/go-containerregistry/blob/a84993334b277818146f48584a1ad748e1b53859/pkg/v1/remote/options.go#L114).

rfcs/0000-windows-containers.md
```diff
@@ -0,0 +1,136 @@
...
+
+Ergo:
+
+- Each windows image installed with `kpack` (`build-init`, `rebase`, `completion`, `lifecycle`) must be for the same os version and must match the cluster's windows nodes
+- Stacks and buildpackages must be for the same os version and must match the cluster's windows nodes
```
micahyoung, line 88: All pack-generated Windows buildpackage images have a blank `os.version`, as currently implemented. Windows daemons give special treatment to a blank `os.version` - all known Windows daemon's will skip validating against their own version and are able to pull them. So `kpack` won't need to validate buildpackage's `os.version`. The only buildpackage-to-stack compatibility restrictions would be expressed as mixins, just like Linux, and should *just work*.

Windows Stack images (build/run) **do** always have `os.version` field (ex: `10.0.17763.1518` format: [`<major>.<minor>.<build>.<revision>`](https://docs.microsoft.com/en-us/dotnet/api/system.version.build?view=netcore-3.1)), set by Microsoft on the base image, that must match the os-version of the daemon where they are run.

tylerphelan: nice!

rfcs/0000-windows-containers.md
```diff
@@ -57,9 +57,10 @@ Notes:
     - `WINDOWS_BUILD_INIT_IMAGE` - a windows build init image
     - `WINDOWS_COMPLETION_IMAGE` - a windows completion image
     - `WINDOWS_REBASE_IMAGE` - a windows rebase image
+    - `WINDOWS_SETUP_CA_CERTS_IMAGE` - a windows setup ca certs image
```
matthewmcnew, line 60: This is not part of kpack.

rfcs/0000-windows-containers.md
```diff
@@ -0,0 +1,150 @@
...
+```
+nodeSelector:
+    kubernetes.io/os: linux
+```
+
+### kpack should restrict cluster stores and stacks to be os-specific
+
+This would improve simplicity and usability. If you create mixed-os resources, it may be difficult to get out of the situation.
+
+If a user creates a cluster store with buildpackages of mixed os, the resource should be marked as not-ready
```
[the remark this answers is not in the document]
sukhil-suresh: also noticed this [comment](https://github.com/pivotal/kpack/pull/534#issuecomment-717321160) which leans towards classifying a clusterstore as one or the other vs both

rfcs/0000-windows-containers.md
tylerphelan: the resource label is now not a part of the rfc

rfcs/0000-windows-containers.md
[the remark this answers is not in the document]
tylerphelan: The rfc was updated to now not restrict stores to a single os

rfcs/0000-windows-containers.md
tylerphelan: I think ARCH concerns is out of scope for the rfc

rfcs/0000-windows-containers.md
```diff
@@ -0,0 +1,132 @@
...
+
+**Actions to Take**
+
+- kpack-controller will accept new configuration via environment variable(s) for windows image(s) needed to run builds. The current behavior that must be recreated on windows is contained in the following images:
+    - `BUILD_INIT_IMAGE` - getting the build ready with creds and source
+    - `COMPLETION_IMAGE` - an image to run as a required regular (non-init) container
+    - `REBASE_IMAGE` - run a rebase
+    - We'll call these "helper images"
+    - Out of scope:
+        - What to name the windows versions of these images
```
sukhil-suresh, line 62: > What to name the windows versions of these images

Not about the naming... but more about do we actually need windows specific versions for ALL of these images? Unless intentionally choosing to go down that path.
What particularly stood out for me was the `COMPLETION_IMAGE ` (and maybe the `BUILD_INIT_IMAGE` to a certain degree - not so sure how it works).

The `COMPLETION_IMAGE`, if I recollect correctly, all it did was show a completion message. But more importantly, there was a recent discussion about using this image for Image signing. And I am not clear if we need a linux and windows variant for that.

tylerphelan: the completion image also serves the requirement of having at least one regular (non-init) container which will need to run on the windows node

rfcs/0000-windows-containers.md
```diff
@@ -0,0 +1,132 @@
...
+- Builders will not be aware of the buildpackages os. The only compatibility constraint between stacks and buildpackages for creating builders will be the stack id.
+
+### kpack should support running builds on windows workers
+
+**Actions to Take**
+
+- kpack-controller will accept new configuration via environment variable(s) for windows image(s) needed to run builds. The current behavior that must be recreated on windows is contained in the following images:
+    - `BUILD_INIT_IMAGE` - getting the build ready with creds and source
+    - `COMPLETION_IMAGE` - an image to run as a required regular (non-init) container
+    - `REBASE_IMAGE` - run a rebase
```
matthewmcnew, line 59: As rebase does not need to execute anything specific to the windows operating system. We probably don't need a "windows" specific rebase image.

sukhil-suresh (approved):

djoyahoy:
@tylerphelan do windows stacks have the same ABI compatibility guarantees fro rebasing?

tylerphelan:
> do windows stacks have the same ABI compatibility guarantees fro rebasing?

yup

djoyahoy (approved):

PR #541, opened 2020-11-04
Notary V1 Integration

djoyahoy:
This PR integrates with Notary for image singing. The following explains the singing process:
1. `notary init example.registry.io/my-app` generates/encrypts the root, targets, and snapshot private keys. The targets key will be used for singing new image tags.
1. `notary key rotate example.registry.io/my-app snapshot -r` rotates the snapshot key to the server so that it is not managed locally. This is the default behavior of the docker CLI when signing images.
1. `notary publish example.registry.io/my-app` publishes local changes to the Notary server.
1. `kubectl create secret generic notary-secret --from-literal=password=<password> --from-file=$HOME/.notary/private/<targets-hash>.key` create a k8s secret with the targets private key and the password used to decrypt the key.
1. Create an image config that references the notary secret and the notary server URL.
1. After export, the completion container uploads the image tag, manifest digest, manifest size to the Notary server. It verifies the content using the private targets key. Registry secrets are used to authenticate with the Notary server as Notary traditionally uses the same auth method as the registry (see [here](https://docs.docker.com/notary/service_architecture/#example-client-server-signer-interaction) and [here](https://github.com/docker/distribution/blob/master/docs/spec/auth/token.md)).
1. docker CLI users with docker content trust enabled will connect to the Notary server when pulling images to validate their authenticity. This simply validates that the manifest digest is correct (as it is a cryptographic hash function).


pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -84,6 +86,16 @@ var (
...
 	}
+	notaryV1Volume = corev1.VolumeMount{
+		Name:      notaryDirName,
+		MountPath: "/var/notary/v1",
+		ReadOnly:  true,
+	}
+	reportVolume = corev1.VolumeMount{
+		Name:      reportDirName,
+		MountPath: "/var/report",
+		ReadOnly:  false,
```
tomkennedy513, line 97: can this be read only if it just contains report.toml

djoyahoy: it has to be writable because the export step writes to it.

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -84,6 +88,21 @@ var (
 		MountPath: "/builderPullSecrets",
 		ReadOnly:  true,
 	}
+	notaryV1Volume = corev1.VolumeMount{
+		Name:      notaryDirName,
+		MountPath: "/var/notary/v1",
+		ReadOnly:  true,
+	}
+	reportVolume = corev1.VolumeMount{
```
matthewmcnew, line 96: Could report.toml just be written to the /layers mount?

pkg/apis/build/v1alpha1/notary_types.go
```diff
@@ -0,0 +1,24 @@
...
+// +k8s:openapi-gen=true
+type NotaryConfig struct {
+	V1 *NotaryV1Config `json:"v1,omitempty"`
+}
+
+// +k8s:openapi-gen=true
+type NotaryV1Config struct {
+	URL             string           `json:"url"`
+	SecretRef       NotarySecretRef  `json:"secretRef"`
+	ConfigMapKeyRef *ConfigMapKeyRef `json:"configMapKeyRef,omitempty"`
```
matthewmcnew, line 12: Is `configMapKeyRef` common for config map references in the k8s api?

pkg/apis/build/v1alpha1/build_types.go
```diff
@@ -66,6 +66,7 @@ type BuildSpec struct {
 	Env       []corev1.EnvVar             `json:"env,omitempty"`
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
 	LastBuild *LastBuild                  `json:"lastBuild,omitempty"`
+	Notary    NotaryConfig                `json:"notary,omitempty"`
```
matthewmcnew, line 69: Should this be nil-able?

pkg/notary/image_signer.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	return gun, targets, nil
+}
+
+func (s *ImageSigner) makeRemoteStore(serverURL string, gun data.GUN, caCertFilePath string, keychain authn.Keychain) (storage.RemoteStore, error) {
+	transport := http.DefaultTransport
+
+	if caCertFilePath != "" {
+		s.Logger.Printf("Using CA certifiacte '%s'\n", caCertFilePath)
```
matthewmcnew, line 133:
```suggestion
		s.Logger.Printf("Using CA certificate '%s'\n", caCertFilePath)
```

pkg/apis/build/v1alpha1/build_pod.go
djoyahoy: No, not exactly sure what export does to the layers volume, but post export, that volume attempts to execute the launcher when mounted.

pkg/apis/build/v1alpha1/notary_types.go
djoyahoy: I borrowed it from the pod spec.

pkg/apis/build/v1alpha1/build_types.go
djoyahoy: It could be, but we'd have to do a nil check on both the notary config and the V1 portion of the notary config.

pkg/apis/build/v1alpha1/build_pod.go
matthewmcnew: Ahh as in mounting `/layers` overwrites the completion binary?

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -347,14 +363,100 @@ func (b *Build) BuildPod(config BuildPodImages, secrets []corev1.Secret, bc Buil
...
+			reportVolume,
+			notaryCertsVolume,
+		),
+		ImagePullPolicy: corev1.PullIfNotPresent,
+	}
+}
+
+func (b *Build) notarySecretVolume() corev1.Volume {
+	config := b.NotaryV1Config()
+	if config != nil {
```
matthewmcnew, line 418: Could we invert this?

pkg/apis/build/v1alpha1/build_types.go
matthewmcnew: Fair enough. Could that be extracted away from the caller in the `NotaryV1Config()` method?

pkg/notary/authenticating_round_tripper.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/pkg/errors"
+)
+
+type AuthenticatingRoundTripper struct {
+	Token               string
+	Keychain            authn.Keychain
+	WrappedRoundTripper http.RoundTripper
+}
+
+func (a *AuthenticatingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
```
matthewmcnew, line 20: I'm a little confused by this. It is using the docker credentials to authenticate to the notary server?

pkg/notary/image_signer.go
```diff
@@ -0,0 +1,203 @@
...
+	cryptoStore := storage.NewMemoryStore(nil)
+
+	fileInfos, err := ioutil.ReadDir(notarySecretDir)
+	if err != nil {
+		return nil, err
+	}
+
+	privateKeyFound := false
+	for _, info := range fileInfos {
+		if strings.HasSuffix(info.Name(), ".key") {
```
matthewmcnew, line 173: Just look for any key in the configured secret?

pkg/apis/build/v1alpha1/notary_types.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+// +k8s:openapi-gen=true
+type NotarySecretRef struct {
+	Name string `json:"name"`
+}
+
+// +k8s:openapi-gen=true
+type ConfigMapKeyRef struct {
```
matthewmcnew, line 21: Sorry, if you already outlined this but, could you explain how this cert is used again?

matthewmcnew:
This is really good PR and really does a good job of implementing the notary integration without unnecessary complexity. The exact mechanics/terminology of how notary signs images is still a bit murky for me. Could you walk through that again at somepoint?

pkg/notary/authenticating_round_tripper.go
djoyahoy: That is correct. The notary server uses the same auth that docker uses.

pkg/notary/image_signer.go
djoyahoy: Right now, it is assumed that there is only one key.

pkg/apis/build/v1alpha1/notary_types.go
djoyahoy: This is the certificate for the notary server. For harbor, that is usually the same cert as the registry, but that may not always be the case. If we drop the ability to provide the cert here, then users have no way of providing a cert without modifying the completion image that we provide (unless they had some kind of cert injection mechanism).

djoyahoy:
> This is really good PR and really does a good job of implementing the notary integration without unnecessary complexity. The exact mechanics/terminology of how notary signs images is still a bit murky for me. Could you walk through that again at somepoint?

Sounds good. I'll create a summary in the PR description.

pkg/apis/build/v1alpha1/build_pod.go
djoyahoy: Ah yup, thats exactly right.

pkg/apis/build/v1alpha1/build_types.go
djoyahoy: Yea it could be. I think thats preferable.

pkg/apis/build/v1alpha1/notary_types.go
matthewmcnew: How is that different from the current ability for users to provide a cert for their registry?

pkg/notary/image_signer.go
djoyahoy: In the future, we could just parse each key file in the volume, as the keys are well formed and contain the GUN.

pkg/apis/build/v1alpha1/notary_types.go
djoyahoy: Removed as it was inconsistent with how registry certs are provided.

pkg/notary/authenticating_round_tripper.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil, errors.Errorf("received status code '%d'", resp.StatusCode)
+	}
+
+	var body map[string]interface{}
+	err = json.NewDecoder(resp.Body).Decode(&body)
+	if err != nil {
+		return nil, err
+	}
+
+	a.Token = body["token"].(string)
```
tylerphelan, line 95: maybe check type assertion here

tylerphelan (commented):
Should we have image and/or build validation for notary config? (must all fields not empty, maybe check that notary secret exists, maybe check url is valid url)

Also maybe run a `goimports`

pkg/apis/build/v1alpha1/build.go
```diff
@@ -119,6 +119,16 @@ func (b *Build) Finished() bool {
 	return !b.Status.GetCondition(corev1alpha1.ConditionSucceeded).IsUnknown()
 }
 
+func (b *Build) NotaryV1Config() *NotaryV1Config {
+	if b == nil {
```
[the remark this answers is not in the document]
djoyahoy: I don't think ORs in go have short circuit eval so if `b` were nil, it would still try and eval the  `b.Spec.Notary == nil` which may panic.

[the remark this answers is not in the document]
tylerphelan: @djoyahoy I thought it did:

```
package main

  import "fmt"

  func main() {
    type foo struct {
      bar string
    }

    var myFoo *foo = nil
    if myFoo == nil || myFoo.bar == "" {
      fmt.Println("hi")
    }
  }
```
```
$ go run main.go
> hi
```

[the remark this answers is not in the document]
djoyahoy: @tylerphelan you're right! 😲

tomkennedy513 (approved):

matthewmcnew (approved):

PR #546, opened 2020-11-12
Windows Support

matthewmcnew:


pkg/filehelpers/file_helper.go
```diff
@@ -0,0 +1,66 @@
+package filehelpers
+
+import (
+	"io"
+	"io/ioutil"
+	"os"
+	"path"
+)
+
+func CopyDir(src string, dst string) error {
```
matthewmcnew, line 10: Where is this used?

matthewmcnew: Just git fetch?

pkg/registry/fetch.go
```diff
@@ -60,10 +67,15 @@ func (f *Fetcher) Fetch(dir, registryImage string) error {
 		handler = handleSource
 	}
 
-	if err := handler(img, dir); err != nil {
+	if err := handler(img, tmpDir); err != nil {
 		return err
 	}
 
+	err = filehelpers.CopyDir(tmpDir, dir)
```
matthewmcnew, line 74: Did we observe that this was necessary?

tomkennedy513: No, we don't need it. I meant to remove that, good catch

pkg/filehelpers/file_helper.go
tomkennedy513: yes, the copy file is used in build init as well, but copy dir is only used in git

tomkennedy513: I ended up getting rid of the file helper since it wasn't really necessary, I just put the function in the two places it was needed

config/values.yaml
```diff
@@ -3,7 +3,9 @@
 controller_image: gcr.io/controller
 webhook_image: gcr.io/webhook
 build_init_image: gcr.io/build-init
+build_init_windows_image: gcr.io/cf-build-service-public/kpack/build-init-windows
```
matthewmcnew, line 6: Is the expectation that this is built/updated out of band of ./hack/release.sh?

tomkennedy513: yes, at least for now. Maybe in the future we could conceivably build the images using something other than pack that could run on linux and still build the windows images

go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	github.com/mgutz/ansi v0.0.0-20170206155736-9520e82c474b
 	github.com/pkg/errors v0.9.1
 	github.com/sclevine/spec v1.4.0
+	github.com/shurcooL/go v0.0.0-20180423040247-9e1955d9fb6e
```
matthewmcnew, line 23: What is this used for?

pkg/git/fetch.go
```diff
@@ -1,17 +1,18 @@
...
-
+	"fmt"
 	"github.com/BurntSushi/toml"
-
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/transport"
 	"github.com/pkg/errors"
+	"io"
```
matthewmcnew, line 11: Can we rearrange these imports?

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -34,13 +34,35 @@ const (
 	notaryDirName = "notary-dir"
 	reportDirName = "report-dir"
 
+	networkWaitLauncherDir = "kpack-network-wait-launcher-dir"
```
tylerphelan, line 37: `networkWaitLauncherDirName` to match?

pkg/buildpod/generator.go
```diff
@@ -145,12 +145,18 @@ func (g *Generator) fetchBuilderConfig(build BuildPodable) (v1alpha1.BuildPodBui
 		return v1alpha1.BuildPodBuilderConfig{}, err
 	}
 
+	config, err := image.ConfigFile()
```
tylerphelan, line 148: maybe don't need this with the os in the builder status now

pkg/git/fetch.go
```diff
@@ -91,6 +103,63 @@ func (f Fetcher) Fetch(dir, gitURL, gitRevision, metadataDir string) error {
...
+		return err
+	}
+
+	srcInfo, err := os.Stat(src)
+	if err != nil {
+		return err
+	}
+
+	return os.Chmod(dest, srcInfo.Mode())
+}
```
sukhil-suresh, line 129: noticed duplicate in [cmd/build-init/main.go](https://github.com/pivotal/kpack/blob/de396d3f95f63165aa598d6cd8c23fe31e3e58e6/cmd/build-init/main.go#L223-L246)
wonder if it makes sense to share it in /pkg/* ?

go.mod
tomkennedy513: I think that showed up when I merged master in to this branch, I'm not using it for anything. It looks like it's being used in rebase/main.go

pkg/buildpod/generator.go
tomkennedy513: we don't have the builder resource at this point, just the builder image so I don't think I can get that status unless I fetch the resource

go.mod
sukhil-suresh: @tomkennedy513 I ran into [this oddity](https://github.com/pivotal/kpack/pull/590#discussion_r552902337) too and @matthewmcnew fixed it with https://github.com/pivotal/kpack/pull/593. not approved/merged yet

cmd/build-init/main.go
```diff
@@ -126,6 +133,24 @@ func main() {
...
+	if runtime.GOOS != "windows" {
+		return nil
+	}
+
+	executablePath, err := os.Executable()
+	if err != nil {
+		return err
+	}
+
+	err = copyFile(filepath.Join(filepath.Dir(executablePath), "network-wait-launcher.exe"), filepath.Join("/kpack", "network-wait-launcher.exe"))
```
tylerphelan, line 146: maybe extract the filename to a const ("network-wait-launcher.exe")

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -3,7 +3,7 @@ package v1alpha1
 import (
 	"encoding/json"
 	"fmt"
-
```
matthewmcnew, line 6: Can we fix this import order?

pkg/buildpod/generator_test.go
```diff
@@ -2,6 +2,7 @@ package buildpod_test
 
 import (
 	"fmt"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
```
matthewmcnew, line 5: Imports

pkg/git/fetch.go
```diff
@@ -1,17 +1,18 @@
 package git
 
 import (
-	"log"
-	"os"
-	"path"
-
+	"fmt"
```
matthewmcnew, line 4: imports :)

api/openapi-spec/swagger.json
```diff
@@ -4266,8 +4266,7 @@
           "$ref": "#/definitions/io.k8s.api.core.v1.LocalObjectReference"
         },
         "name": {
-          "type": "string",
```
matthewmcnew, line 4269: What happened with all these changes?

matthewmcnew:
What are are remaining issues with Windows?

- Cache Support

pkg/git/fetch.go
```diff
@@ -25,7 +27,12 @@ func (f Fetcher) Fetch(dir, gitURL, gitRevision, metadataDir string) error {
 		return err
 	}
 
-	repo, err := git.PlainInit(dir, false)
+	tmpDir, err := ioutil.TempDir("", "git-clone-")
+	if err != nil {
+		return err
+	}
+
+	repo, err := git.PlainInit(tmpDir, false)
```
aemengo, line 35: Maybe it's worth a comment about we clone to temp directory here. This is something that might become tribal knowledge.

pkg/apis/build/v1alpha1/build_pod.go
tomkennedy513: ya is need ti fix goland, it keeps messing them up

cmd/build-init/main.go
```diff
@@ -126,6 +133,24 @@ func main() {
...
+	if err != nil {
+		return err
+	}
+
+	return waitForDns(*hostName)
```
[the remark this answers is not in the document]
tomkennedy513: I think it works because we are passing in as an env variable and flag.Parse() is for parsing the arguments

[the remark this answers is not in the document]
tomkennedy513: but maybe we should move it anyway. Also, I can extract to a separate variable

cmd/network-wait-launcher/main.go
```diff
@@ -0,0 +1,49 @@
...
+	// example: /kpack/network-wait-launcher gcr.io -- completion --notary-url="some-url"
+
+	hostname := os.Args[1]
+	command := os.Args[3:]
+
+	if err := waitForDns(hostname); err != nil {
+		log.Fatal(err)
+	}
+
+	cmd := exec.Command(command[0], command[1:]...)
```
tylerphelan, line 22: Should we consider using `syscall.Exec` here? this would completely replace the network-wait-launcher Go process with the launcher process instead of spawning a new one. 

1. Maybe more closely mirrors linux buildpods
2. Are there edge cases around using subprocesses here?
  - Exit codes
  - Process interrupts
  - Error messages
3. Does `syscall.Exec` work on windows?

tylerphelan: it is [not supported](https://github.com/golang/go/issues/30662) on windows.. but maybe worth thinking about.

matthewmcnew: We should still think through exit codes and process interrupts.

pkg/git/fetch.go
tomkennedy513: I ended up putting it in the commit message, but happy to add a comment if you think we still need one

api/openapi-spec/swagger.json
tomkennedy513: this was fixed after rerunning the openapi codegen script with the updated binary

aemengo (approved):

pkg/git/fetch.go
tomkennedy513: I have mixed opinions on this. On one hand I don't like seeing duplicate code, but on the other hand it might be unnecessary to extract one function to a new package if it's only being used in two places.

sukhil-suresh (approved):

tylerphelan (approved):
amazing work!

cmd/network-wait-launcher/main.go
tylerphelan: I'm going to put this in a separate issue

Issue #536, opened 2020-10-27
Support k8s Service Bindings


matthewmcnew:
With [CNB Service bindings deprecated ](https://github.com/buildpacks/rfcs/pull/105) the new [k8s service bindings spec](https://k8s-service-bindings.github.io/spec/) should be supported. 

The CF Capi team needs this functionality to add service bindings to cf pushed apps during build. 

The outstanding question is how to model this in the image spec and in what timeframe we will deprecate the old service bindings.

djoyahoy:
The new spec should be low effort to support. We would need to do the following:
1. Modify the image config to take a list of objectRefs instead of secrets + configmaps for service bindings.
2. a. During build reconciliation, find the service objects from the objectRefs list, and use their binding names to mount secret volumes into the build pod.
2. b. Use the `SERVICE_BINDING_ROOT` env var from the image config as the mount path root if provided. Otherwise, use `/bindings`.
2. c. Everything else remains the same.

To add support for this feature, I would recommend bumping the image config version from `v1alpha1` to `v1alpha2` or `v1beta1` and dropping the support for CNB style bindings in the latest image config version.

Alternatively, we simply add a new field to the image config and use the same version, while maintaining support for both binding specs. This may cause some confusion for our users.

tylerphelan:
Migration path Image v1alpha1 -> v1alpha2

Old bindings config:

```
kind: Image
spec:
  build:
    bindings:
    - name: settings
       secretRef:
         name: settings-xml
       metadataRef:
         name: settings-binding-metadata
---
apiVersion: v1
kind: Secret
metadata:
  name: settings-xml
type: Opaque
stringData:
  settings.xml: <settings>...</settings>
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: settings-binding-metadata
data:
  kind: maven
  provider: sample
```

Directory structure of old bindings:

```
<platform>
└── bindings
    └── settings
        ├── metadata
        │   ├── kind
        │   └── provider
        └── secret
            └── setting.xml
```

The same config should create the following directory structure for new bindings:

```
$SERVICE_BINDING_ROOT
├── settings
│   ├── type
│   ├── provider
│   ├── setting.xml
```

Doesn't seem too complex

matthewmcnew:
It seems like the complication is how to handle to represent both the old/new service bindings in both  v1alpha1/v1alpha2 api versions. 

Also of note, this information needs to be represented on the build as well.

[...]

PR #555, opened 2020-11-20
Support k8s service bindings

tylerphelan:
Issue: https://github.com/pivotal/kpack/issues/536

- Add v1alpha2 for Builds and Images
- Add conversion webhook for crds
- Continue to support v1alpha1 bindings in controller
- Small breaking change in v1alpha1 api code (lastSource)
- Conversion strategy for v1alpha1 -> v1alpha2 is via annotation
- Do not remove any v1alpha1 api code

Open questions:
- Does this adequately support `ProvisionedService`s? 
  ~- Is there a race condition between `ProvisionedService`s and `Build`s reconciling? (if a `ProvisionedService` is not `Ready` yet will the `Build` fail?)~
  - This uses `meta.UnsafeGuessKindToResource`, is that alright?
  - Would love to hear thoughts.
- Went with `v1alpha2` but we could do `v1beta1`
- Should we update the docs now? Or wait for the release so that `master` branch does not have inaccurate docs?
- I added validation in the `generator.go` to check that service binding secrets include the required `type` and `data.type` based on the [spec](https://github.com/k8s-service-bindings/spec#provisioned-service). Would love thoughts on that. Is it necessary?

Notes:
- In order to integrate a `ProvisionedService` with `kpack` users will need to create rbac that allows the `kpack-controller` to `get` the resource via the [spec](https://github.com/k8s-service-bindings/spec#for-cluster-operators-and-crd-authors)
  - We should document this


scothis:
> if a ProvisionedService is not Ready yet will the Build fail?

ProvisionedService has no notion of "Ready" either there is a Secret that can be read, or there isn't

pkg/buildpod/generator.go
```diff
@@ -154,6 +166,46 @@ func (g *Generator) fetchBuilderConfig(build BuildPodable) (v1alpha1.BuildPodBui
 	}, nil
 }
 
+func (g *Generator) fetchServiceBindings(build BuildPodable) ([]v1alpha2.ServiceBinding, error) {
+	bindings := []v1alpha2.ServiceBinding{}
+	if len(build.Services()) != 0 {
+		for _, s := range build.Services() {
+			if s.Kind == "Secret" {
```
scothis, line 173:
```suggestion
			if s.APIVersion == "v1" && s.Kind == "Secret" {
```

scothis (commented):
Perhaps I missed it, but I didn't see where the "application projection" is occurring. The Service Binding for K8s Spec is not the CNB Binding Spec minus a ConfigMap. Check out https://github.com/k8s-service-bindings/spec#application-projection

RBAC can be defined generically, take a look at https://github.com/k8s-service-bindings/spec#for-service-binding-implementors

tylerphelan:
Many thanks @scothis for looking at this! Made some changes from the feedback.

> Perhaps I missed it, but I didn't see where the "application projection" is occurring

I was missing the `type` and `data.type` requirements previously but the [apis/build/v1alpha2/build_pod.go](https://github.com/pivotal/kpack/pull/555/files#diff-796b50408de7dbf79786503ffc9edfbc1eb29abc4c36902f37c6c5714606333b) should implement the mounting as I understand it.

config/controllerrole.yaml
```diff
@@ -9,6 +9,10 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: kpack-controller-admin
+aggregationRule:
```
scothis, line 12: split `rules` and `aggreationRule` to use different roles. The aggregation will overwrite the rules rather then augment them.

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -0,0 +1,652 @@
...
+			)
+			volumeMounts = append(volumeMounts,
+				corev1.VolumeMount{
+					Name:      metadataVolume,
+					MountPath: filepath.Join(platformVolume.MountPath, "bindings", s.Name, "metadata"),
+					ReadOnly:  true,
+				},
+			)
+		}
+	}
```
scothis, line 540: There is no "metadata" directory in the k8s service binding spec, and the secret is in the wrong place for the cnb service binding spec.

rm unless this is doing something meaningful that I don't understand.

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -0,0 +1,652 @@
...
+
+	return volumes, volumeMounts
+}
+
+func (b *Build) setupV1Bindings() ([]corev1.Volume, []corev1.VolumeMount, error) {
```
scothis, line 545: not sure what a V1Binding is, did you mean V1alpha1?

```suggestion
func (b *Build) setupV1alpha1Bindings() ([]corev1.Volume, []corev1.VolumeMount, error) {
```

pkg/apis/build/v1alpha2/build_validation.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+func (b *Build) SetDefaults(ctx context.Context) {
+	if b.Spec.ServiceAccount == "" {
+		b.Spec.ServiceAccount = "default"
+	}
+
+	for i, s := range b.Spec.Services {
+		if s.Kind == "Secret" && s.APIVersion == "" {
+			b.Spec.Services[i].APIVersion = "v1"
```
scothis, line 21: is there precedent for defaulting the APIVersion in an object reference?

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -0,0 +1,652 @@
...
+		ReadOnly:  true,
+	}
+	builderPullSecretsVolume = corev1.VolumeMount{
+		Name:      builderPullSecretsDirName,
+		MountPath: "/builderPullSecrets",
+		ReadOnly:  true,
+	}
+	serviceBindingRootEnv = corev1.EnvVar{
+		Name:  serviceBindingRootEnvVar,
+		Value: filepath.Join(platformVolume.MountPath, "bindings"),
```
scothis, line 100: K8s service bindings and cnb service bindings should not share the same directory.

pkg/apis/build/v1alpha2/build_validation.go
tylerphelan: This was just me speculating that this might help usability

pkg/apis/build/v1alpha2/build_pod.go
tylerphelan: total oversight

pkg/apis/build/v1alpha2/build_pod.go
tylerphelan: As I understand it, buildpacks will only read from one location for bindings based on [this]( https://github.com/buildpacks/libcnb/blob/main/build.go#L166) and a quick convo with @nebhale 

Given that, I made build pods only project CNB bindings OR K8s bindings in which case why not use the same directory?

Curious to peoples' thoughts

pkg/apis/build/v1alpha2/build_pod.go
tylerphelan: yea I did this wrong, fixing

tylerphelan: this should be fixed now

pkg/apis/build/v1alpha2/build_pod.go
djoyahoy: @tylerphelan that makes sense to me. I always assumed it would be either/or and not both.

matthewmcnew:
@tylerphelan  If we are holding off on this for a bit could we move it draft?

pkg/buildpod/generator.go
```diff
@@ -10,11 +10,16 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/dynamic"
 	k8sclient "k8s.io/client-go/kubernetes"
+	"knative.dev/pkg/apis/duck"
 
 	"github.com/pivotal/kpack/pkg/apis/build/v1alpha1"
+	"github.com/pivotal/kpack/pkg/apis/build/v1alpha2"
```
c0d1ngm0nk3y, line 20: Is it really the best approach to mix `build/v1alpha1` and `build/v1alpha2`? It can get very tricky which struct to take from which import and the errors can be very subtle.

Another approach would bump the  api as a whole and only use `build/v1alpha2`. Maybe even use a import like

```bash
buildapi "github.com/pivotal/kpack/pkg/apis/build/v1alpha2"
```

So the code didn't have to be changed all over if a version is bumped.

matthewmcnew:
Tagging this as `0.4.0` to target releasing alongside the api changes in our `0.4.0` release.

tomkennedy513:
closed in favor of #823

PR #556, opened 2020-11-23
Build Reasons and Changes diff

sukhil-suresh:
The PR updates the `prepare` and `rebase` stages of the Build process to display the Build Reason and corresponding Change diff.
- Sample output captured in [this comment](https://github.com/pivotal/kpack/pull/556#issuecomment-732460239)
- Build Reason sort index is currently set to "TRIGGER,COMMIT,CONFIG,BUILDPACK,STACK"
- Username was dropped from the TRIGGER message (at least for the first iteration) after discussion


djoyahoy:
@sukhil-suresh is there an environment variable limit that would prevent use from storing all changes in the `BUILD_CHANGES` env var?

sukhil-suresh:
> @sukhil-suresh is there an environment variable limit that would prevent use from storing all changes in the `BUILD_CHANGES` env var?

Yes, there is a size limit on linux per environment variable and I believe for the combined block of env variables. I don't have a definitive number on hand; but based on google search it seems to be a 128KiB limit. It is also configurable.

The alternative considered was the [Downward API Volume files](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/) but opted not to go down that path based on some very rough math of the changes value maxing out between 1500..2000 chars (at the moment)

But yes, still open to changing approach of passing the Changes value around

djoyahoy:
> > @sukhil-suresh is there an environment variable limit that would prevent use from storing all changes in the `BUILD_CHANGES` env var?
> 
> Yes, there is a size limit on linux per environment variable and I believe for the combined block of env variables. I don't have a definitive number on hand; but based on google search it seems to be a 128KiB limit. It is also configurable.
> 
> The alternative considered was the [Downward API Volume files](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/) but opted not to go down that path based on some very rough math of the changes value maxing out between 1500..2000 chars (at the moment)
> 
> But yes, still open to changing approach of passing the Changes value around

That makes sense, thanks!

sukhil-suresh:
Captured Build reason(s) and Change diff for the different Build Reasons: CONFIG, TRIGGER, COMMIT, STACK, and BUILDPACK
The last one is for a combination of TRIGGER and BUILDPACK 

**CONFIG** 
```
$ kp image create test-image --tag sukhilsuresh/test-image --git https://github.com/paketo-buildpacks/samples --git-revision main --sub-path go/mod
Creating Image...
Image "test-image" created

$ kp build logs test-image
===> PREPARE
Build reason(s): CONFIG
CONFIG change:
	resources:
	  Limits: null
	  Requests: null
	source:
	  Blob: null
	-   Git: null
	+   Git:
	+     revision: 948b2eff6a21580a44a0f4d8c609a2af45359d41
	+     url: https://github.com/paketo-buildpacks/samples
	  Registry: null
	-   SubPath: ""
	+   SubPath: go/mod
Loading secret for "gcr.io" from secret "gcr" at location "/var/build-secrets/gcr"
....
```

**TRIGGER**
```
$ kp image trigger test-image
Triggered build for Image "test-image"

$ kp build logs test-image
===> PREPARE
Build reason(s): TRIGGER
TRIGGER change:
	A new build was manually triggered on Mon, 23 Nov 2020 16:16:26 -0500
Loading secret for "gcr.io" from secret "gcr" at location "/var/build-secrets/gcr"
...
```

**COMMIT**
```
$ kp build logs test-image
===> PREPARE
Build reason(s): COMMIT
COMMIT change:
	- 948b2eff6a21580a44a0f4d8c609a2af45359d41
	+ ec6f737864fa12374d5e8ee1db09058ab4e21ebf
Loading secret for "gcr.io" from secret "gcr" at location "/var/build-secrets/gcr"
...
```

**STACK**
```
$ kp clusterstack update full -b registry.pivotal.io/tbs-dependencies/build-full@sha256:e60c6717b2b6e8458f8f782c13fd2f60ef9e2f1f378fbb219be5da13538114aa -r registry.pivotal.io/tbs-dependencies/run-full@sha256:42841631725942db48b7ba8b788b97374a2ada34c84ee02ca5e02ef3d4b0dfca
Updating ClusterStack...
Uploading to 'gcr.io/cf-build-service-dev-219913/ssuresh/install'...
	Uploading 'gcr.io/cf-build-service-dev-219913/ssuresh/install/build@sha256:e60c6717b2b6e8458f8f782c13fd2f60ef9e2f1f378fbb219be5da13538114aa'
	Uploading 'gcr.io/cf-build-service-dev-219913/ssuresh/install/run@sha256:42841631725942db48b7ba8b788b97374a2ada34c84ee02ca5e02ef3d4b0dfca'


$ kp build logs test-image
===> REBASE
Build reason(s): STACK
STACK change:
	- RunImage: sha256:87302783be0a0cab9fde5b68c9954b7e9150ca0d514ba542e9810c3c6f2984ad
	+ RunImage: sha256:42841631725942db48b7ba8b788b97374a2ada34c84ee02ca5e02ef3d4b0dfca
Loading secret for "gcr.io" from secret "gcr" at location "/var/build-secrets/gcr"
...
```

**BUILDPACK**
```
$ kp clusterstore status default
Status:    Ready

BUILDPACKAGE ID                       VERSION    HOMEPAGE
paketo-buildpacks/procfile            2.0.2      https://github.com/paketo-buildpacks/procfile
tanzu-buildpacks/dotnet-core          0.0.6
tanzu-buildpacks/go                   1.0.5
tanzu-buildpacks/httpd                0.0.39
tanzu-buildpacks/java                 3.8.0      https://github.com/pivotal-cf/tanzu-java
tanzu-buildpacks/java-native-image    3.6.0      https://github.com/pivotal-cf/tanzu-java-native-image
tanzu-buildpacks/nginx                0.0.46
tanzu-buildpacks/nodejs               1.2.2
tanzu-buildpacks/php                  0.0.5

$ kp clusterstore add default -b registry.pivotal.io/tanzu-go-buildpack/go@sha256:ed99e73df43380c86c983e09d97151caf1952e8f423053a6e0a15de526515f4c
Adding to ClusterStore...
	Uploading 'gcr.io/cf-build-service-dev-219913/ssuresh/install/tanzu-buildpacks_go@sha256:ed99e73df43380c86c983e09d97151caf1952e8f423053a6e0a15de526515f4c'
	Added Buildpackage
ClusterStore "default" updated

$ kp clusterstore status default
Status:    Ready

BUILDPACKAGE ID                       VERSION    HOMEPAGE
paketo-buildpacks/procfile            2.0.2      https://github.com/paketo-buildpacks/procfile
tanzu-buildpacks/dotnet-core          0.0.6
tanzu-buildpacks/go                   1.0.5
tanzu-buildpacks/go                   1.0.7
tanzu-buildpacks/httpd                0.0.39
tanzu-buildpacks/java                 3.8.0      https://github.com/pivotal-cf/tanzu-java
tanzu-buildpacks/java-native-image    3.6.0      https://github.com/pivotal-cf/tanzu-java-native-image
tanzu-buildpacks/nginx                0.0.46
tanzu-buildpacks/nodejs               1.2.2
tanzu-buildpacks/php                  0.0.5

$ kp build logs test-image
===> PREPARE
Build reason(s): BUILDPACK
BUILDPACK change:
	- tanzu-buildpacks/go-build	0.0.22
	- tanzu-buildpacks/go-dist	0.1.0
	- tanzu-buildpacks/go-mod-vendor	0.0.24
	+ tanzu-buildpacks/go-build	0.0.23
	+ tanzu-buildpacks/go-dist	0.1.2
	+ tanzu-buildpacks/go-mod-vendor	0.0.25
Loading secret for "gcr.io" from secret "gcr" at location "/var/build-secrets/gcr"
...
```

**Combination of TRIGER and BUILDPACK**
```
$ kp build logs test-image
===> PREPARE
Build reason(s): TRIGGER,BUILDPACK
TRIGGER change:
	A new build was manually triggered on Mon, 23 Nov 2020 17:14:53 -0500
BUILDPACK change:
	- tanzu-buildpacks/go-build	0.0.23
	- tanzu-buildpacks/go-dist	0.1.2
	- tanzu-buildpacks/go-mod-vendor	0.0.25
	+ tanzu-buildpacks/go-build	0.0.22
	+ tanzu-buildpacks/go-dist	0.1.0
	+ tanzu-buildpacks/go-mod-vendor	0.0.24
```

sampeinado:
This looks AWESOME !!

A few comments: 
- for trigger, it says "TRIGGER change:" which doesn't really make sense. Would it be possible to say just "TRIGGER"?
- for the CONFIG change, how does the system decide how many additional lines to print other than the diffed lines?

sukhil-suresh:
> * for trigger, it says "TRIGGER change:" which doesn't really make sense. Would it be possible to say just "TRIGGER"?


@sampeinado Assuming you were referring to a format similar to the one below? Yes, can make that change 
```
===> PREPARE
Build reason(s): TRIGGER,BUILDPACK
TRIGGER: A new build was manually triggered on Mon, 23 Nov 2020 16:16:26 -0500
BUILDPACK change:
	- tanzu-buildpacks/go-build	0.0.23
	- tanzu-buildpacks/go-dist	0.1.2
	- tanzu-buildpacks/go-mod-vendor	0.0.25
	+ tanzu-buildpacks/go-build	0.0.22
	+ tanzu-buildpacks/go-dist	0.1.0
	+ tanzu-buildpacks/go-mod-vendor	0.0.24
```

sampeinado:
yes exactly! I think that makes more sense

sukhil-suresh:
@sampeinado 
> * for the CONFIG change, how does the system decide how many additional lines to print other than the diffed lines?

For context...
```
$ kp build logs test-image
===> PREPARE
Build reason(s): CONFIG
CONFIG change:
	resources:
	  Limits: null
	  Requests: null
	source:
	  Blob: null
	-   Git: null
	+   Git:
	+     revision: 948b2eff6a21580a44a0f4d8c609a2af45359d41
	+     url: https://github.com/paketo-buildpacks/samples
	  Registry: null
	-   SubPath: ""
	+   SubPath: go/mod
Loading secret for "gcr.io" from secret "gcr" at location "/var/build-secrets/gcr"
...
```

I am assuming you are referring to the `resources:` being listed. I do see the oddity of it since it provides no additional context. I am not sure why that gets listed (in spite of the effort to suppress it). Will look more into it.

About how the diff is generated? The [Config objects](https://github.com/pivotal/kpack/blob/552762a8a09f7f4cf9eb5023e40d823f4cdd2926/pkg/buildchange/change.go#L48-L58) are serialized to string and then compared. [Test case](https://github.com/pivotal/kpack/blob/552762a8a09f7f4cf9eb5023e40d823f4cdd2926/pkg/buildchange/change_logger_test.go#L159-L260) that covers the CONFIG scenario.

sukhil-suresh:
> A few comments:
> 
> * for trigger, it says "TRIGGER change:" which doesn't really make sense. Would it be possible to say just "TRIGGER"?
> * for the CONFIG change, how does the system decide how many additional lines to print other than the diffed lines?

Updated CONFIG, and TRIGGER change (based on @sampeinado comment)

**CONFIG**
```
$ kp image create test-image --tag sukhilsuresh/test-image --git https://github.com/paketo-buildpacks/samples --git-revision main --sub-path go/mod
Creating Image...
Image "test-image" created

$ kp build logs test-image
===> PREPARE
Build reason(s): CONFIG
CONFIG change:
	- source: {}
	+ source:
	+   git:
	+     revision: 948b2eff6a21580a44a0f4d8c609a2af45359d41
	+     url: https://github.com/paketo-buildpacks/samples
	+   subPath: go/mod
Loading secret for "gcr.io" from secret "gcr" at location "/var/build-secrets/gcr"
...
```

**TRIGGER**
```
$ kp image trigger test-image
Triggered build for Image "test-image"

$ kp build logs test-image
===> PREPARE
Build reason(s): TRIGGER
TRIGGER: A new build was manually triggered on Tue, 24 Nov 2020 21:12:39 -0500
Loading secret for "gcr.io" from secret "gcr" at location "/var/build-secrets/gcr"
...
```

pkg/apis/build/v1alpha1/image_builds.go
```diff
@@ -17,18 +16,35 @@ const (
...
+	BuildReasonAnnotation  = "image.kpack.io/reason"
+	BuildChangesAnnotation = "image.kpack.io/buildChanges"
+	BuildNeededAnnotation  = "image.kpack.io/additionalBuildNeeded"
 
 	BuildReasonConfig    = "CONFIG"
 	BuildReasonCommit    = "COMMIT"
 	BuildReasonBuildpack = "BUILDPACK"
 	BuildReasonStack     = "STACK"
 	BuildReasonTrigger   = "TRIGGER"
+	BuildReasonSortIndex = "TRIGGER,COMMIT,CONFIG,BUILDPACK,STACK"
```
tylerphelan, line 28: can this be a map?

sukhil-suresh: confused as to how changing this to a map would help, Tyler? 😕 
It is currently used [here](https://github.com/pivotal/kpack/blob/9fcfdc89ea5a60ffde2506cf06ba1bc8e8cf749b/pkg/buildchange/change_processor.go#L72-L86) for sorting purpose only

sukhil-suresh:
We as a team talked about the pros/cons of a thick vs thin logger and **opted to adopt a thin logger**.

Thick logger as in the current commit does validation on the reasons and changes string by unmarshalling it back into the changes object. Besides, it also makes sure there is consistency between the reasons and changes. For corrupt data, it errors.

The thin logger will only use the changes string, which will be unmarshalled into map[string]struct{old: string, new, string}

Also, it was the consensus that the decoupling of change data from the buildchange.logger outweighed the benefits of keeping build change data distinct from the change presentation (as is currently done).

matthewmcnew:
@sukhil-suresh 

Could the thin logger unmarshall into `map[string]struct{old: interface{}, new, interface{}}`?

matthewmcnew:
Also might be worth considering a list instead of a map like `struct{name: string, old: interface{}, new, interface{}}[]`

sukhil-suresh:
> Could the thin logger unmarshall into `map[string]struct{old: interface{}, new, interface{}}`?

sorry that was my bad/typo thing... I was using the structure that you suggested in my [thin logger experimental branch](https://github.com/pivotal/kpack/blob/d8c064401b1106557c8d065fa988e7184990197c/pkg/buildchange/thin_logger.go#L19-L22)

sukhil-suresh:
> Also might be worth considering a list instead of a map like `struct{name: string, old: interface{}, new, interface{}}[]`

yes, it does make sense. especially to keep the sorting order, since we will be using only the changes string.

pkg/apis/build/v1alpha1/image_builds.go
tylerphelan:
```
BuildReasonSortIndex := map[string]int{"TRIGGER":0,"COMMIT":1,"CONFIG":2,"BUILDPACK":3,"STACK":4}
sort.SliceStable(reasons, func(i, j int) bool {
		return BuildReasonSortIndexMap[v1alpha1.BuildReasonSortIndex][reasons[i]] <
			BuildReasonSortIndexMap[v1alpha1.BuildReasonSortIndex][reasons[i]])
	})
```

sukhil-suresh: @tylerphelan thanks for clarifying what you mean. Honestly I don't see considerable pros/cons between the 2 approaches, given that the BuildReasonSortIndex string is very small. I actually prefer the current implementation because the SortIndex is declared right next to the list of BuildReasons (and is a form of documentation)

https://github.com/pivotal/kpack/blob/072cf212feea355d2ec01d56a6bae4b13a6cf7dd/pkg/apis/build/v1alpha1/image_builds.go#L23-L28

the way they are used now
https://github.com/pivotal/kpack/blob/072cf212feea355d2ec01d56a6bae4b13a6cf7dd/pkg/buildchange/change_processor.go#L77-L111

sukhil-suresh:
Listing the different Build reason and changes logging, after switching implementation to the [light weight logger](https://github.com/pivotal/kpack/pull/556#issuecomment-736015627)

**CONFIG**
```
===> PREPARE
Build reason(s): CONFIG
CONFIG change:
	resources: {}
	- source: {}
	+ source:
	+   git:
	+     revision: 2a3513e88251624fb45abcd2f3454511944960b3
	+     url: https://github.com/paketo-buildpacks/samples
	+   subPath: go/mod
Loading secret for "gcr.io" from secret "gcr" at location "/var/build-secrets/gcr"
...
```

**TRIGGER**
```
===> PREPARE
Build reason(s): TRIGGER
TRIGGER: A new build was manually triggered on Mon, 07 Dec 2020 18:11:30 -0500
Loading secret for "gcr.io" from secret "gcr" at location "/var/build-secrets/gcr"
...
```

**STACK**
```
===> REBASE
Build reason(s): STACK
STACK change:
	- sha256:16515ead81f41ecdb8b391ab931853138425bb4b75486c0bc434531fb67ca2e5
	+ sha256:42841631725942db48b7ba8b788b97374a2ada34c84ee02ca5e02ef3d4b0dfca
Loading secret for "gcr.io" from secret "gcr" at location "/var/build-secrets/gcr"
...	
```

**BUILDPACK**
```
===> PREPARE
Build reason(s): BUILDPACK
BUILDPACK change:
	- id: tanzu-buildpacks/go-build
	-   version: 0.0.22
	+   version: 0.0.23
	- id: tanzu-buildpacks/go-dist
	-   version: 0.1.0
	+   version: 0.1.2
	- id: tanzu-buildpacks/go-mod-vendor
	-   version: 0.0.24
	+   version: 0.0.25
Loading secret for "gcr.io" from secret "gcr" at location "/var/build-secrets/gcr"
...
```

**Combination of TRIGGER and BUILDPACK**
```
===> PREPARE
Build reason(s): TRIGGER,BUILDPACK
TRIGGER: A new build was manually triggered on Mon, 07 Dec 2020 18:29:42 -0500
BUILDPACK change:
	- id: tanzu-buildpacks/go-build
	-   version: 0.0.23
	+   version: 0.0.22
	- id: tanzu-buildpacks/go-dist
	-   version: 0.1.2
	+   version: 0.1.0
	- id: tanzu-buildpacks/go-mod-vendor
	-   version: 0.0.25
	+   version: 0.0.24
Loading secret for "gcr.io" from secret "gcr" at location "/var/build-secrets/gcr"
...
```

pkg/reconciler/image/build_needed.go
```diff
@@ -1,85 +1,156 @@
...
+	return buildchange.NewTriggerChange(time)
+}
+
+func (b *BuildDeterminer) commitChange() buildchange.Change {
+	if b.lastBuild == nil || b.srcResolver.Status.Source.Git == nil {
+		return nil
 	}
 
-	return reasons, corev1.ConditionTrue
+	oldRevision := b.lastBuild.Spec.Source.Git.Revision
```
matthewmcnew, line 89: What if the last build was not a git build?

pkg/reconciler/image/reconcile_build.go
```diff
@@ -18,16 +18,22 @@ func (c *Reconciler) reconcileBuild(image *v1alpha1.Image, latestBuild *v1alpha1
 		return v1alpha1.ImageStatus{}, err
 	}
 
-	reasons, needed := buildNeeded(image, latestBuild, sourceResolver, builder)
-	switch needed {
+	buildDeterminer := NewBuildDeterminer(image, latestBuild, sourceResolver, builder)
```
matthewmcnew, line 21: In golang world the construction of a `BuildDeterminer` object seems a bit unnecessary. Could we just call a method called determineBuild which returns a build result that knows if a build is needed, etc?

pkg/buildchange/stack_change.go
```diff
@@ -0,0 +1,50 @@
+package buildchange
```
matthewmcnew, line 1: I like the ability to model each of these BuildReason's on their own.

pkg/buildchange/change_processor.go
```diff
@@ -0,0 +1,111 @@
...
+	changes []GenericChange
+	errStrs []string
+}
+
+func (c *ChangeProcessor) Process(change Change) *ChangeProcessor {
+	if change == nil {
+		return c
+	}
+
+	if !change.Reason().IsValid() {
```
matthewmcnew, line 36: Is there a reason to validate this? It seems like the change processor does not need care about the change being a `known` type.

pkg/buildchange/change_processor.go
```diff
@@ -0,0 +1,111 @@
...
+		errStr := fmt.Sprintf("unsupported change reason '%s'", change.Reason())
+		c.errStrs = append(c.errStrs, errStr)
+		return c
+	}
+
+	valid, err := change.IsValid()
```
matthewmcnew, line 42: While reading this PR, I kept getting a bit confused by the term `isValid`. It seems that `IsValid` is essentially asking if the change is necessary for a new build. Could we do something like `RequiresRebuild` or something?

pkg/buildchange/change_logger.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (c *changeLogger) logReasons() {
+	reasons := strings.Join(c.reasons, reasonsSeparator)
+	c.logger.Printf("Build reason(s): %s\n", reasons)
+}
+
+func (c *changeLogger) logChanges() error {
+	for _, change := range c.changes {
+		if change.Reason == "TRIGGER" {
```
matthewmcnew, line 73: Is this Trigger check needed?

pkg/buildchange/change_processor.go
```diff
@@ -0,0 +1,111 @@
...
+
+	return strings.Join(reasons, reasonsSeparator)
+}
+
+func (c *ChangeProcessor) changesStr() (string, error) {
+	if !c.hasChanges() {
+		return "", nil
+	}
+
+	sort.SliceStable(c.changes, func(i, j int) bool {
```
matthewmcnew, line 100: Is this sorting necessary?

pkg/buildchange/stack_change.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	if len(errStrs) > 0 {
+		change.err = errors.New(strings.Join(errStrs, "; "))
+	}
+	return change
+}
+
+type stackChange struct {
+	oldRunImageId string
```
matthewmcnew, line 37: Could we call these digests instead of id?

pkg/buildchange/change_processor.go
sukhil-suresh: > Is this sorting necessary?

The sorting is done to ensure that the reasons string and the corresponding changes are listed in the same order.
A safety, to ensure that the Process() change call sequence on the ChangeProcessor does not determine the output order.
I think it is a small overhead to ensure changes on the build_needed.go file does not inadvertently change the order for printing changes.

pkg/buildchange/change_logger.go
sukhil-suresh: > Is this Trigger check needed?

The TRIGGER check was done to match @sampeinado 's requirement on how to display aTRIGGER change
Happy to remove the conditional check If displaying a TRIGGER change as a diff like the other changes is acceptable

pkg/buildchange/change_processor.go
sukhil-suresh: Yes, it can be avoided since all the changes are defined by us anyway. Will take it out.

pkg/reconciler/image/build_needed.go
sukhil-suresh: Aye. Will fix it.

pkg/buildchange/stack_change.go
sukhil-suresh: It was called `oldRunImageId` because it holds the image identifier (which happens to be a digest) 
```
change.oldRunImageId = oldRunImageRef.Identifier()
```

pkg/buildchange/change_processor.go
sukhil-suresh: resolved by this commit - https://github.com/pivotal/kpack/pull/556/commits/3a47900f134f03876914ae0a94788698dd7dca45

pkg/buildchange/stack_change.go
sukhil-suresh: marking as resolved. made the change anyway with commit https://github.com/pivotal/kpack/commit/6bcb2c1a346ae2149b5fd1d2681cf759f93ae3a7

pkg/buildchange/change_logger.go
sukhil-suresh: @matthewmcnew and @sampeinado : please let me know how to proceed on this. I don't feel strongly about going either way.

pkg/apis/build/v1alpha1/image_builds.go
sukhil-suresh: @tylerphelan please let me know if you feel strongly about this. I am open to discussion.

If not, would like resolve this conversation.

pkg/buildchange/change_processor.go
sukhil-suresh: @matthewmcnew please let me know if you still consider this as unnecessary. I would prefer to keep it. Either way, would like to resolve this comment.

pkg/buildchange/change_processor.go
sukhil-suresh: @matthewmcnew please let me know if you still consider this as not necessary. I would prefer to keep it. Either way, would like to resolve this comment.

pkg/reconciler/image/reconcile_build.go
sukhil-suresh: marking as resolved. addressed in https://github.com/pivotal/kpack/commit/e6e1245a58ff31c160a2607f6f7d9aac1092993f

pkg/apis/build/v1alpha1/image_builds.go
tylerphelan: `BuildReasonSortIndex` is sufficiently small so performance shouldn't really matter

pkg/reconciler/image/build_needed.go
matthewmcnew: Couldn't the current build be a git build but the last build be a blob build for example?

sukhil-suresh: yikes... I definitely missed that. will fix. thanks for pointing it out

pkg/buildchange/change_processor.go
matthewmcnew: I seems like a bit of an unnecessary open/closed violation for the change processor to need to know about the known types of acceptable buildReasons. I don't think we need to be concerned about internals of kpack producing a random buildReason string as this does not come from a user input.

pkg/buildchange/change_processor.go
matthewmcnew: Similar to the [validation discussion](https://github.com/pivotal/kpack/pull/556/files#r538868069) it seems like the change processor shouldn't need to understand what types it is processing. It seems like the order of the calls to `Process()` should be sufficent to enforce order.

pkg/buildchange/change_logger.go
matthewmcnew: Hmm. It seems like we should try to avoid treating build changes individually if that is at all possible. 

Would it be possible to just drop the phrase `change:`?

matthewmcnew: So, that could look like: 

```
Build reason(s): TRIGGER,BUILDPACK
TRIGGER: A new build was manually triggered on Mon, 23 Nov 2020 16:16:26 -0500
BUILDPACK:
	- tanzu-buildpacks/go-build	0.0.23
...

```

sukhil-suresh: Fine.. will treat them all the same. It won't look exactly like how you listed. more like this
```
```

sukhil-suresh: Fine.. will treat them all the same. It won't look exactly like how you listed. more like this
```
Build reason(s): TRIGGER,BUILDPACK
TRIGGER:
        + A new build was manually triggered on Mon, 07 Dec 2020 18:29:42 -0500
BUILDPACK:
	- id: tanzu-buildpacks/go-build
	-   version: 0.0.23
	+   version: 0.0.22
	- id: tanzu-buildpacks/go-dist
	-   version: 0.1.2
	+   version: 0.1.0
	- id: tanzu-buildpacks/go-mod-vendor
	-   version: 0.0.25
	+   version: 0.0.24
```

just a general observation... bending/altering design requirements for avoiding special handling (which is not a big deal) in implementation is a questionable approach.

However, I do agree that In this scenario, the output does look good even if the changes are handled the same way.

pkg/buildchange/change_processor.go
sukhil-suresh: I do not agree with your observation and am hesitant to make change which I am not convinced is correct.  Maybe we should chat about this?

The ChangeProcess processes Changes for what? Producing the Reasons and Changes string. It is the output of the Summarize call on the ChangeProcessor. Considering that, I disagree with your premise that the ChangeProcessor should not be responsible for ensuring the reasons and changes string are sorted the same way.

pkg/buildchange/change_processor.go
sukhil-suresh: [This comment](https://github.com/pivotal/kpack/pull/556#discussion_r540403788) applies here too

pkg/buildchange/change_processor.go
sukhil-suresh: > It seems like the order of the calls to Process() should be sufficent to enforce order.

I did address this in my [prior comment](https://github.com/pivotal/kpack/pull/556#discussion_r538882526)

At least 2 ways that I can see foresee change order going out of sync with reasons (by oversight)
- BuildReasonSortIndex is updated without a similar update to the Process() call sequence
- Future PR/Code changes which involve a reorder of the Process() call sequence without an update to the BuildReasonSortIndex

I feel an implicit dependence on the Process() call sequence (vs the current implementation of explicit ordering of reasons and changes) makes it very fragile to unintended breakage - i.e. the reasons listing and changes listing go out of sequence.

pkg/reconciler/image/build_needed.go
sukhil-suresh: resolved by https://github.com/pivotal/kpack/commit/be3ad6dd90eb514a1ce408e7c72bb7f649a74ff2

pkg/buildchange/change_logger.go
sukhil-suresh: resolved by https://github.com/pivotal/kpack/commit/be3ad6dd90eb514a1ce408e7c72bb7f649a74ff2

tylerphelan (approved):

pkg/buildchange/change_processor.go
sukhil-suresh: Update from discussion: `Is this sorting necessary?` was addressing both the sorting of changes and reasons within ChangeProcessor; which would mean the possibility for reasons string and change string going out of sync is removed.
Moving ahead with dropping sort of both changes and reason in change_processor

sukhil-suresh: addressed in https://github.com/pivotal/kpack/commit/e73935fa92379241fdd65e1d2d916a4ae487afc5

pkg/buildchange/change_processor.go
sukhil-suresh: addressed in https://github.com/pivotal/kpack/commit/e73935fa92379241fdd65e1d2d916a4ae487afc5

djoyahoy (approved):

pkg/reconciler/image/build_required.go
```diff
@@ -0,0 +1,137 @@
...
+	time, ok := lastBuild.Annotations[v1alpha1.BuildNeededAnnotation]
+	if !ok {
+		return nil
+	}
+	return buildchange.NewTriggerChange(time)
+}
+
+func commitChange(lastBuild *v1alpha1.Build, srcResolver *v1alpha1.SourceResolver) buildchange.Change {
+	// If the lastBuild was not a Git source, then it is not a COMMIT change
+	if lastBuild == nil || lastBuild.Spec.Source.Git == nil || srcResolver.Status.Source.Git == nil {
```
matthewmcnew, line 66: Is the git commit still displayed when the lastBuild was not a git source build but, the current build is a git commit source?

matthewmcnew (approved):

tomkennedy513 (approved):

pkg/reconciler/image/build_required.go
sukhil-suresh: >  when the lastBuild was not a git source build but, the current build is a git commit source

the above is treated as a CONFIG change and not a COMMIT change

changes are marked as COMMIT happens when the only diff is the git revision.

sukhil-suresh:
```
$kp image create test-image --tag gcr.io/cf-build-service-dev-219913/ssuresh/test-image --local-path /Users/ssuresh/workspace/samples/go/mod
Creating Image...
Uploading to 'gcr.io/cf-build-service-dev-219913/ssuresh/test-image-source'...
	Uploading 'gcr.io/cf-build-service-dev-219913/ssuresh/test-image-source@sha256:99bbe5fac447867858669d17471295387056e28fed2fe4bd9e8403ae8bd51759'
Image "test-image" created

$ kp build logs test-image
===> PREPARE
Build reason(s): CONFIG
CONFIG:
	resources: {}
	- source: {}
	+ source:
	+   registry:
	+     image: gcr.io/cf-build-service-dev-219913/ssuresh/test-image-source@sha256:99bbe5fac447867858669d17471295387056e28fed2fe4bd9e8403ae8bd51759
Loading secret for "gcr.io" from secret "gcr" at location "/var/build-secrets/gcr"
...

$ kp image patch test-image --git  https://github.com/paketo-buildpacks/samples --git-revision main --sub-path go/mod
Patching Image...
Image "test-image" patched

$ kp build logs test-image
===> PREPARE
Build reason(s): CONFIG
CONFIG:
	resources: {}
	source:
	-   registry:
	-     image: gcr.io/cf-build-service-dev-219913/ssuresh/test-image-source@sha256:99bbe5fac447867858669d17471295387056e28fed2fe4bd9e8403ae8bd51759
	+   git:
	+     revision: d6046f84d39aee034a69cc4dc1e5d33b06a651eb
	+     url: https://github.com/paketo-buildpacks/samples
	+   subPath: go/mod
Loading secret for "gcr.io" from secret "gcr" at location "/var/build-secrets/gcr"
...
```

PR #560, opened 2020-11-25
upgradeable lifecycle rfc

tomkennedy513:
[Readable](https://github.com/pivotal/kpack/blob/upgradeable-lifecycle/rfcs/0006-upgradeable-lifecycle.md)


aemengo:
I like the dichotomy drawn between the 2 approaches. This is something that I'd prefer user research guiding us.

We initially considered upgradeable lifecycles because we didn't want to bar users from newer versions of buildpacks and buildpack API. But _why_ are the majority of kpack users upgrading their buildpacks? If they aren't too concerned with upcoming buildpack features and are instead upgrading to protect themselves from CVEs, then it's more likely that they intend to upgrade throughout their cluster. Then the appeal of "lifecycle flexibility" becomes less pronounced.

Personally, I'm in favor of approach 2. Despite kpack have enough k8s abstractions as it is, I don't want a case where an operator unintentionally upgrades the builder that builds production images, as opposed to the one builds dev and test images.

tomkennedy513:
> Despite kpack have enough k8s abstractions as it is, I don't want a case where an operator unintentionally upgrades the builder that builds production images, as opposed to the one builds dev and test images.

That's a good point. Users may not want to make sweeping, cluster-wide changes as frequently if this is a possible issue that may arise. Giving them the ability to be focused with upgrades seems valuable to me.

sukhil-suresh:
I am also leaning toward approach 2, involving a defined Lifecycle Resource with kp support. just a couple of things I am wondering about
- if a cluster scoped lifecycle makes more sense? It seems (operator focused and) similar to ClusterStack, ClusterStore resources
- approach 2 can could still include descriptor support of the lifecycle resource?

tomkennedy513:
> * if a cluster scoped lifecycle makes more sense? It seems (operator focused and) similar to ClusterStack, ClusterStore resources

ya, I agree with that

> * approach 2 could still include descriptor _support_ for the lifecycle resource?

yes, definitely

rfcs/0000-upgradeable-lifecycle.md
```diff
@@ -0,0 +1,76 @@
...
+   	  - group:
+   	    - id: paketo-buildpacks/ruby
+   	  - group:
+   	    - id: paketo-buildpacks/nodejs
+   	  - group:
+   	    - id: paketo-buildpacks/java
+   ```
+
+**Complexity:**
+The first approach is less complex than the second, but could limit our options to change things down the road. The second option is definitely more complex, but follows the same pattern that we have for stores and stacks.
```
mgibson1121, line 62: I feel like most of our concern for this feature was around the case of newer buildpack versions not being compatible with older lifecycle versions.  But could the inverse be possible?  Incompatibility because the buildpacks in a clusterstore are outdated and the lifecycle is newer?  Perhaps this is a moot point because the longest I've ever heard of an active user of CF/kpack going without updating buildpacks is 3 months.

rfcs/0000-upgradeable-lifecycle.md
```diff
@@ -0,0 +1,76 @@
...
+1. A simpler, monolithic approach that would keep a single lifecycle version for the whole cluster like it is today, but with the functionality fo configuring that version exposed. The main benefit of soing something like that it will likely be the least intrusive on user workflow. The lifecycle would continue to be something that is more back of mind for users. There are multiple ways to handle this approach:
+	1. Create a command in `kp` (i.e `kp lifecycle upgrade`) that would give users the ability to manually update their lifecycle
+	2. Provide a script similiar to our `hack/lifecycle.go` that users can run 
+	3. Add the concept of lifecycle version to our descriptor files so that `kp import` can upgrade the lifecycle.  
+
+2. A more complex approach that would treat the lifecycle as a resource like a store or stack. Lifecycles would be tied to each builder so that different builders can have different lifecycle versions.This has the benefit of allowing users to test new lifecycle versions without having to spin up another cluster. Also, users who created their own buildpack would not have to worry about a newer lifecycle dropping support for the Buildpack Api they are using.
```
mgibson1121: In a user interview with a kpack users at Nasdaq he stated that the currently assign a cluster per application.  Just food for thought that it may not be burdensome to utilize many clusters.

mgibson1121: Good point about custom buildpack creation.  If we go with the option of creating a lifecycle resource, this would allow App Ops (primary buildpack author persona imo) to not need to ask a Platform Operator to upgrade the lifecycle.  However, if we support upgrading via the descriptor file and customers are `kp import`ing frequently, seems like this wouldn't be a problem.

mgibson1121:
Definitely in favor of approach 2.  @aemengo we don't have specific user research on this topic as up until this point it has been something rather in the background for many users of kpack who are not intimately familiar with CNBs.  But we can make some assumptions based on what we've observed so far.  Our assumption was that users of kpack within large organizations would configure a multitude of builders to suit the idiosyncrasies of each team.  So far we've seen that some early users (like the App Ops at Nationwide and Operators at Wells Fargo) just edit the default clusterbuilder.  Similarly, users of kpack within the Department of Defense are adding custom buildpacks along with VMware provided buildpacks to a custom clusterbuilder intended for broad use.  This is not to say that more complex behavior will not happen eventually, it just has not happened so far.

My thinking is that multi-lifecycle support is most valuable if app ops working with individual teams are dynamically managing builders according to their preferences and use cases.  If this isn't happening, then doing the simple thing seems like the best option.  Particularly if users are willing to `kp import` frequently (which we're seeing), seems like a lot of this complexity could just be swept under the rug (in a good way).

rfcs/0000-upgradeable-lifecycle.md
```diff
@@ -0,0 +1,75 @@
...
+
+**Actions to take:**
+We would create a ClusterLifecycle resource similiar a store or stack. The resources would be tied to each builder so that different builders can have different lifecycle versions. This has the benefit of allowing users to test new lifecycle versions without having to spin up another cluster. Also, users who created their own buildpack would not have to worry about a newer lifecycle dropping support for the Buildpack Api they are using. Due to the operator focused nature of this resource, and its similarity to ClusterStores and ClusterStacks, implementing this as a cluster scoped resource makes more sense at this time.
+
+On the kp side, we could create commands that would assist users in uploading or relocating the lifecycle to their registry as well as managing existing lifecycles like they do with ClusterStores or ClusterStacks:
+	
+`kp clusterlifecycle create my-lifecycle --image "https://github.com/buildpacks/lifecycle/releases/download/v0.9.2/lifecycle-v0.9.2+linux.x86-64.tgz" --tag "gcr.io/my-registry/lifecycle:v0.9.2"`
```
matthewmcnew, line 12: I like providing the ability to use the actual lifecycle release

rfcs/0000-upgradeable-lifecycle.md
```diff
@@ -0,0 +1,75 @@
+**Problem:**
+Currently, the lifecycle version used in kpack is not configurable, nor is it user upgradeable. Users must wait for a new kpack release to be able to upgrade the version of the lifecycle used by their builders.
+
+**Outcome:**
+We would like to decouple the lifecycle upgrade process from the kpack upgrade process, allowing users to upgrade the lifecycle on their own.
+
+**Actions to take:**
+We would create a ClusterLifecycle resource similiar a store or stack. The resources would be tied to each builder so that different builders can have different lifecycle versions. This has the benefit of allowing users to test new lifecycle versions without having to spin up another cluster. Also, users who created their own buildpack would not have to worry about a newer lifecycle dropping support for the Buildpack Api they are using. Due to the operator focused nature of this resource, and its similarity to ClusterStores and ClusterStacks, implementing this as a cluster scoped resource makes more sense at this time.
```
matthewmcnew, line 8: > Also, users who created their own buildpack would not have to worry about a newer lifecycle dropping support for the Buildpack Api they are using

I think realistically this should not be concern as the lifecycle should maintain backwards compatibility to old buildpack api versions.

tylerphelan:
I like the idea of storing 1 lifecycle version per `kpack` and making it configurable via `descriptor.yaml` and a `kp` command and then all builders use that lifecycle version.

- I think the lifecycle should be as invisible to users as possible until users request lifecycle features.
- I think builders should always use the latest lifecycle version that `kpack` supports until users request the ability to use builders with different lifecycle versions
- We can still use `descriptor.yaml` and a `kp` command to change the lifecycle without creating a new resource
- If users want to rollback lifecycle versions, they should still be able to do that

tomkennedy513:
After discussion in working group, we think that keeping the lifecycle version tied to kpack and not as a resource tied to a builder. One thing we need to figure out is how to upgrade the lifecycle without requiring a kpack controller restart. Will update the rfc to reflect this.

rfcs/0000-upgradeable-lifecycle.md
```diff
@@ -0,0 +1,82 @@
+**Problem:**
+Currently, the lifecycle version used in kpack is not configurable, nor is it user upgradeable. Users must wait for a new kpack release to be able to upgrade the version of the lifecycle used by their builders.
+
+**Outcome:**
+We would like to decouple the lifecycle upgrade process from the kpack upgrade process, allowing users to upgrade the lifecycle on their own.
+
+**Actions to take:**
```
matthewmcnew, line 7: I think it would be useful to discuss how the lifecycle is monolithically configured. Currently, it is provided as a an env var on the kpack controller so, updating that would require a kpack controller restart. Perhaps, a monitored configmap would make sense?

rfcs/0000-upgradeable-lifecycle.md
```diff
@@ -0,0 +1,82 @@
...
+A simple, monolithic approach that would keep a single lifecycle version for the whole cluster like it is today, but with the functionality fo configuring that version exposed. The main benefit of doing something like this it will likely be the least intrusive on user workflow. The lifecycle would continue to be something that is more back of mind for users. Additionally, since kpack has to be compatible with the platform api of the lifecycle, keeping the lifecycle version tied to kpack instead of a resource on a builder seems to make more sense right now. We would implement this by adding a few features to kp:
+
+1. Create a command in `kp` that would give users the ability to manually update their lifecycle:
+
+	`kp lifecycle upgrade --path "https://github.com/buildpacks/lifecycle/releases/download/v0.9.2/lifecycle-v0.9.2+linux.x86-64.tgz" --tag "gcr.io/my-registry/lifecycle:v0.9.2"`
```
tylerphelan, line 12: we could default use the `canonical-repository` similar to clusterstack/clusterbuilder creation and remove the need for `--tag` flag

rfcs/0000-upgradeable-lifecycle.md
```diff
@@ -0,0 +1,91 @@
...
+
+1. Create a command in `kp` that would give users the ability to manually update their lifecycle:
+
+	`kp lifecycle upgrade --path "https://github.com/buildpacks/lifecycle/releases/download/v0.9.2/lifecycle-v0.9.2+linux.x86-64.tgz"`  
+	or  
+	`kp lifecycle upgrade --image "gcr.io/my-registry/lifecycle:v0.9.2"`  
+
+2. Add the concept of lifecycle version to our descriptor files so that `kp import` can upgrade the lifecycle. 
+
+We would also need compatibility checking to make sure that kpack can support the version of the lifecycle that is being used. Additionally, we would not want all images to be rebuilt on a lifecycle upgrade.
```
tylerphelan, line 18: is it possible to accomplish this in a webhook validation?

rfcs/0000-upgradeable-lifecycle.md
```diff
@@ -0,0 +1,91 @@
...
+
+**Outcome:**
+We would like to decouple the lifecycle upgrade process from the kpack upgrade process, allowing users to upgrade the lifecycle on their own.
+
+**Actions to take:**
+A simple, monolithic approach that would keep a single lifecycle version for the whole cluster like it is today, but with the functionality fo configuring that version exposed. The main benefit of doing something like this it will likely be the least intrusive on user workflow. The lifecycle would continue to be something that is more back of mind for users. We would ship a version of the lifecycle with each kpack release like we do today. Additionally, since kpack has to be compatible with the platform api of the lifecycle, keeping the lifecycle version tied to kpack instead of a resource on a builder seems to make more sense right now. We would implement this by adding a few features to kp:
+
+1. Create a command in `kp` that would give users the ability to manually update their lifecycle:
+
+	`kp lifecycle upgrade --path "https://github.com/buildpacks/lifecycle/releases/download/v0.9.2/lifecycle-v0.9.2+linux.x86-64.tgz"`
```
tylerphelan, line 12: did we talk about how this might be hard b/c it's not in image form? maybe we just allow lifecycle images?

rfcs/0000-upgradeable-lifecycle.md
```diff
@@ -0,0 +1,91 @@
...
+We would like to decouple the lifecycle upgrade process from the kpack upgrade process, allowing users to upgrade the lifecycle on their own.
+
+**Actions to take:**
+A simple, monolithic approach that would keep a single lifecycle version for the whole cluster like it is today, but with the functionality fo configuring that version exposed. The main benefit of doing something like this it will likely be the least intrusive on user workflow. The lifecycle would continue to be something that is more back of mind for users. We would ship a version of the lifecycle with each kpack release like we do today. Additionally, since kpack has to be compatible with the platform api of the lifecycle, keeping the lifecycle version tied to kpack instead of a resource on a builder seems to make more sense right now. We would implement this by adding a few features to kp:
+
+1. Create a command in `kp` that would give users the ability to manually update their lifecycle:
+
+	`kp lifecycle upgrade --path "https://github.com/buildpacks/lifecycle/releases/download/v0.9.2/lifecycle-v0.9.2+linux.x86-64.tgz"`  
+	or  
+	`kp lifecycle upgrade --image "gcr.io/my-registry/lifecycle:v0.9.2"`
```
tylerphelan, line 14: do we need a kp command for the windows lifecycle?

rfcs/0000-upgradeable-lifecycle.md
```diff
@@ -0,0 +1,91 @@
...
+
+1. Create a command in `kp` that would give users the ability to manually update their lifecycle:
+
+	`kp lifecycle upgrade --path "https://github.com/buildpacks/lifecycle/releases/download/v0.9.2/lifecycle-v0.9.2+linux.x86-64.tgz"`  
+	or  
+	`kp lifecycle upgrade --image "gcr.io/my-registry/lifecycle:v0.9.2"`  
+
+2. Add the concept of lifecycle version to our descriptor files so that `kp import` can upgrade the lifecycle. 
+
+One issue with this approach is that currently, the lifecycle image is passed to the kpack controller as an environment variable, so any change to the lifecycle image would require a kpack controller restart to propogate that change. We could get around this by introducing a `kpack-config` config map that would be monitored by the kpack controller for changes. To start, this config will probably just have the lifecycle version in it, but it does open the door for us to add other config options further down the road. An example `kpack-config` could look like:
```
sukhil-suresh, line 18: should we just use the existing `kp-config` config map used for the canonical values? having a `kp-config` and a `kpack-config` seems confusing to me

rfcs/0000-upgradeable-lifecycle.md
matthewmcnew: Perhaps, we could either remove this command or require the path to the windows lifecycle & the linux lifecycle?

If the path is provided we would generate the image in the cli?

rfcs/0000-upgradeable-lifecycle.md
```diff
@@ -0,0 +1,91 @@
...
+
+```yaml
+
+apiVersion: v1
+data:
+  lifecycle.image: gcr.io/my-registry/lifecycle:v0.9.2
```
matthewmcnew, line 24: Currently, [kbld](https://get-kbld.io/) utilizes the image key to detect images. (That is why each image is currently in a standalone config map). Will `lifecycle.image` will not be identified as an image by kbld?

rfcs/0000-upgradeable-lifecycle.md
```diff
@@ -31,6 +29,8 @@ metadata:
 
 ```
 
+We would also need compatibility checking to make sure that kpack can support the version of the lifecycle that is being used. Additionally, we would not want all images to be rebuilt on a lifecycle upgrade. To accomplish the compatibility checking, we can create a validating webhook that would make sure that the lifecycle in the config map was supported by kpack.
```
matthewmcnew, line 32: Note, this would require the config map to include metadata on what platform/buildpack apis its lifecycle supports.

mgibson1121 (approved):

aemengo (approved):

rfcs/0000-upgradeable-lifecycle.md
tomkennedy513: I think we can use kbld's keyMatcher for this

tomkennedy513: something like 
```yaml
---
apiVersion: kbld.k14s.io/v1alpha1
kind: Config

searchRules:
- keyMatcher:
    name: lifecycle.image
```

rfcs/0000-upgradeable-lifecycle.md
```diff
@@ -0,0 +1,33 @@
...
+
+**Outcome:**
+We would like to decouple the lifecycle upgrade process from the kpack upgrade process, allowing users to upgrade the lifecycle on their own.
+
+**Actions to take:**
+A simple, monolithic approach that would keep a single lifecycle version for the whole cluster like it is today, but with the functionality fo configuring that version exposed. The main benefit of doing something like this it will likely be the least intrusive on user workflow. The lifecycle would continue to be something that is more back of mind for users. We would ship a version of the lifecycle with each kpack release like we do today. Additionally, since kpack has to be compatible with the platform api of the lifecycle, keeping the lifecycle version tied to kpack instead of a resource on a builder seems to make more sense right now. We would implement this by adding a few features to kp:
+
+1. Create a command in `kp` that would give users the ability to manually update their lifecycle:
+
+	`kp lifecycle upgrade --image "gcr.io/my-registry/lifecycle:v0.9.2"`
```
tylerphelan, line 12: we could be less prescriptive with the flag in the event that we decide to use the same flag for images and files/urls:

ie.
```
kp lifecycle upgrade --linux "gcr.io/my-registry/lifecycle:v0.9.2" --windows "gcr.io/my-registry/windows-lifecycle:v0.9.2"
```

sukhil-suresh (approved):

rfcs/0000-upgradeable-lifecycle.md
```diff
@@ -9,11 +9,11 @@ A simple, monolithic approach that would keep a single lifecycle version for the
 
 1. Create a command in `kp` that would give users the ability to manually update their lifecycle:
 
-	`kp lifecycle upgrade --image "gcr.io/my-registry/lifecycle:v0.9.2"`
+	`kp lifecycle upgrade --linux "gcr.io/my-registry/lifecycle:v0.9.2" --windows "gcr.io/my-registry/windows-lifecycle:v0.9.2"`
```
matthewmcnew, line 12: The active windows pr uses a single image for the lifecycle: https://github.com/pivotal/kpack/pull/546.

rfcs/0000-upgradeable-lifecycle.md
```diff
@@ -0,0 +1,33 @@
+**Problem:**
+Currently, the lifecycle version used in kpack is not configurable, nor is it user upgradeable. Users must wait for a new kpack release to be able to upgrade the version of the lifecycle used by their builders.
+
+**Outcome:**
+We would like to decouple the lifecycle upgrade process from the kpack upgrade process, allowing users to upgrade the lifecycle on their own.
+
+**Actions to take:**
+A simple, monolithic approach that would keep a single lifecycle version for the whole cluster like it is today, but with the functionality fo configuring that version exposed. The main benefit of doing something like this it will likely be the least intrusive on user workflow. The lifecycle would continue to be something that is more back of mind for users. We would ship a version of the lifecycle with each kpack release like we do today. Additionally, since kpack has to be compatible with the platform api of the lifecycle, keeping the lifecycle version tied to kpack instead of a resource on a builder seems to make more sense right now. We would implement this by adding a few features to kp:
```
matthewmcnew, line 8:
```suggestion
A simple, monolithic approach that would keep a single lifecycle version for the whole cluster like it is today, but with the functionality for configuring that version exposed. The main benefit of doing something like this it will likely be the least intrusive on user workflow. The lifecycle would continue to be something that is more back of mind for users. We would ship a version of the lifecycle with each kpack release like we do today. Additionally, since kpack has to be compatible with the platform api of the lifecycle, keeping the lifecycle version tied to kpack instead of a resource on a builder seems to make more sense right now. We would implement this by adding a few features to kp:
```

matthewmcnew (approved):

tylerphelan (approved):

PR #568, opened 2020-12-07
Handle delegation roles when signing images with notary

djoyahoy:


pkg/notary/authenticating_round_tripper.go
```diff
@@ -39,28 +41,34 @@ func (a *AuthenticatingRoundTripper) RoundTrip(req *http.Request) (*http.Respons
 
 	header := resp.Header.Get("www-authenticate")
 	parts := strings.SplitN(header, " ", 2)
```
tylerphelan, line 43: should `len(parts) == 2` be checked?

tylerphelan (approved):

pkg/notary/authenticating_round_tripper.go
djoyahoy: I'm not checking it here because I assume the response is well formed as `Bearer options...`. That seems like a fair assumption to me, but I totally hear you.

docs/image.md
```diff
@@ -106,6 +107,37 @@ build:
...
+    secretRef:
+      name: "notary-secret"
+```
+- `v1.url`: The URL of the notary server.
+- `v1.secretRef.name`: A [secret](#notary-secret) containing the encrypted private key and private key password.
+
+#### Generate Signing Key
+To generate a singing key, use the following commands from the [Docker Content Trust](https://docs.docker.com/engine/security/trust/#signing-images-with-docker-content-trust) documentation:
+```shell script
+% export DOCKER_CONTENT_TRUST_SERVER=<notary-server-url>
```
matthewmcnew, line 126: Is there a reason to recommend the docker cli here vs [notary](https://github.com/pivotal/kpack/blob/master/rfcs/0005-notary-integration.md#key-generation-and-storage)?

pkg/notary/authenticating_round_tripper.go
```diff
@@ -103,6 +111,15 @@ func (a *AuthenticatingRoundTripper) RoundTrip(req *http.Request) (*http.Respons
 	return a.WrappedRoundTripper.RoundTrip(req)
 }
 
+func extractBearerOption(kind string, from string) (string, error) {
+	r := regexp.MustCompile(fmt.Sprintf(`%s="(.*?)"`, kind))
```
matthewmcnew, line 115: Interesting evolution here as the result of multiple scopes.

matthewmcnew (approved):

docs/image.md
djoyahoy: Using the docker CLI is significantly easier. I also suspect thats what most users will use.

Issue #513, opened 2020-09-29
Add reason-message and reasons diff to build logs 


sampeinado:
[RFC](https://github.com/sampeinado/kpack/blob/master/rfcs/0002-reason-message-diff.md)
[Thread on PR](https://github.com/pivotal/kpack/pull/492)

Check out the mockups in the RFC !

mgibson1121:
@sukhil-suresh Not seeing the reason diff in the status output on staging using the latest kp binary I could find (`0.1.4.build.8)`.  As I type, it strikes me that this is probably the problem as that binary was published on Dec. 8.  I've been getting the binaries from here https://console.cloud.google.com/storage/browser/kp-public;tab=objects?project=cf-build-service-dev-219913&prefix=&forceOnObjectsSortingFiltering=false&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22)).  Is there a better place?

mgibson1121:
This story passes acceptance.

PR #590, opened 2021-01-05
Fix build trigger timestamp

sukhil-suresh:
Server side generation of build trigger timestamp used in the build change message

Depending on client side timestamp is problematic, since a `time.Now().String()` results in different formats based on the OS. Explicitly specifying a time format in the cli would fix things, but usage of an older cli would cause problems. More details in [this Slack thread](https://vmware.slack.com/archives/GJBSG61B8/p1608068169254200)

Issue: https://github.com/pivotal/kpack/issues/513

### DIFF
--- a/pkg/buildchange/change_processor_test.go
+++ b/pkg/buildchange/change_processor_test.go
@@ -43,21 +43,9 @@ func testChangeProcessor(t *testing.T, when spec.G, it spec.S) {
 
 	when("single change processed", func() {
 		when("TRIGGER", func() {
-			when("date string is given in unsupported format", func() {
-				change := buildchange.NewTriggerChange("bad date string")
-				expectedErrorStr := `error determining if build is required for reason 'TRIGGER': parsing time "bad date string" as "2006-01-02 15:04:05.000000 -0700 MST m=+0.000000000": cannot parse "bad date string" as "2006"`
-
-				it("errors for Summarize", func() {
-					_, err := cp.Process(change).Summarize()
-					assert.Error(t, err)
-					assert.Equal(t, expectedErrorStr, err.Error())
-				})
-			})
-
-			when("date string is given in supported format", func() {
-				change := buildchange.NewTriggerChange("2020-11-20 15:38:15.794105 -0500 EST m=+0.022963826")
-				expectedReasonsStr := "TRIGGER"
-				expectedChangesStr := testhelpers.CompactJSON(`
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	github.com/mgutz/ansi v0.0.0-20170206155736-9520e82c474b
 	github.com/pkg/errors v0.9.1
 	github.com/sclevine/spec v1.4.0
+	github.com/shurcooL/go v0.0.0-20180423040247-9e1955d9fb6e
```
matthewmcnew, line 23: What is this used for?

sukhil-suresh: not directly used

`go mod graph` gives this as the output
```
...
github.com/sourcegraph/go-diff@v0.5.3 github.com/gogo/protobuf@v1.1.1
github.com/sourcegraph/go-diff@v0.5.3 github.com/shurcooL/go@v0.0.0-20180423040247-9e1955d9fb6e
github.com/sourcegraph/go-diff@v0.5.3 github.com/shurcooL/go-goon@v0.0.0-20170922171312-37c2f522c041
...
```
and `go mod why github.com/shurcooL/go` gives
```
go: finding module for package github.com/shurcooL/go
# github.com/shurcooL/go
(main module does not need package github.com/shurcooL/go)
```
However `go mod tidy` does not take it out. And if I manually remove it from go.mod and go.sum and run tests, it comes right back and so, I left it in

sukhil-suresh: happy to take it out manually. should I?

matthewmcnew: It looks like it got accidentally pulled in here. https://github.com/pivotal/kpack/blob/693ee761d2df7e0a41ed15e64b349e15a892f82c/cmd/rebase/main.go#L15

matthewmcnew: I will create a PR right now to remove it.

matthewmcnew: https://github.com/pivotal/kpack/pull/593

matthewmcnew (approved):

go.mod
sukhil-suresh: rebased from master. resolved with #593

tylerphelan:
should we be more precise than seconds?

tomkennedy513 (approved):

sukhil-suresh:
> should we be more precise than seconds?

Since the current build status for an image build Started/Finished times limit precision to seconds, I am leaning towards doing the same, unless we have clear requirement for it. 

If you still feel strongly about this, I think we might have to get buy in from rest of the team on this and I would be happy to make the change. Merging for now.

Issue #594, opened 2021-01-08
The lifecycle image should be configurable


tylerphelan:
RFC: https://github.com/pivotal/kpack/pull/560 (soon to be merged)

**Problem:**
Users have no way of upgrading the lifecycle image without upgrading the kpack release

**Criteria:**
1. An update to `ConfigMap` `lifecycle-image.data.image` results in builders being recreated
2. The `ConfigMap` `lifecycle-image.annotations.platformApiVersions` should be validated that kpack supports all of the platform apis 

**Actions:**
- The ConfigMap `lifecycle-image` should no longer be mounted in the kpack-controller
- kpack should watch for changes to ConfigMap `lifecycle-image` in the `kpack` namespace
- Updates to the ConfigMap `lifecycle-image.data.image` value should result in builders being recreated
- `lifecycle-image.annotations.platformApiVersions` is a new optional field containing space-separated values in the form of X.X
- if `lifecycle-image.annotations.platformApiVersions` contains no versions supported by kpack, the ConfigMap should fail validation

Notes:
- This will NOT fan out to all images. Builds will only use the new lifecycle on the next build scheduled normally

matthewmcnew:
Perhaps `lifecycle-image.data.platformApiVersions` should be an annotation to prevent implying that is a functional configuration attribute?

mgibson1121:
Moving to accepted and will accept the user facing issue https://github.com/vmware-tanzu/kpack-cli/issues/142

PR #596, opened 2021-01-13
Lifecycle image is configurable

tylerphelan:
- Builders watch configmap lifecycle-image for the image reference
- Lifecycle image platform api version is validated via annotation
- New validatingwebhookconfiguration for the configmap

https://github.com/pivotal/kpack/issues/594

Looking for feedback on:
- Lifecycle image validation lives in `pkg/apis/v1alpha1` because I thought it was part of our api
- Added a new ValidatingWebhookConfiguration resource because we could scope it to the desired ConfigMap using `objectSelector`. 
  - Added a label to the `lifecycle-image` ConfigMap to enable the `objectSelector`
- Naming:
  - lifecycle-image ConfigMap Annotation: `lifecycle.kpack.io/platformApiVersions: "0.3 0.4"`
  - lifecycle-image ConfigMap Label: `lifecycle.kpack.io/lifecycleImage: "true"`
  - new ValidatingWebhookConfiguration Name: `configmapvalidation.webhook.kpack.io`

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -174,8 +174,8 @@ func main() {
 	clusterStoreController := clusterstore.NewController(options, clusterStoreInformer, remoteStoreReader)
 	clusterStackController := clusterstack.NewController(options, clusterStackInformer, remoteStackReader)
 
-	lifecycleProvider.RegisterCallback(builderResync)
-	lifecycleProvider.RegisterCallback(clusterBuilderResync)
+	lifecycleProvider.AddEventHandler(builderResync)
+	lifecycleProvider.AddEventHandler(clusterBuilderResync)
 
 	stopChan := make(chan struct{})
 	informerFactory.Start(stopChan)
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -33,6 +33,7 @@ import (
 	"github.com/pivotal/kpack/pkg/client/clientset/versioned"
 	"github.com/pivotal/kpack/pkg/client/informers/externalversions"
 	"github.com/pivotal/kpack/pkg/cnb"
+	"github.com/pivotal/kpack/pkg/config"
 	"github.com/pivotal/kpack/pkg/dockercreds/k8sdockercreds"
 	"github.com/pivotal/kpack/pkg/duckbuilder"
 	"github.com/pivotal/kpack/pkg/git"
@@ -125,10 +126,10 @@ func main() {
 
 	buildpodGenerator := &buildpod.Generator{
 		BuildPodConfig: v1alpha1.BuildPodImages{
-			BuildInitImage:        *buildInitImage,
-			CompletionImage:       *completionImage,
-			RebaseImage:           *rebaseImage,
-			BuildInitWindowsImage: *buildInitWindowsImage,
+			BuildInitImage:         *buildInitImage,
+			CompletionImage:        *completionImage,
+			RebaseImage:            *rebaseImage,
+			BuildInitWindowsImage:  *buildInitWindowsImage,
 			CompletionWindowsImage: *completionWindowsImage,
 		},
 		K8sClient:       k8sClient,
@@ -155,20 +156,26 @@ func main() {
 		Keychain:       kpackKeychain,
 	}
 
+	lifecycleProvider := config.NewLifecycleProvider(*lifecycleImage, &registry.Client{}, kpackKeychain)
+	configMapWatcher.Watch(config.LifecycleConfigName, lifecycleProvider.UpdateImage)
+
 	builderCreator := &cnb.RemoteBuilderCreator{
 		RegistryClient:         &registry.Client{},
-		LifecycleImage:         *lifecycleImage,
 		KpackVersion:           cmd.Identifer,
+		LifecycleProvider:      lifecycleProvider,
 		NewBuildpackRepository: newBuildpackRepository(kpackKeychain),
 	}
 
 	buildController := build.NewController(options, k8sClient, buildInformer, podInformer, metadataRetriever, buildpodGenerator)
 	imageController := image.NewController(options, k8sClient, imageInformer, buildInformer, duckBuilderInformer, sourceResolverInformer, pvcInformer)
 	sourceResolverController := sourceresolver.NewController(options, sourceResolverInformer, gitResolver, blobResolver, registryResolver)
-	builderController := builder.NewController(options, builderInformer, builderCreator, keychainFactory, clusterStoreInformer, clusterStackInformer)
-	clusterBuilderController := clusterBuilder.NewController(options, clusterBuilderInformer, builderCreator, keychainFactory, clusterStoreInformer, clusterStackInformer)
+	builderController, builderResync := builder.NewController(options, builderInformer, builderCreator, keychainFactory, clusterStoreInformer, clusterStackInformer)
+	clusterBuilderController, clusterBuilderResync := clusterBuilder.NewController(options, clusterBuilderInformer, builderCreator, keychainFactory, clusterStoreInformer, clusterStackInformer)
 	clusterStoreController := clusterstore.NewController(options, clusterStoreInformer, remoteStoreReader)
 	clusterStackController := clusterstack.NewController(options, clusterStackInformer, remoteStackReader)
+
+	lifecycleProvider.RegisterCallback(builderResync)
+	lifecycleProvider.RegisterCallback(clusterBuilderResync)
 
 	stopChan := make(chan struct{})
 	informerFactory.Start(stopChan)
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -79,7 +79,7 @@ func validatingAdmissionController(ctx context.Context, _ configmap.Watcher) *co
 		"validation.webhook.kpack.io",
 		// The path on which to serve the webhook.
 		"/validate",
-		// The resources to default.
+		// The resources to validate.
 		types,
 		// A function that infuses the context passed to Validate/SetDefaults with custom metadata.
 		func(ctx context.Context) context.Context {
--- a/pkg/apis/build/v1alpha1/build_pod.go
+++ b/pkg/apis/build/v1alpha1/build_pod.go
@@ -210,7 +210,7 @@ func (b *Build) BuildPod(images BuildPodImages, secrets []corev1.Secret, config 
 						),
 						ImagePullPolicy: corev1.PullIfNotPresent,
 					},
-					ifWindows(config.OS, addNetworkWaitLauncherVolume(), useNetworkWaitLauncher(dnsProbeHost))...)
+						ifWindows(config.OS, addNetworkWaitLauncherVolume(), useNetworkWaitLauncher(dnsProbeHost))...)
 				}
 			}),
 			SecurityContext: podSecurityContext(config),
@@ -494,7 +494,7 @@ func useNetworkWaitLauncher(dnsProbeHost string) stepModifier {
 		startCommand := container.Command
 		if len(startCommand) == 0 {
 			container.Args = args([]string{dnsProbeHost}, container.Args)
-		}else {
+		} else {
 			container.Args = args([]string{dnsProbeHost, "--"}, startCommand, container.Args)
 		}
 
--- a/pkg/cnb/create_builder.go
+++ b/pkg/cnb/create_builder.go
@@ -16,12 +16,16 @@ type BuildpackRepository interface {
 	FindByIdAndVersion(id, version string) (RemoteBuildpackInfo, error)
 }
 
+type LifecycleProvider interface {
+	GetImage() (v1.Image, error)
+}
+
 type NewBuildpackRepository func(clusterStore *v1alpha1.ClusterStore) BuildpackRepository
 
 type RemoteBuilderCreator struct {
 	RegistryClient         RegistryClient
-	LifecycleImage         string
 	NewBuildpackRepository NewBuildpackRepository
+	LifecycleProvider      LifecycleProvider
 	KpackVersion           string
 }
 
@@ -33,7 +37,7 @@ func (r *RemoteBuilderCreator) CreateBuilder(keychain authn.Keychain, clusterSto
 		return v1alpha1.BuilderRecord{}, err
 	}
 
-	lifecycleImage, _, err := r.RegistryClient.Fetch(keychain, r.LifecycleImage)
+	lifecycleImage, err := r.LifecycleProvider.GetImage()
 	if err != nil {
 		return v1alpha1.BuilderRecord{}, err
 	}
--- a/pkg/cnb/create_builder_test.go
+++ b/pkg/cnb/create_builder_test.go
@@ -42,7 +42,6 @@ func testCreateBuilderOs(os string, t *testing.T, when spec.G, it spec.S) {
 		stackID              = "io.buildpacks.stacks.some-stack"
 		mixin                = "some-mixin"
 		tag                  = "custom/example"
-		lifecycleImage       = "index.docker.io/kpack/lifecycle@sha256:d19308ce0c1a9ec083432b2c850d615398f0c6a51095d589d58890a721925584"
 		buildImage           = "index.docker.io/paketo-buildpacks/build@sha256:d19308ce0c1a9ec083432b2c850d615398f0c6a51095d589d58890a721925584"
 		runImage             = "index.docker.io/paketo-buildpacks/run@sha256:469f092c28ab64c6798d6f5e24feb4252ae5b36c2ed79cc667ded85ffb49d996"
 		buildImageTag        = "paketo-buildpacks/build:full-cnb"
@@ -168,12 +167,9 @@ func testCreateBuilderOs(os string, t *testing.T, when spec.G, it spec.S) {
 			},
 		}
 
-		subject = RemoteBuilderCreator{
-			RegistryClient:         registryClient,
-			LifecycleImage:         lifecycleImage,
-			KpackVersion:           "v1.2.3 (git sha: abcdefg123456)",
-			NewBuildpackRepository: newBuildpackRepo,
-		}
+		lifecycleProvider *fakeLifecycleProvider
+
+		subject RemoteBuilderCreator
 	)
 
 	buildpackRepository.AddBP("io.buildpack.1", "v1", []buildpackLayer{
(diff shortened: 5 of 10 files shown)


config/controller.yaml
```diff
@@ -22,6 +22,10 @@ kind: ConfigMap
 metadata:
   name:  lifecycle-image
   namespace: kpack
+  annotations:
+    lifecycle.kpack.io/platformApiVersions: "0.3 0.4"
```
tomkennedy513, line 26: is there a reason why these are space separated and not comma seperated

config/controller.yaml
```diff
@@ -22,6 +22,10 @@ kind: ConfigMap
...
+  labels:
+    lifecycle.kpack.io/lifecycleImage: "true"
```
matthewmcnew, line 28: What is the intent of this label?

tylerphelan: for the `ValidatingWebhookConfiguration.objectSelector` to scope the webhook to only the one ConfigMap instead of all ConfigMaps

cmd/controller/main.go
```diff
@@ -111,6 +110,9 @@ func main() {
 	pvcInformer := k8sInformerFactory.Core().V1().PersistentVolumeClaims()
 	podInformer := k8sInformerFactory.Core().V1().Pods()
 
+	namespacedK8sInformerFactory := informers.NewSharedInformerFactoryWithOptions(k8sClient, options.ResyncPeriod, informers.WithNamespace("kpack"))
```
matthewmcnew, line 113: This `main` already has a `configMapWatcher`. Could we use that instead?

config/controller.yaml
tylerphelan: No opinion about this. I guess it's whatever's clearer between:
`lifecycle.kpack.io/platformApiVersions: "0.3 0.4"`
`lifecycle.kpack.io/platformApiVersions: "0.3,0.4"`

matthewmcnew:
The lifecycle image is monolithic and not tied to the configuration of a single builder which allowed the old configuration to just inject the lifecycleImage as string to the `RemoteBuilderCreator`. It seems like we can preserve this behavior by injecting an atomic lifecycleImage provider that gets updated on its own from a config map watcher (perhaps knative's configmapwatcher). It seems like that would simplify the need to have reconcilers fetch the lifecycle image?

config/controller.yaml
matthewmcnew: Should we start using `app.kubernetes.io/part-of` https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ similar to tekton?

pkg/apis/build/v1alpha1/lifecycle_image_validation.go
```diff
@@ -0,0 +1,44 @@
...
+	LifecycleConfigNamespace                     = "kpack"
+	LifecycleConfigImageKey                      = "image"
+	LifecycleConfigPlatformApiVersionsAnnotation = "lifecycle.kpack.io/platformApiVersions"
+)
+
+type LifecycleImageValidator struct {
+	SupportedPlatformApiVersions []string
+}
+
+func (v LifecycleImageValidator) Validate(configMap *corev1.ConfigMap) (interface{}, error) {
```
matthewmcnew, line 21: This is elegantly implemented but, seeing it makes me wonder if the validation is worth setting up a fringe webhook? At this time we don't expect any lifecycles to not be supported in the forseeable future. What does everyone think?

tylerphelan: I kinda agree. Is this worth it?

tomkennedy513: I'm down to skip it for now

config/controller.yaml
aemengo: comma is clearer.

cmd/webhook/main.go
```diff
@@ -79,7 +79,7 @@ func validatingAdmissionController(ctx context.Context, _ configmap.Watcher) *co
 		"validation.webhook.kpack.io",
 		// The path on which to serve the webhook.
 		"/validate",
-		// The resources to default.
+		// The resources to validate.
```
matthewmcnew, line 82: Haha. nice.

pkg/cnb/create_builder.go
```diff
@@ -33,7 +37,7 @@ func (r *RemoteBuilderCreator) CreateBuilder(keychain authn.Keychain, clusterSto
 		return v1alpha1.BuilderRecord{}, err
 	}
 
-	lifecycleImage, _, err := r.RegistryClient.Fetch(keychain, r.LifecycleImage)
+	lifecycleImage, err := r.LifecycleProvider.GetImage()
```
matthewmcnew, line 40: Nice.

pkg/lifecycle/lifecycle_provider.go
```diff
@@ -0,0 +1,78 @@
...
+	Keychain       authn.Keychain
+	lifecycleImage atomic.Value
+	callbacks      []func()
+	err            error
+}
+
+func NewProvider(lifecycleImageRef string, client RegistryClient, keychain authn.Keychain) (*Provider, error) {
+	img, _, err := client.Fetch(keychain, lifecycleImageRef)
+	if err != nil {
+		return &Provider{}, errors.Wrap(err, "failed to get lifecycle image")
```
matthewmcnew, line 32: Does this error prevent the controller from starting?

pkg/lifecycle/lifecycle_provider.go
```diff
@@ -0,0 +1,78 @@
...
+		Keychain:       keychain,
+	}
+	p.lifecycleImage.Store(img)
+	return p, nil
+}
+
+func (l *Provider) UpdateImage(cm *corev1.ConfigMap) {
+	imageRef, ok := cm.Data[ConfigKey]
+	if !ok {
+		l.err = errors.New("lifecycle-image config invalid")
```
matthewmcnew, line 45: Should the err use sync/atomic as well?

sukhil-suresh (commented):
@tylerphelan did we chose not to use `ConfigMap.lifecycle-image.platformApiVersions` for validation?

pkg/apis/build/v1alpha1/lifecycle_image_validation.go
sukhil-suresh: considering that the current lifecycle image only supports 0.3+ for platform API, I would assume validation against the supported platform API does make sense. deprecation is bound to happen eventually (based on the precedence)

matthewmcnew: The lifecycle started supporting multiple platform APIs at the 0.3 platform api and the team doesn't have plans to remove APIS in the near future.

matthewmcnew: We can still validate the platform api @ builder creation.

cmd/controller/main.go
```diff
@@ -155,21 +156,31 @@ func main() {
...
 	sourceResolverController := sourceresolver.NewController(options, sourceResolverInformer, gitResolver, blobResolver, registryResolver)
-	builderController := builder.NewController(options, builderInformer, builderCreator, keychainFactory, clusterStoreInformer, clusterStackInformer)
-	clusterBuilderController := clusterBuilder.NewController(options, clusterBuilderInformer, builderCreator, keychainFactory, clusterStoreInformer, clusterStackInformer)
+	builderController, builderResync := builder.NewController(options, builderInformer, builderCreator, keychainFactory, clusterStoreInformer, clusterStackInformer)
+	clusterBuilderController, clusterBuilderResync := clusterBuilder.NewController(options, clusterBuilderInformer, builderCreator, keychainFactory, clusterStoreInformer, clusterStackInformer)
 	clusterStoreController := clusterstore.NewController(options, clusterStoreInformer, remoteStoreReader)
 	clusterStackController := clusterstack.NewController(options, clusterStackInformer, remoteStackReader)
 
+	lifecycleProvider.RegisterCallback(builderResync)
+	lifecycleProvider.RegisterCallback(clusterBuilderResync)
```
aemengo, line 182: There's something a little unnerving seeing callbacks like this. I'm wondering if listening for messages on a channel is idiomatic way to do this in Golang..

pkg/lifecycle/lifecycle_provider.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	p := &Provider{
+		RegistryClient: client,
+		Keychain:       keychain,
+	}
+	p.lifecycleImage.Store(img)
+	return p, nil
+}
+
+func (l *Provider) UpdateImage(cm *corev1.ConfigMap) {
```
aemengo, line 42: I'm trying to understand:

There's a **default lifecycle** that comes through flag/environment variables.
Now a user can specify **custom lifecycle** through configMap.
What happens when the controller is restarted? Which **lifecycle** wins?

pkg/lifecycle/lifecycle_provider.go
tylerphelan: I thought it would be a good idea, but now I think it should not fail to start

pkg/lifecycle/lifecycle_provider.go
tylerphelan: If the controller is restarted whatever the configmap value is would be mounted to the container.

I agree it's a little strange. For context, the only reason I added the seeding of the lifecycle at startup was because the kpack controller errors because the lifecycle provider hasn't updated before the builders reconcile and it's a potentially alarming error message ("lifecycle has not been loaded"). It eventually resolves it's just the log line.

sukhil-suresh: I noticed this option
```
configMapWatcher.WatchWithDefault(cm corev1.ConfigMap, o ...Observer)
```
vs what we have now
```
configMapWatcher.Watch(name string, o ...Observer)
```
 and was wondering if `WatchWithDefault` could be used for the lifecycle image passed through the flag/env   variables... pondering if it would be helpful

cmd/controller/main.go
matthewmcnew: Perhaps, `callback` is not the correct term. 

Registering event handlers is pretty ubiquitous in the k8s/go ecosystem ([for example](https://github.com/pivotal/kpack/blob/46fe4fc3a99696d451a9e2fef2902451dd120108/pkg/reconciler/build/build.go#L52-L57)).  Channel listening seems like it would introduce unnecessary concurrency.

pkg/lifecycle/lifecycle_provider.go
matthewmcnew: It is important to note that the "default lifecycle" will be [mounted from the contents of the config map](https://github.com/pivotal/kpack/blob/46fe4fc3a99696d451a9e2fef2902451dd120108/config/controller.yaml#L110-L114).

matthewmcnew: It might make sense for the provider to not execute callbacks when there is no. change to the image.

pkg/apis/build/v1alpha1/lifecycle_image_validation.go
sukhil-suresh: yes, validating at builder creation is possible, but takes the error further away from the cause (I think). In general, giving error feedback closer to the action causing might be more helpful - not a deal breaker.

pkg/lifecycle/lifecycle_provider.go
tylerphelan: @sukhil-suresh 
`configMapWatcher.WatchWithDefault` provides a nice initial default resource for the watcher, but it processes the same way as `configMapWatcher.Watch` meaning that we would still not guarantee the `configMapWatcher` runs an update before the `clusterbuilder` controller and we still get the initial "lifecycle is not loaded" error. ultimately `configMapWatcher.WatchWithDefault` works the same as `configMapWatcher.Watch` if the expected configmap is already there.

cmd/controller/main.go
aemengo: In particular, it's the fact that we're passing the following around for another class to execute:

```go
	return impl, func() {
		impl.GlobalResync(builderInformer.Informer())
	}
```

Perhaps we've discussed this before, but it doesn't feel very open/closed. Just wondering if there's a different way to express this.

pkg/config/lifecycle_provider.go
```diff
@@ -0,0 +1,90 @@
...
+	defer l.lifecycleData.Store(data)
+
+	imageRef, ok := cm.Data[LifecycleConfigKey]
+	if !ok {
+		data.err = errors.New("lifecycle-image config invalid")
+		return
+	}
+
+	data.image, data.err = l.fetchImage(imageRef)
+}
```
sukhil-suresh, line 59: when the image is invalid, are we not updating the value and doing callbacks? do we need to do it?

sukhil-suresh: I see from the test that this is intentional behaviour, but haven't figured out why we want to do a callback...

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -210,7 +210,7 @@ func (b *Build) BuildPod(images BuildPodImages, secrets []corev1.Secret, config
 						),
 						ImagePullPolicy: corev1.PullIfNotPresent,
 					},
-					ifWindows(config.OS, addNetworkWaitLauncherVolume(), useNetworkWaitLauncher(dnsProbeHost))...)
+						ifWindows(config.OS, addNetworkWaitLauncherVolume(), useNetworkWaitLauncher(dnsProbeHost))...)
```
aemengo, line 213: did `go fmt` add this?

pkg/cnb/create_builder_test.go
```diff
@@ -168,12 +167,9 @@ func testCreateBuilderOs(os string, t *testing.T, when spec.G, it spec.S) {
...
 
-		subject = RemoteBuilderCreator{
-			RegistryClient:         registryClient,
-			LifecycleImage:         lifecycleImage,
-			KpackVersion:           "v1.2.3 (git sha: abcdefg123456)",
-			NewBuildpackRepository: newBuildpackRepo,
-		}
+		lifecycleProvider *fakeLifecycleProvider
+
+		subject RemoteBuilderCreator
```
matthewmcnew, line 172: Can we still initialize the remote builder creator here?

pkg/config/lifecycle_provider.go
```diff
@@ -0,0 +1,124 @@
...
+	newImg, err := l.fetchImage(imageRef)
+	if err != nil {
+		data.err = err
+		return
+	}
+
+	// Don't care if old image errored
+	oldImg, _ := l.GetImage()
+
+	isNewImg, err = isNewImage(oldImg, newImg)
```
matthewmcnew, line 68: It seems that avoiding calling callbacks if the digest has not changed is actually introducing a good amount of complexity here for a rather inconsequential optimization. Could we essentially accomplish the same thing by saving the configured image in the lifecycle data and exiting early if it is the same?

pkg/config/lifecycle_provider.go
```diff
@@ -0,0 +1,124 @@
...
+	defer l.callCallBacks(&isNewImg)
+	defer l.lifecycleData.Store(data)
+
+	imageRef, ok := cm.Data[LifecycleConfigKey]
+	if !ok {
+		data.err = errors.Errorf("%s config invalid", LifecycleConfigName)
+		return
+	}
+
+	newImg, err := l.fetchImage(imageRef)
```
matthewmcnew, line 59: A good amount of the complexity in this func is due it not returning an err. Could we avoid that complexity by calling a method that does return an error and only error handle here?
```
func (l *LifecycleProvider) UpdateImage(cm *corev1.ConfigMap) {
    defer l.callCallBacks(&isNewImg)
    data := &lifecycleData{}
    data.img, data.err = readLifecycleImage(cm)
    l.lifecycleData.Store(data)
}
```

matthewmcnew: Perhaps we could take a page from reconcilers:
```
func (l *LifecycleProvider) UpdateImage(cm *corev1.ConfigMap) {
    data := &lifecycleData{}
    data.img, data.err = readLifecycleImage(cm)
    if !equality.DeepEqual(l.lifecycleData.get(), data) {
        l.lifecycleData.Store(data)
        l.callCallbacks()

   }
}
```

pkg/config/lifecycle_provider.go
```diff
@@ -0,0 +1,124 @@
...
+
+	data := &lifecycleData{}
+
+	data.image, data.err = p.fetchImage(lifecycleImageRef)
+
+	p.lifecycleData.Store(data)
+	return p
+}
+
+func (l *LifecycleProvider) UpdateImage(cm *corev1.ConfigMap) {
```
aemengo, line 47: Here's the best that I could come up with. Hopefully there's a good idea in there somewhere:

```suggestion
func (l *LifecycleProvider) UpdateImage(cm *corev1.ConfigMap) {
	data, shouldCallCallBacks := l.updateImage(cm)
	if shouldCallCallBacks {
		l.callCallBacks()
	}

	l.lifecycleData.Store(data)
}

func (l *LifecycleProvider) updateImage(cm *corev1.ConfigMap) (*lifecycleData, bool) {
	data := &lifecycleData{}

	imageRef, ok := cm.Data[LifecycleConfigKey]
	if !ok {
		data.err = errors.Errorf("%s config invalid", LifecycleConfigName)
		return data, true
	}

	l.fetchImage(imageRef, data)
	if data.err != nil {
		return data, true
	}

	// Don't care if old image errored
	oldImg, _ := l.GetImage()

	var isNewImg bool
	isNewImg, data.err = isNewImage(oldImg, data.image)
	return data, isNewImg
}

func (l *LifecycleProvider) fetchImage(imageRef string, v *lifecycleData) {
	img, _, err := l.RegistryClient.Fetch(l.Keychain, imageRef)
	if err != nil {
		v.err = errors.Wrap(err, "failed to fetch lifecycle image")
		return
	}

	v.image = img
}
```

pkg/apis/build/v1alpha1/build_pod.go
tylerphelan: yea

pkg/config/lifecycle_provider.go
tylerphelan: I like this one!

pkg/cnb/create_builder_test.go
tylerphelan: updated

aemengo (approved):

matthewmcnew (approved):

sukhil-suresh (approved):

PR #609, opened 2021-01-26
add intelligent windows tolerations based on node taints

tomkennedy513:
- we only do this if the taints across all windows nodes are homogenous

### DIFF
--- a/config/controllerrole.yaml
+++ b/config/controllerrole.yaml
@@ -62,6 +62,12 @@ rules:
   - update
   - delete
   - watch
+- apiGroups:
+    - ""
+  resources:
+    - nodes
+  verbs:
+    - list
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
--- a/pkg/apis/build/v1alpha1/build_pod.go
+++ b/pkg/apis/build/v1alpha1/build_pod.go
@@ -73,6 +73,7 @@ type BuildPodBuilderConfig struct {
 	Gid         int64
 	PlatformAPI string
 	OS          string
+	NodeTaints  []corev1.Taint
 }
 
 var (
@@ -420,6 +421,7 @@ func (b *Build) BuildPod(images BuildPodImages, secrets []corev1.Secret, config 
 			NodeSelector: map[string]string{
 				"kubernetes.io/os": config.OS,
 			},
+			Tolerations: config.tolerations(),
 			Volumes: append(append(
 				secretVolumes,
 				corev1.Volume{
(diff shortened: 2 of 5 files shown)


config/controllerrole.yaml
```diff
@@ -62,6 +62,13 @@ rules:
   - update
   - delete
   - watch
+- apiGroups:
+    - ""
+  resources:
+    - nodes
+  verbs:
+    - get
```
matthewmcnew, line 70: Do we need `get` or is `list` suffice?

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -776,6 +778,25 @@ func (bc *BuildPodBuilderConfig) unsupported() bool {
 	return bc.PlatformAPI != "0.2" && bc.PlatformAPI != "0.3"
 }
 
+func (bc *BuildPodBuilderConfig) tolerations() []corev1.Toleration {
+	tolerations := make([]corev1.Toleration, 0)
+
+	if bc.OS != "windows" {
```
matthewmcnew, line 784: If taints are passed in do we care if it is windows/linux?

pkg/buildpod/generator.go
```diff
@@ -167,3 +176,49 @@ func parseCNBID(image ggcrv1.Image, env string) (int64, error) {
...
+
+	return taints, nil
+}
+
+func taintsEqual(taint1, taint2 []v1.Taint) bool {
+	if len(taint1) != len(taint2) {
+		return false
+	}
+
+	for i := 0; i < len(taint1); i++ {
```
matthewmcnew, line 215:
```suggestion
	for i  := range taint2 {
```

pkg/buildpod/generator_test.go
```diff
@@ -220,6 +277,155 @@ func testGenerator(t *testing.T, when spec.G, it spec.S) {
...
+								Value:     "different-value",
+								Effect:    corev1.TaintEffectNoSchedule,
+								TimeAdded: nil,
+							},
+						},
+					},
+				}
+
+				fakeK8sClientWithDifferentNode := fake.NewSimpleClientset(serviceAccount, dockerSecret, gitSecret, ignoredSecret, linuxNode, windowsNode1, windowsNode2, windowsNode3)
+				generator.K8sClient = fakeK8sClientWithDifferentNode
```
matthewmcnew, line 404: You could also `fakeK8sClient.CoreV1().Nodes().Create()` to add to the existing client.

pkg/apis/build/v1alpha1/build_pod.go
tomkennedy513: no probably not

pkg/buildpod/generator_test.go
tomkennedy513: oh sweet thanks

config/controllerrole.yaml
tomkennedy513: we don't get individual nodes, we just list all of them and iterate so I think we can ditch the get

tylerphelan:
should we do a provider pattern?

pkg/buildpod/generator.go
```diff
@@ -167,3 +176,49 @@ func parseCNBID(image ggcrv1.Image, env string) (int64, error) {
...
+	taints := nodeList[0].Spec.Taints
+	sort.Slice(taints, func(i, j int) bool {
+		return taints[i].Key < taints[j].Key
+	})
+
+	for _, node := range nodeList[1:] {
+		taintsToCompare := node.Spec.Taints
+		sort.Slice(taintsToCompare, func(i, j int) bool {
+			return taintsToCompare[i].Key < taintsToCompare[j].Key
+		})
```
aemengo, line 200: As an optimization, maybe we can move this second sort to be inside the `func taintsEqua(`, after the size check.

aemengo (approved):

aemengo (approved):

matthewmcnew (approved):

tylerphelan (approved):

pkg/buildpod/generator.go
tomkennedy513: i had that originally, but i changed it because even though its slightly less performant, i felt like it was a little weird to be sorting in the equals function, especially since i was only sorting one slice

Issue #608, opened 2021-01-26
handle duplicate secret names in image/build service accounts


tylerphelan:
This is possible on certain k8s deployments and results in multiple PVCs of the same name which errors

PR #610, opened 2021-01-26
Dedupe build secrets if service account has repeats

tylerphelan:
fixes https://github.com/pivotal/kpack/issues/608


tomkennedy513 (approved):

pkg/buildpod/generator.go
```diff
@@ -95,6 +95,7 @@ func (g *Generator) fetchServiceAccounts(build BuildPodable) ([]corev1.ServiceAc
 
 func (g *Generator) fetchBuildSecrets(build BuildPodable) ([]corev1.Secret, error) {
 	var secrets []corev1.Secret
+	var secretSet = map[string]bool{}
```
matthewmcnew, line 98: Use map[string]struct{} if we don't care about the values in a map.

matthewmcnew (approved):

aemengo (approved):

sukhil-suresh (approved):

Issue #607, opened 2021-01-26
Update kpack lifecycle to 0.10.2.


matthewmcnew:
[Lifecycle 0.10.2](https://github.com/buildpacks/lifecycle/releases/tag/v0.10.2)

PR #611, opened 2021-01-27
Update lifecycle to 0.10.2 & provide support for platform apis 0.5,0.4,0.3

matthewmcnew:
- Drop support for platform api 0.2
- If using 0.4+ default the process type to web (identical to pack behavior)
- Prevent notary/windows on platform api 0.3

Closes: https://github.com/pivotal/kpack/issues/607

### DIFF
--- a/hack/lifecycle/main.go
+++ b/hack/lifecycle/main.go
@@ -42,11 +42,11 @@ func main() {
 	flag.Parse()
 
 	image, err := lifecycleImage(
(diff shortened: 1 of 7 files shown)


matthewmcnew:
Perhaps we should validate that there is a supported platform api on builder creation. cc' @tylerphelan

matthewmcnew:
When this [rfc](https://github.com/buildpacks/rfcs/blob/main/text/0064-buildpacks-contribute-default-process-type.md) is implemented (likely in platform api 0.6) we can remove the default process type to web flag.

tomkennedy513 (approved):

tylerphelan (approved):

PR #620, opened 2021-02-18
Rfc for using git2go library to support Azure devops

tylerphelan:
[Readable](https://github.com/pivotal/kpack/blob/use-git2go/rfcs/0007-use-git2go.md)

### DIFF
--- a/rfcs/0007-use-git2go.md
+++ b/rfcs/0007-use-git2go.md
@@ -15,37 +15,65 @@ Azure Devops represents a significant use case
 
 [git2go](https://github.com/libgit2/git2go) is a library for go bindings to the git method library in c, [libgit2](https://libgit2.org/).
 
-## Proposal
+## Options
 
-### Use git2go for Azure Devops Repositories and go-git otherwise
+### 1. Use git2go
 
-The main reasoning for this is that git2go is the only fully-featured go library that supports the v2 git protocol, but it doesn't support shallow cloning which makes a big difference for builds.
+The main reasoning for this is that git2go is the only fully-featured go library that supports the v2 git protocol which is required for Azure Devops git.
 
-## Actions
+#### Actions
 
-### Add required dependencies for libgit2 to the build-init, windows-build-init, and kpack-controller image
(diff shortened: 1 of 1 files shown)


tylerphelan:
I put this in an rfc to get the ideas rolling and as this might have significant implications

rfcs/0007-use-git2go.md
```diff
@@ -0,0 +1,69 @@
...
+## Outcome
+Users can use Azure Devops for git repositories
+
+## Background
+
+Azure Devops represents a significant use case
+
+[go-git](https://github.com/go-git/go-git) is the git implementation library in pure go
+[git2go](https://github.com/libgit2/git2go) is a library for go bindings to the git method library in c, [libgit2](https://libgit2.org/).
+[Shallow cloning](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---depthltdepthgt) basically means we don't pull in the full history which is a performance benefit.
```
matthewmcnew, line 16: It is worth noting that we don't actually use shallow cloning at the [moment](https://github.com/pivotal/kpack/blob/12689fbfdf9d44a0d24d7acc435efcff8392f8c6/pkg/git/fetch.go#L51) (
depth: 0 is a full clone). My understanding is we do this to enable us to checkout commits that are not the tip of a branch.

matthewmcnew: We could probably do a shallow clone if possible and I believe we discussed investigating that for performance issues but, haven't it seems because we haven't had that many performance concerns 🤷‍♂️.

rfcs/0007-use-git2go.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Use git2go for Azure Devops Repositories and go-git otherwise
+
+The main reasoning for this is that git2go is the only fully-featured go library that supports the v2 git protocol, but it doesn't support shallow cloning which makes a big difference for builds.
+
+## Actions
+
+### Add required dependencies for libgit2 to the build-init image
+
+Build-init should be the only container that required this as it is the only container using git. This will require changing the build process for build-init.
```
matthewmcnew, line 28: I think we need it in the kpack controller as well and windows?

PR #631, opened 2021-03-02
Do not use direct execute for buildpod containers

tylerphelan:
- kpack images use the newest lifecycle platform version which will set a default process type


tomkennedy513 (approved):

matthewmcnew (approved):

matthewmcnew:
Finally! This has been a thorn in our side for a while. Thanks @tylerphelan !

PR #632, opened 2021-03-02
Update resources

tylerphelan:


config/controller.yaml
```diff
@@ -114,8 +114,8 @@ spec:
               key: image
         resources:
           requests:
-            cpu: 10m
-            memory: 50Mi
+            cpu: 20m
```
matthewmcnew, line 117: Is `20m` consistent with your observations? I didn't pay much attention to cpu.

matthewmcnew (approved):

config/controller.yaml
tylerphelan: yup. only when the controller is doing things though.

PR #635, opened 2021-03-04
Configure the pod security context to match the builder config

tylerphelan:
- This explicitly sets the user:group to run as for each container as well as the volume groups
- This is useful for when MustRunAs is set in a pod security policy


matthewmcnew (approved):

tomkennedy513 (commented):
we should delete the `removeSecurityContext` function now since windows builds don't need to remove it from the prepare container anymore

tomkennedy513 (approved):

Issue #344, opened 2020-04-08
Image name length requirement


thephw:
# Image name length requirement

```
Reconcile error	{"error": "image name requirement: invalid label value: \"github-onspaceship-booster-rack-app-690156b7cf99c7093a1ec4fd77ae3327ff8921b8\": at key: \"image.build.pivotal.io/image\": must be no more than 63 characters"}
knative.dev/pkg/controller.(*Impl).handleErr
```

At this time we utilize the image name as a label on an image's builds so the image name must adhere to an image's name to a labels length constraint of 63 characters. 

# Context

We were working around an issue where we programmatically made an Image resource from a github org/repo name and a SHA. The programmatic creation of the image still succeeds, but the build does not due to the image name length requirement on the build tag. The error popped up within the logs of kpack controller. 

Not a hard workaround to check the name length and adjust before creating the image. However we could add some validation to the name length to throw an error at creation time of the image knowing how the downstream implementation works.

dumez-k:
Fixed in [pr 640 ](https://github.com/pivotal/kpack/pull/640)

mgibson1121:
Nice!  Accepting.

PR #640, opened 2021-03-09
Validate length of image name

dumez-k:
resolves issue: https://github.com/pivotal/kpack/issues/344

### DIFF
--- a/pkg/apis/build/v1alpha1/image_validation.go
+++ b/pkg/apis/build/v1alpha1/image_validation.go
@@ -50,10 +50,22 @@ func (i *Image) SetDefaults(ctx context.Context) {
 }
 
 func (i *Image) Validate(ctx context.Context) *apis.FieldError {
-	return i.Spec.Validate(ctx).ViaField("spec")
+	return i.Spec.ValidateSpec(ctx).ViaField("spec").
+		Also(i.ValidateMetadata(ctx).ViaField("metadata"))
 }
 
-func (is *ImageSpec) Validate(ctx context.Context) *apis.FieldError {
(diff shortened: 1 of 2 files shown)


tomkennedy513 (approved):

matthewmcnew:
I notice that knative projects use `NameIsDNS1035Label from k8s.io/apimachinery aka https://github.com/knative/serving/blob/f7ecef165c02a745eb9c1cec04714e393b86ed3e/vendor/knative.dev/pkg/apis/metadata_validation.go#L55. Would it make sense to use that as well?

pkg/apis/build/v1alpha1/image_validation.go
```diff
@@ -50,10 +50,22 @@ func (i *Image) SetDefaults(ctx context.Context) {
 }
 
 func (i *Image) Validate(ctx context.Context) *apis.FieldError {
-	return i.Spec.Validate(ctx).ViaField("spec")
+	return i.Spec.ValidateSpec(ctx).ViaField("spec").
+		Also(i.ValidateMetadata(ctx).ViaField("metadata"))
```
matthewmcnew, line 54: Nice!

matthewmcnew (approved):

dumez-k:
> I notice that knative projects use `NameIsDNS1035Label from k8s.io/apimachinery aka https://github.com/knative/serving/blob/f7ecef165c02a745eb9c1cec04714e393b86ed3e/vendor/knative.dev/pkg/apis/metadata_validation.go#L55. Would it make sense to use that as well?

updated to use this validation approach

Issue #643, opened 2021-03-11
Error verifying write access to "kube-registry.infra.svc:5000/demo/spring-petclinic:1.0"


cmoulliard:
## Issue

When the step `prepare` is [executed](https://gist.github.com/cmoulliard/bc53910804cf951b36bea3a7c5804ab2#file-gistfile1-txt-L74-L96), then we got an auth error 

```
 logs -image tutorial-image -namespace demo
===> SETUP-CA-CERTS
Create certificate...
Populate certificate...
Update CA certificates...
Copying CA certificates...
Finished setting up CA certificates
===> PREPARE
Build reason(s): CONFIG
CONFIG:
	resources: {}
	- source: {}
	+ source:
	+   git:
	+     revision: master
	+     url: https://github.com/spring-projects/spring-petclinic
Loading secret for "https://kube-registry.infra.svc:5000/" from secret "tutorial-registry-credentials" at location "/var/build-secrets/tutorial-registry-credentials"
Error verifying write access to "kube-registry.infra.svc:5000/tutorial": POST https://kube-registry.infra.svc:5000/v2/tutorial/blobs/uploads/: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:tutorial Type:repository] map[Action:push Class: Name:tutorial Type:repository]]
```

The error happens here within the code of the kpack controller: https://github.com/pivotal/kpack/blob/6fb317827d243bd6d4f7a403e8e9eb90a95cde1c/pkg/dockercreds/access_checker.go#L44-L43

## Additional information

The `dockerconfig` file mounted from the secret - tutorial-registry-credentials and passed as arg to the prepare container contains the correct credential
```
kc get secret/tutorial-registry-credentials -n demo -o yaml
apiVersion: v1
data:
  .dockerconfigjson: eyJhdXRocyI6eyJodHRwczovL2t1YmUtcmVnaXN0cnkuaW5mcmEuc3ZjOjUwMDAvIjp7InVzZXJuYW1lIjoiYWRtaW4iLCJwYXNzd29yZCI6InNub3dkcm9wIiwiYXV0aCI6IllXUnRhVzQ2YzI1dmQyUnliM0E9In19fQ==
kind: Secret
metadata:
  name: tutorial-registry-credentials
  namespace: demo
type: kubernetes.io/dockerconfigjson
```

if you decode the base64 string

[...]

PR #644, opened 2021-03-11
Match registries with trailing slash

matthewmcnew:
- Trailing slashes are commonly provided by users such as `https://myreg.io/` & `myreg.io/`
-  Would fix the issue seen on this issue https://github.com/pivotal/kpack/issues/643
- Is there a risk we are being to permissive with this and matching on things docker itself would not match on?

### DIFF
--- a/pkg/dockercreds/match.go
+++ b/pkg/dockercreds/match.go
@@ -10,9 +10,18 @@ import (
 var registryDomains = []string{
 	// Allow naked domains
 	"%s",
+
+	// Allow naked domains with trailing slash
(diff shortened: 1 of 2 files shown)


tomkennedy513 (approved):

tylerphelan (approved):

attack:
The risk of being permissive it diverging from expected or spec'd behaviour, and then having to maintain this extra logic. This is more of a pain point if we leverage existing libraries to provide this logic.

matthewmcnew:
Closing for now.

PR #658, opened 2021-04-07
Don't write build cache to emptydir

phil9909:
Previously, when no cache-volume was specified the caching-data got written to an EmptyDir.
EmptyDirs are thrown away once the Pod is finished.
For some buildpacks/apps (e.g. gradle when building spring-music) generating the cache takes a considerable amount of time. If we know that we will throw it away anyway we can skip that.

Therefore we:
- Don't pass the `--cache-dir` arg to the lifecycle steps
- Don't add the cache-dir to the `VolumeMounts`
- Don't create an EmptyDir for the cache

Tagging my pairs: @bodymindarts @Haegi

### DIFF
--- a/pkg/apis/build/v1alpha1/build_pod.go
+++ b/pkg/apis/build/v1alpha1/build_pod.go
@@ -163,6 +163,16 @@ func (b *Build) BuildPod(images BuildPodImages, secrets []corev1.Secret, taints 
 		SubPath:   b.Spec.Source.SubPath, // empty string is a nop
 	}
 
+	var cacheArgs []string
+	var cacheVolumes []corev1.VolumeMount
+	if b.Spec.CacheName == "" {
+		cacheArgs = nil
+		cacheVolumes = nil
+	} else {
+		cacheArgs = []string{"-cache-dir=/cache"}
+		cacheVolumes = []corev1.VolumeMount{cacheVolume}
+	}
+
 	return &corev1.Pod{
 		ObjectMeta: v1.ObjectMeta{
 			Name:      b.PodName(),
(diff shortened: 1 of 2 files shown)


tylerphelan (approved):

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -633,16 +635,25 @@ func (b *Build) rebasePod(secrets []corev1.Secret, images BuildPodImages, config
 	}, nil
 }
 
-func (b *Build) cacheVolume(os string) corev1.VolumeSource {
-	if b.Spec.CacheName != "" && os != "windows" {
-		return corev1.VolumeSource{
+func (b *Build) cacheVolume(os string) []corev1.Volume {
+	if b.Spec.CacheName == "" {
+		return []corev1.Volume{}
```
matthewmcnew, line 640: Can we return the this nonexistent cacheVolume on windows instead of the empty dir?

phil9909: I don't know why an `emptyDir` is used for Windows until now, so I didn't want to change that.
Also, I don't have access to a Windows machine to test it. But I can update the PR, if you think it's safe.

matthewmcnew: We don't (hopefully yet) support persistent volume claims on windows so similar to linux builds we just used an emptyDir. However, with this PR on linux we just don't mount a cache dir if no cache is provided so it makes sense to just apply a similar strategy on windows.

phil9909: We updated the PR: We now use the same strategy for Windows (always) as for Linux when `CacheName` is empty.
As I mentioned before, I can not test this as I don't have an Windows Environment.

matthewmcnew (approved):
Thanks! 😍

pkg/apis/build/v1alpha1/build_pod.go
matthewmcnew: Awesome! We can verify this on windows @phil9909.

tomkennedy513 (approved):

matthewmcnew:
@phil9909 could you rebase real fast?

phil9909:
> @phil9909 could you rebase real fast?

Sorry, I'm in a different timezone and had already started my weekend.

Rebase is done.

PR #668, opened 2021-04-20
Try reading kubeconfig from env

modulo11:
This is a small quality of life improvement for running e2e tests locally. It will first try to read the `KUBECONFIG`environment variable and use the former mechanism as a fallback.

### DIFF
--- a/test/e2e.go
+++ b/test/e2e.go
@@ -2,6 +2,7 @@ package test
 
 import (
 	"flag"
+	"os"
 	"os/user"
 	"path"
 	"sync"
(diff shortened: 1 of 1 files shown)


test/e2e.go
```diff
@@ -57,6 +53,17 @@ func newClients(t *testing.T) (*clients, error) {
 	}, nil
 }
 
+func getKubeConfig() string {
+	config, found := os.LookupEnv("KUBECONFIG")
+	if found {
+		return config
+	}
```
tomkennedy513, line 60:
```suggestion
	if config, found := os.LookupEnv("KUBECONFIG"); found {
		return config
	}
```
just to match what you do below

tomkennedy513 (approved):

matthewmcnew (approved):

matthewmcnew:
Thanks @modulo11 !

Issue #652, opened 2021-03-29
Allow using a remote registry for caching


samj1912:
As dicussed in slack at https://buildpacks.slack.com/archives/CQ10WHMT4/p1613732146033900

TL;DR - 

Platforms like `pack` allow storing the cache layers in a remote registry. `kpack` should allow doing the same for clusters that don't have PVCs or for more flexibility with caching in general. This might require a spec change from `spec.cacheSize` to `spec.cache` which looks like - 

Something like:
```
spec: 
  cache: 
    volume: 
      request: 2G
      storageClassName: some-class-name
```
and:
```
spec: 
  cache: 
    registry: 
      tag: myregistry.com/image/tag
```

divoxx:
It would also be great if the two caching strategies can be used in conjunction. Being able to have local builds as fast as possible, while still allowing some level of caching for multi-clusters setup, or when the cache needs to be busted.

matthewmcnew:
> It would also be great if the two caching strategies can be used in conjunction. i.e. being able to have local builds as fast as possible, while still allowing some level of caching for multi-clusters setup, or when the cache needs to be busted. 

@divoxx Interesting, idea! I think that might require a change to the underlying cloud native buildpack project to either support multiple types of cache or some cache hierarchy. Would you be interested in bringing that up in Buildpacks slack or Buildpacks working group?

PR #636, opened 2021-03-04
Utilize libgit2 as git implementation

matthewmcnew:
- All git actions use git2go  
- Updates github actions to build in the golang alpine image with libgit2 installed. 

Closes https://github.com/pivotal/kpack/pull/620
Resolves https://github.com/pivotal/kpack/issues/213

Issue #213, opened 2019-11-09
Azure DevOps not able fetch code from Git repo


amitmele:
Using Azure DevOps private registry, 

Authorisation is successful but 
While building Image getting bellow error

PS C:\Test\kpack\kpack-ms> .\logs.exe -image sample-image
prepare:main.go:78: unable to fetch git repository: unexpected client error: unexpected requesting "https://private-test.visualstudio.com/PROJECT/_git/test-sb-kp-ms/git-upload-pack" status code: 400

amitmele:
Ref May Help: https://github.com/argoproj/argo-cd/issues/1067, issue mostly with go-git library

matthewmcnew:
It looks like this is a direct result of the issue in the go-git library. https://github.com/src-d/go-git/issues/1058

lsilvapvt:
Any plans to address this issue with the product? It prevents the adoption of the product by any customer that uses Azure DevOps repositories in the field.

tylerphelan:
new issue: https://github.com/go-git/go-git/issues/64

rajivmca2004:
Any update on this, when it will be fixed? Our partner/customer is getting impacted with this.

phillebaba:
@rajivmca2004 I have created a PR https://github.com/go-git/go-git/pull/200 with a fix that would make it work for Azure DevOps.

djoyahoy:
@phillebaba Thank you for letting us know and for your hard work on that PR!

dirien:
hi, 

i just wanted to mention that @phillebaba enabled in the GitOps Toolkit the git2go library as an alternative. This approach could be useful here too?

phillebaba:
Yep so our solution after a failed attempt at adding multi_ack support to go-git was to switch over to git2go. It is probably the only other proper git client for golang. After spending some time with reading git wire protocol specs I can say that fully implementing multi_ack and multi_ack_detailed is not an easy job in go-git and would require someone who could work on it full time for some weeks. This makes implementing it in go-git improbable.

Saying that, git2go does have its limitations as well that I found along the way. Hopefully this would help you guys when choosing if it is right for you to switch.
* git2go is a golang binding for libgit2 which is built in C. This means that if you want to run anything using git2go locally you will need to install the libgit2 library. This is pretty easy on macOS and most linux distros, not to sure about Windows. If you run everything in a container this really is not an issue.
[...]

PR #669, opened 2021-04-20
Image registry cache

modulo11:
This implements parts of #652 and additionally offers images to be used as caches. The current default behavior is not changed but some parts are renamed for better clarification.

Currently, it's an either volume or registry cache although supporting both (as depicted in #652) can be achieved easily. Using both caching arguments (`-cache-dir` and `-cache-image`) with the CNB lifecycle resulted in only utilizing the cache image (during my experiments).

@c0d1ngm0nk3y
@loewenstein


pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -165,16 +164,19 @@ func (b *Build) BuildPod(images BuildPodImages, secrets []corev1.Secret, taints
 
 	var cacheArgs []string
 	var cacheVolumes []corev1.VolumeMount
-	if b.Spec.CacheName == "" || config.OS == "windows" {
+	if b.Spec.Cache.ImageTag == "" && b.Spec.Cache.VolumeName == "" || config.OS == "windows" {
```
matthewmcnew, line 167: The `home` volume with registry creds will need to be mounted in the restore step now to provide access to the registry cache.

pkg/apis/build/v1alpha1/build_types.go
```diff
@@ -59,7 +59,7 @@ type BuildSpec struct {
 	Builder        BuildBuilderSpec `json:"builder,omitempty"`
 	ServiceAccount string           `json:"serviceAccount,omitempty"`
 	Source         SourceConfig     `json:"source"`
-	CacheName      string           `json:"cacheName,omitempty"`
+	Cache          BuildCacheConfig `json:"cache,omitempty"`
```
matthewmcnew, line 62: We should probably support this being nil?

pkg/apis/build/v1alpha1/build_types.go
```diff
@@ -69,12 +69,17 @@ type BuildSpec struct {
 	Notary    *NotaryConfig               `json:"notary,omitempty"`
 }
 
+type BuildCacheConfig struct {
+	VolumeName string `json:"volumeName,omitempty"`
+	ImageTag   string `json:"imageTag,omitempty"`
```
matthewmcnew, line 74: In the spirit of reproducibility should we provide an image with a fully qualified digest here?

matthewmcnew:
Thanks for submitting this @modulo11! It will be a breaking change to the image (and build) api so we need to provide it in a new api version. This [PR for service bindings](https://github.com/pivotal/kpack/pull/555) introduces the v1alpha2 api. I think it would make sense to include this api update alongside that api update.

modulo11:
Hi @matthewmcnew thanks for the review. We tried bumping the api from v1alpha1 to v1alpha2 for just `Build` and `Image` types analogous to PR #555. We gave up in the middle because we we're overwhelmed by subtle changes which are easily overlooked when mixing both versions in the code. For us it would be more feasible to bump all api objects to v1alpha2 and update all references.

We use a `latest` api package (referencing the current api object versions) in our own controllers, so bumping all versions can be done in one place. We can supply a (draft) PR if you're interested.

WDYT?

pkg/apis/build/v1alpha1/build_types.go
modulo11: We think no: If "someone" pushed an update to the cache we'd like to always use the most current version.

matthewmcnew: The current [image reference](https://github.com/pivotal/kpack/blob/d5c86c908752e6271eb27179a18ae493705d9007/pkg/apis/build/v1alpha1/build_pod.go#L301-L305) that is provided to the analyze step is the fully qualified built image reference from the last build. This allows deterministic builds that will not run into outside interference if something else is writing to the same tag. I think the same thing applies here, another image could write to the provided cache image tag and we would want to ensure we are using the exact cache digest from the last build.

matthewmcnew:
I'm not sure of the best path forward as implementing a new api version will be quite involved. Likely we should merge our [PR for service bindings](https://github.com/pivotal/kpack/pull/555) that will introduce a new api version and then allow you to provide your PR on top of that so you don't need to manage adding a new api version.

matthewmcnew:
To help me better understand the desired timelines could you tell me a bit more about your desired usecase for this functionality?

c0d1ngm0nk3y:
> I'm not sure of the best path forward as implementing a new api version will be quite involved. 

But we have doubts that this is the best way forward. The #555 shows the problems with the imports:

´´´
"github.com/pivotal/kpack/pkg/apis/build/v1alpha1"
"github.com/pivotal/kpack/pkg/apis/build/v1alpha2"
´´´

So you have to be very careful which api object take from which import ( ´Build´ -> v1alpha2, but ´SourceResolver` -> v1alpha1). Those errors can be quite subtle.

When bumping the ´build´ api as a whole, we could do something like 

´´´
build "github.com/pivotal/kpack/pkg/apis/build/v1alpha2"
´´´

and then changing the api version is changing only the import in each relevant file.

matthewmcnew:
Yeah, we should introduce an entire new api version for all resources at `v1apha2`

pkg/apis/build/v1alpha1/build_types.go
modulo11: Got it. Should be implemented with 6b3e580df3c99b7ed57dd1ac50730f3da1ed2474 and 177e1397ba8e3112cc9ae28fc2840e529f4c5071. I'll also add some test for that. WDYT?

modulo11:
Alright, then we'll put that PR on hold and continue after #555 is merged.

c0d1ngm0nk3y:
> Yeah, we should introduce an entire new api version for all resources at `v1alpha2`

But then, there would be no real benefit waiting for #555, right? `v1alpha2` can be changed again until it is released.

matthewmcnew:
> But then, there would be no real benefit waiting for #555, right? v1alpha2 can be changed again until it is released.

I'm not quite sure what you mean by this @c0d1ngm0nk3y. Are you suggesting that we create a different PR to bump to v1alpha?

c0d1ngm0nk3y:
> I'm not quite sure what you mean by this @c0d1ngm0nk3y. Are you suggesting that we create a different PR to bump to v1alpha?

Sorry, yes. Our idea was to create a different PR bumping to v1alpha2. And then both PRs could rebase on that and focus only on the relevant parts. We have started with some documentation and renaming to make the bumping not so "involved".

c0d1ngm0nk3y:
@matthewmcnew  Maybe you can have a look at #675 - we would propose to use this as a first step before actually bumping the api. We could then add another PR bumping the api to `v1alpha2`, but not changing anything. #555 and #669 could then base on that and only adapt what is really relevant. WDYT?

matthewmcnew:
Hi, @modulo11 & @c0d1ngm0nk3y. Thanks for your patience on this! Our current goal is to provide our planned v0.3 release of kpack in the next few weeks with the underlying git library moved to use [git2go libgit2](https://github.com/pivotal/kpack/pull/636) bindings as a fix to our long running [azure devops woes](https://github.com/pivotal/kpack/issues/213). Migrating/building kpack images to use the libgit2 has been a major hurdle so our plan is to release this before releasing any other major kpack updates.   

We would like to target the release of the new api version in a short follow-up v0.4 release. I think we are going to wait on merging this an other changes to the api until after the v0.3 release is released.

build-webhook.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -e
+
+pack build europe-west3-docker.pkg.dev/sap-se-gcp-istio-dev/public/kpack-667/webhook --builder gcr.io/cf-build-service-public/ci/tiny-builder -e BP_GO_TARGETS=./cmd/controller --publish --trust-builder
```
tylerphelan, line 5: could the the tag argument to `pack` be parameterized in this script? also `pack` the args have changed since this to 

```
-B gcr.io/cf-build-service-public/ci/kpack-builder -e BP_GO_TARGETS=./cmd/controller -e BP_BUILD_LIBGIT2=true -e BP_GO_BUILD_FLAGS="-tags=\"static\""  --publish
```

You could also just remove the scripts if you don't want to update them.

docs/build.md
```diff
@@ -19,7 +19,8 @@ spec:
     image: gcr.io/paketo-buildpacks/builder:base
     imagePullSecrets: 
     - name: builder-secret  
-  cacheName: persisent-volume-claim-name
+  cache:
+    volumeName: persisent-volume-claim-name
```
tylerphelan, line 23: I think `pvcName` is more accurate here

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -641,14 +652,14 @@ func (b *Build) rebasePod(secrets []corev1.Secret, images BuildPodImages, config
 }
 
 func (b *Build) cacheVolume(os string) []corev1.Volume {
-	if b.Spec.CacheName == "" || os == "windows" {
+	if b.Spec.Cache.VolumeName == "" || os == "windows" {
```
tylerphelan, line 655: do we need to protect from `.Cache` being `nil` here?

pkg/apis/build/v1alpha2/build_pod_test.go
```diff
@@ -668,9 +671,72 @@ func testBuildPod(t *testing.T, when spec.G, it spec.S) {
...
+
+				exportContainer := pod.Spec.InitContainers[5]
+				assert.NotContains(t, exportContainer.Args, "-cache-image")
+			})
+		})
+
 		when("CacheName is empty", func() {
 			podWithCache, _ := build.BuildPod(config, nil, nil, buildPodBuilderConfig)
-			build.Spec.CacheName = ""
+			build.Spec.Cache.VolumeName = ""
```
tylerphelan, line 739: I think this test should be with `build.Spec.Cache = nil`

pkg/apis/build/v1alpha2/image_validation_test.go
```diff
@@ -262,9 +279,9 @@ func testImageValidation(t *testing.T, when spec.G, it spec.S) {
 		it("image.cacheSize has not decreased", func() {
```
tylerphelan, line 279: might be nice to update the text here

pkg/cnb/cnb_metadata.go
```diff
@@ -53,6 +53,28 @@ func (r *RemoteMetadataRetriever) GetBuiltImage(ctx context.Context, build *buil
 	return readBuiltImage(appImage, appImageId)
 }
 
+func (r *RemoteMetadataRetriever) GetCacheImage(ctx context.Context, build *buildapi.Build) (string, error) {
+	if len(build.Spec.Cache.ImageTag) == 0 {
```
tylerphelan, line 57: can `build.Spec.Cache` be `nil`?

docs/image.md
```diff
@@ -9,7 +9,9 @@ The following defines the relevant fields of the `image` resource spec in more d
 - `builder`: Configuration of the `builder` resource the image builds will use. See more info [Builder Configuration](builders.md).
 - `serviceAccount`: The Service Account name that will be used for credential lookup.
 - `source`: The source code that will be monitored/built into images. See the [Source Configuration](#source-config) section below.
-- `cacheSize`: The size of the Volume Claim that will be used by the build cache.
+- `cache`: Caching configuration, two variants are available:
+  - `volume.request`: Creates a Volume Claim of the given size
```
tylerphelan, line 13: should this be `volume.size` ? `request` seems less user-friendly. anyone else have thoughts?

tylerphelan (changes requested):
very nice pr, I had just a few notes

c0d1ngm0nk3y:
> very nice pr, I had just a few notes

Thanks, we will have a look. Probably after #670 is merged because they are in conflict anyway.

docs/build.md
matthewmcnew: I wonder if we should match the long name in [VolumeSource](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#persistentvolumeclaimvolumesource-v1-core) and use the long name `persistentVolumeClaimName`?

docs/image.md
matthewmcnew: Hmm. I like how request matches the k8s terminology for PVCs.

matthewmcnew: Should this actually be requests and be a RequestsList?

pkg/apis/build/v1alpha2/build_types.go
```diff
@@ -69,20 +69,26 @@ type BuildSpec struct {
...
 	MetadataRef *corev1.LocalObjectReference `json:"metadataRef,omitempty"`
 	SecretRef   *corev1.LocalObjectReference `json:"secretRef,omitempty"`
 }
 
 // +k8s:openapi-gen=true
 type LastBuild struct {
-	Image   string `json:"image,omitempty"`
-	StackId string `json:"stackId,omitempty"`
+	Image      string `json:"image,omitempty"`
+	CacheImage string `json:"cacheImage,omitempty"`
```
matthewmcnew, line 90: To this point we have tried to keep every reference in the api to an image to use the `image` key. Could we move this to `cache.image`?

pkg/apis/build/v1alpha2/image_types.go
```diff
@@ -44,14 +44,27 @@ type ImageSpec struct {
...
+	Volume   *VolumeCache   `json:"volume,omitempty"`
+	Registry *RegistryCache `json:"registry,omitempty"`
+}
+
+type VolumeCache struct {
+	Request *resource.Quantity `json:"request,omitempty"`
+}
+
+type RegistryCache struct {
+	Tag string `json:"tag"`
```
c0d1ngm0nk3y, line 65: Shouldn't it be `ImageTag`? Why not `omitempty`?

pkg/apis/build/v1alpha2/image_builds.go
```diff
@@ -57,13 +57,35 @@ func (im *Image) Build(sourceResolver *SourceResolver, builder BuilderResource,
...
 		},
 	}
 }
 
+func (is *ImageSpec) NeedVolumeCache() bool {
+	return is.Cache != nil && is.Cache.Volume != nil && is.Cache.Volume.Request != nil
+}
+
+func (is *ImageSpec) NeedRegistryCache() bool {
+	return is.Cache != nil && is.Cache.Registry != nil && len(is.Cache.Registry.Tag) > 0
```
c0d1ngm0nk3y, line 72: Should we use `imageTag != ""` like in the rest of the code?

pkg/apis/build/v1alpha2/image_builds.go
```diff
@@ -57,13 +57,35 @@ func (im *Image) Build(sourceResolver *SourceResolver, builder BuilderResource,
...
+
+func (im *Image) getBuildCacheConfig() *BuildCacheConfig {
+	buildCacheConfig := BuildCacheConfig{}
+
+	if im.Spec.NeedRegistryCache() {
+		buildCacheConfig.ImageTag = im.Spec.Cache.Registry.Tag
+	}
+
+	if im.Spec.NeedVolumeCache() {
+		buildCacheConfig.VolumeName = im.Status.BuildCacheName
```
c0d1ngm0nk3y, line 83: Why is it taken from the `Status` and not the `Spec`?

pkg/apis/build/v1alpha2/image_validation.go
```diff
@@ -45,8 +45,12 @@ func (i *Image) SetDefaults(ctx context.Context) {
 		i.Spec.SuccessBuildHistoryLimit = &defaultSuccessfulBuildHistoryLimit
 	}
 
-	if i.Spec.CacheSize == nil && ctx.Value(HasDefaultStorageClass) != nil {
-		i.Spec.CacheSize = &defaultCacheSize
+	if i.Spec.Cache == nil && ctx.Value(HasDefaultStorageClass) != nil {
```
c0d1ngm0nk3y, line 48: Is this correct? Should we check for `i.Spec.Cache.Request != nil` as well? What if it uses `registryCache`?

pkg/apis/build/v1alpha2/image_validation.go
```diff
@@ -89,18 +93,20 @@ func (is *ImageSpec) validateTag(ctx context.Context) *apis.FieldError {
 	return validate.Tag(is.Tag)
 }
 
-func (is *ImageSpec) validateCacheSize(ctx context.Context) *apis.FieldError {
-	if is.CacheSize != nil && ctx.Value(HasDefaultStorageClass) == nil {
-		return apis.ErrGeneric("spec.cacheSize cannot be set with no default StorageClass")
+func (is *ImageSpec) validateVolumeCache(ctx context.Context) *apis.FieldError {
+	if is.Cache.Volume != nil && ctx.Value(HasDefaultStorageClass) == nil {
```
c0d1ngm0nk3y, line 97: What is with `is.Cache != nil`?

pkg/apis/build/v1alpha2/image_validation_test.go
```diff
@@ -38,7 +38,11 @@ func testImageValidation(t *testing.T, when spec.G, it spec.S) {
 					Revision: "master",
 				},
 			},
-			CacheSize:                &cacheSize,
+			Cache: &CacheConfig{
```
c0d1ngm0nk3y, line 41: What is with an empty `volumeCache`?

pkg/apis/build/v1alpha2/image_validation_test.go
c0d1ngm0nk3y: @tylerphelan Which text? 😕 Sorry...

pkg/reconciler/image/image_test.go
```diff
@@ -713,6 +717,7 @@ func testImageReconciler(t *testing.T, when spec.G, it spec.S) {
 									Image: builder.Status.LatestImage,
 								},
 								ServiceAccount: image.Spec.ServiceAccount,
+								Cache:          &buildapi.BuildCacheConfig{},
```
c0d1ngm0nk3y, line 720: Shouldn't `nil` be valid as well?

c0d1ngm0nk3y (changes requested):

test/execute_build_test.go
```diff
@@ -357,68 +361,89 @@ func testCreateImage(t *testing.T, when spec.G, it spec.S) {
...
+		Limits: corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse("1G"),
+		},
+		Requests: corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse("512M"),
+		},
+	}
 
-		imageName := fmt.Sprintf("%s-%s", "test-git-image", "cluster-builder")
+	imageName := fmt.Sprintf("%s-%s", "test-git-image", "cluster-builder")
```
c0d1ngm0nk3y, line 398: Maybe it is worth to use different `imageNames`? Since this `generateRebuild` is called twice (for each cache).

pkg/apis/build/v1alpha2/image_types.go
```diff
@@ -44,7 +44,7 @@ type ImageSpec struct {
 	Builder                  corev1.ObjectReference `json:"builder,omitempty"`
```
matthewmcnew, line 44: Should we update the Image status to include the lastCacheImage so that the cache is not lost when the last build failed?

pkg/reconciler/build/build.go
```diff
@@ -110,8 +111,14 @@ func (c *Reconciler) reconcile(ctx context.Context, build *buildapi.Build) error
 			return err
 		}
 
+		cacheImageId, err := c.MetadataRetriever.GetCacheImage(ctx, build)
```
matthewmcnew, line 114: It would be interesting to see if we could acquire this from `report.toml`. Perhaps we should create a cnb rfc?

pkg/apis/build/v1alpha2/build.go
```diff
@@ -96,6 +96,17 @@ func (b *Build) BuiltImage() string {
 	return b.Status.LatestImage
 }
 
+func (b *Build) CacheImage() string {
+	if b == nil {
+		return ""
+	}
+	if !b.IsSuccess() {
+		return ""
```
matthewmcnew, line 104: This is interesting. If the last build failed no cache image will be provided. This means the cache can get wiped with a single failed build due to a transient network error, etc. I wonder if we could find a way to make the provided cache image more resilient to build failures. This could be done by storing the last successful cache image or perhaps just providing the tag of cache image?

c0d1ngm0nk3y: We don't quite get it. The cache image will still be on the registry. It won't be updated with the failing build, but this is probably what we want, right?
This is used in `Build.Spec.LastBuild` which is used for `rebase`. But do we really need at a `Build.Spec.LastBuild.Cache` anyway? And also `Build.Spec` is immutable. So I think, we will just remove `Build.Spec.LastBuild.Cache`.

pkg/apis/build/v1alpha2/image_validation_test.go
c0d1ngm0nk3y: We got it :) Thanks.

pkg/apis/build/v1alpha2/image_types.go
c0d1ngm0nk3y: Can you explain it a bit more? The cache image will still be there, right? But not updated.

c0d1ngm0nk3y:
@tylerphelan @matthewmcnew We think we have addressed all feedback. Except the part with "cache image lost on failure" which we didn't understand 100% to be honest. Could you have another look?

docs/image.md
dumez-k:
```
- `cache`: Caching configuration, two variants are available:
  - `volume.size`: Creates a Volume Claim of the given size
  - `registry.imageTag`: Creates an image with cached contents
 ```

pkg/apis/build/v1alpha2/image_validation.go
tylerphelan: I think this is right!

pkg/apis/build/v1alpha2/image_validation.go
tylerphelan: with the current implementation it will get defaulted, so I think it's ok

pkg/apis/build/v1alpha2/image_validation_test.go
tylerphelan: I actually meant the test text: `it("image.cacheSize has not decreased", func() {` -> `it("image.cache.volume.size has not decreased", func() {` 😀

tylerphelan:
Nice, all of my comments have been addressed. I had one more thought, what do others think about adding validation ensuring only one of `image.spec.cache.volume` and `image.spec.cache.registry` is set? Right now the registry will be used by the lifecycle and in the future multiple cache types may be possible, but it might be nice to validate this for now.

pkg/reconciler/build/build.go
c0d1ngm0nk3y: When comparing it to the actual logs of the image pushed to the registry, it is somehow the wrong `sha256` and the content of `LastCacheImage` can't be pulled. So we should look into that!!!

c0d1ngm0nk3y (changes requested):

c0d1ngm0nk3y:
> I had one more thought, what do others think about adding validation ensuring only one of `image.spec.cache.volume` and `image.spec.cache.registry` is set?

Good point. We have added validation to `Build` and `Image` that only one type of cache can be used.

c0d1ngm0nk3y:
@matthewmcnew  @tylerphelan We have rebased and squashed this pr and from our point of view we are good to go. Please check and let us know if there is still something else to do.

PR #675, opened 2021-04-28
Introducing stable alias for buildapi to ease version bumping

c0d1ngm0nk3y:
Instead of doing those changes for each bump in the build api (`v1alpha2`, `v1beta1`, ...), we introduce a stable alias "buildapi" which points to the api version which is used (currently still `v1alpha1`).

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -27,7 +27,7 @@ import (
 	"knative.dev/pkg/signals"
 
 	"github.com/pivotal/kpack/cmd"
-	"github.com/pivotal/kpack/pkg/apis/build/v1alpha1"
+	buildapi "github.com/pivotal/kpack/pkg/apis/build/v1alpha1"
 	"github.com/pivotal/kpack/pkg/blob"
 	"github.com/pivotal/kpack/pkg/buildpod"
 	"github.com/pivotal/kpack/pkg/client/clientset/versioned"
@@ -40,7 +40,7 @@ import (
 	"github.com/pivotal/kpack/pkg/reconciler"
 	"github.com/pivotal/kpack/pkg/reconciler/build"
 	"github.com/pivotal/kpack/pkg/reconciler/builder"
-	"github.com/pivotal/kpack/pkg/reconciler/clusterbuilder"
+	clusterBuilder "github.com/pivotal/kpack/pkg/reconciler/clusterbuilder"
 	"github.com/pivotal/kpack/pkg/reconciler/clusterstack"
 	"github.com/pivotal/kpack/pkg/reconciler/clusterstore"
 	"github.com/pivotal/kpack/pkg/reconciler/image"
@@ -125,7 +125,7 @@ func main() {
 	}
 
 	buildpodGenerator := &buildpod.Generator{
-		BuildPodConfig: v1alpha1.BuildPodImages{
+		BuildPodConfig: buildapi.BuildPodImages{
 			BuildInitImage:         *buildInitImage,
 			CompletionImage:        *completionImage,
 			RebaseImage:            *rebaseImage,
@@ -236,8 +236,8 @@ func runGroup(ctx context.Context, fns ...doneFunc) error {
 	return eg.Wait()
 }
 
-func newBuildpackRepository(keychain authn.Keychain) func(clusterStore *v1alpha1.ClusterStore) cnb.BuildpackRepository {
-	return func(clusterStore *v1alpha1.ClusterStore) cnb.BuildpackRepository {
+func newBuildpackRepository(keychain authn.Keychain) func(clusterStore *buildapi.ClusterStore) cnb.BuildpackRepository {
+	return func(clusterStore *buildapi.ClusterStore) cnb.BuildpackRepository {
 		return &cnb.StoreBuildpackRepository{
 			Keychain:     keychain,
 			ClusterStore: clusterStore,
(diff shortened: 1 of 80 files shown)


c0d1ngm0nk3y:
I resolved the merge conflict with the generated file. @matthewmcnew Should we continue with this? WDYT?

matthewmcnew:
Tagging this as `0.4.0` to target releasing alongside the api changes in our `0.4.0` release.

c0d1ngm0nk3y:
> I know @tomkennedy513 had concerns on this. Could you share them here?

That would be good to hear. Our main concern about using different versions of build api at once like in #555 is that it becomes very error prone and the errors might get very subtle.

tomkennedy513:
My main concern was that it is now vague which version we are using. Also, I assumed that going to a new api version would be due to changes in the underlying structure so bumping the api isn't as simple as changing the import. if we were to add a new source type for an image or change the cache structure simply changing the import wouldn't work, but I do believe it would greatly reduce the surface area of changes required. 

That being said, I actually think the value of preventing multiple versions from being used at once may make this worth accepting, unless there is a good reason to allow multiple versions (ie bumping some resource versions but not all)

c0d1ngm0nk3y:
> My main concern ...

That is right, but I would argue that this is one of the nice things of this approach. When doing structure changes, it probably involves 1 or 2 files. Using a hardcoded api version everywhere in the code, causes the risk that latest api version is used if the structural change is not relevant.
Correct me if I am wrong, but except from the api itself (`pkg/api/<oldversion>`), there is no need to ever use a version that is not the latest, right? So why hardcode it all over the place when it has to be changed everytime anyway?

> bumping the api isn't as simple as changing the import

For most file it should be, shouldn't it?

> bumping some resource versions but not all

What would be a valid reason to bump only a specific resource? Wouldn't you potentially end with something like...
* `Build` exists in `v1alpha1`,`v1alpha2` and `v1alpha3`
* `ClusterStore` exists in `v1alpha1` and `v1alpha2
* `ClusterStack` only exists in `v1alpha1`

imho, each api version should be "complete" and the code should only work with the latest version (except webhooks). But this is enforced by using `served: true` on the latest version, right?

tomkennedy513 (approved):

c0d1ngm0nk3y:
@tomkennedy513 @matthewmcnew This pr has still a `needs discussion` label. If I recall correctly, there were no open questions in the last community meeting and we agreed this could be merged. Let me know if there are still open points!

c0d1ngm0nk3y:
I have rebased it and it is not based on 9e103a523e624be1f4e17424853500d59a00428d (HEAD of main). I do not know why it is still shown as "out-of-date". Any ideas?

tomkennedy513:
@c0d1ngm0nk3y the rest of the team is working on reviewing this, but in the meantime is there any way you could squash these commits down

c0d1ngm0nk3y:
> @c0d1ngm0nk3y the rest of the team is working on reviewing this, but in the meantime is there any way you could squash these commits down

I normally keep the commits because they sometimes help reviewing and use the "squash and merge" feature from github. But for `kpack`, the PR gets automatically merged once reviewed, right? I will squash them manually.

tomkennedy513:
> > @c0d1ngm0nk3y the rest of the team is working on reviewing this, but in the meantime is there any way you could squash these commits down
> 
> I normally keep the commits because they sometimes help reviewing and use the "squash and merge" feature from github. But for `kpack`, the PR gets automatically merged once reviewed, right? I will squash them manually.

They actually don't get merged automatically so we can use the squash and merge feature when we do merge it

pkg/client/informers/externalversions/build/v1alpha1/builder.go
```diff
@@ -22,10 +22,10 @@ import (
 	"context"
 	time "time"
 
-	buildv1alpha1 "github.com/pivotal/kpack/pkg/apis/build/v1alpha1"
+	buildapi "github.com/pivotal/kpack/pkg/apis/build/v1alpha1"
```
matthewmcnew, line 25: This file is autogenerated so we should leave it unmodified.

api/openapi-spec/swagger.json
```diff
@@ -4266,7 +4266,8 @@
           "$ref": "#/definitions/io.k8s.api.core.v1.LocalObjectReference"
         },
         "name": {
-          "type": "string"
+          "type": "string",
```
matthewmcnew, line 4269: I'm having trouble recreating this file. Which version of `openapi-gen` are you using? I am running against [release-1.19](https://github.com/kubernetes/kube-openapi) locally.

matthewmcnew (changes requested):
I believe we should not modify the generated code in: `pkg/client/`.

api/openapi-spec/swagger.json
c0d1ngm0nk3y: How to check it? This is a bit of a strange tool. It was a bit of a struggle to install. There are no releases on github or aome "install" section. No `tags` in git either. And now that I finally have it running, there is no `version` or `--version` when calling `openapi-gen`.

@matthewmcnew Could you point me to how to install `openapi-gen 1.19`?

pkg/client/informers/externalversions/build/v1alpha1/builder.go
c0d1ngm0nk3y: > This file is autogenerated so we should leave it unmodified.

I've reset `pkg/client` to `main`.

api/openapi-spec/swagger.json
c0d1ngm0nk3y: But for now, I've reset `swagger.json` to `main`. There should be no change anyway, right?

matthewmcnew (approved):

matthewmcnew:
Thanks!

PR #670, opened 2021-04-21
 Support private registry for ClusterStack & ClusterStore

c0d1ngm0nk3y:
Introducing an optional `serviceAccountRef` for `ClusterStore` and `ClusterStack` in order to allow credentials to a private registry.

Fixes #649

PR #676, opened 2021-04-28
Bump build api from v1alpha1 v1alpha2

c0d1ngm0nk3y:
This only introduces the api version `v1alpha2` with no changes yet.

This is based on #675

This PR would be the common base for #555 and #669 which both introduce incompatible changes.


c0d1ngm0nk3y:
@matthewmcnew Could you have a look? This takes the "dumb" bumping of #555 and #669. So the PRs could rebase and focus on needed changes to the api. Could this be a way forward? WDYT?

c0d1ngm0nk3y:
@matthewmcnew Could you have a look? This takes the "dumb" bumping of #555 and #669. So the PRs could rebase and focus on needed changes to the api. Could this be a way forward? WDYT?

modulo11:
Note: After 0.3.0 is released, we'll review all changes and copy them to the next api version.

c0d1ngm0nk3y:
@matthewmcnew @tomkennedy513 As discussed in the last community meeting, this is the follow up on #675

c0d1ngm0nk3y:
I have rebased it and it is not based on 9e103a523e624be1f4e17424853500d59a00428d (HEAD of main). I do not know why it is still shown as "out-of-date". Any ideas?

config/builder.yaml
```diff
@@ -6,6 +6,22 @@ spec:
   group: kpack.io
   versions:
   - name: v1alpha1
+    served: true
+    storage: false
```
matthewmcnew, line 10: Should we store this version?

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -0,0 +1,837 @@
...
+	networkWaitLauncherVolume = corev1.VolumeMount{
+		Name:      networkWaitLauncherDir,
+		MountPath: "/networkWait",
+		ReadOnly:  false,
+	}
+)
+
+type stepModifier func(corev1.Container) corev1.Container
+
+func (b *Build) BuildPod(images BuildPodImages, secrets []corev1.Secret, taints []corev1.Taint, config BuildPodBuilderConfig) (*corev1.Pod, error) {
```
matthewmcnew, line 133: Should me delete the equivalent of these domain methods in v1alpha1?

matthewmcnew (approved):

matthewmcnew:
I am a bit concerned about the proliferation of api versions and the added complexity introduced with multiple conversions/etc. I wonder if similar to tekton we should jump to v1beta1 or introduce v1alpha1 as unstable "testing" version before ultimately releasing it as a stable api in v1alpha1. 

I'll add it to the agenda as this would be a great topic to rediscuss during working group.

config/builder.yaml
c0d1ngm0nk3y: We can only store one version and I think it is easier to store `v1alpha1` objects as `v1alpha2` objects (e.g. no new features are used) than to store `v1alpha2`  as `v1alpha1` (any new feature must be stored in the annotation to not get lost).

pkg/apis/build/v1alpha2/build_pod.go
c0d1ngm0nk3y: You are right I guess. I have removed them, but in a separate commit. So it is easier to revert again. I can squash it, once this is ready to merge.

c0d1ngm0nk3y:
> I am a bit concerned about the proliferation of api versions and the added complexity introduced with multiple conversions/etc. I wonder if similar to tekton we should jump to v1beta1 or introduce v1alpha1 as unstable "testing" version before ultimately releasing it as a stable api in v1alpha1.

There is definitely some complexity coming with having more versions. But on the other hand, adding new features like #669 #555 or #670 and going from `alpha` to `beta` might contradict.

> I'll add it to the agenda as this would be a great topic to rediscuss during working group.

Thanks.

loewenstein:
I'd argue that alpha is the right stage to start practicing versioning of CRDs. Once something is `v1`, making mistakes with a potential successor version will be worse.

config/builder.yaml
matthewmcnew: Yep, that makes sense. I read that wrong.

pkg/apis/build/v1alpha2/build_pod.go
modulo11: Actually we'd like to keep them to minimize the manual effort when doing a version bump. Also we should not touch old versions to not break existing code (maybe deprecate the methods instead).

tomkennedy513 (approved):

matthewmcnew (approved):

matthewmcnew:
Thanks @c0d1ngm0nk3y!

PR #678, opened 2021-04-29
Add rbac for k8s leases on the webhook

matthewmcnew:


config/webhookrole.yaml
```diff
@@ -35,6 +35,18 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - "coordination.k8s.io"
+  resources:
+  - "leases"
+  verbs:
```
dumez-k, line 42: notice we don't have `deletecollection` in the verb [list](https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/lease-v1/) but have all the others, is this intentional?

matthewmcnew: Good Question! I took this list from `tekton`: https://github.com/tektoncd/pipeline/blob/1f5980f8c8a05b106687cfa3e5b3193c213cb66e/config/200-role.yaml#L80. I don't believe we utilize the delete collection verb.

dumez-k (approved):

PR #686, opened 2021-05-08
Add a helm chart for kpack

### LINKED ISSUE #685: Add a helm chart for kpack
It will be useful to have a helm chart for kpack deployments.

samj1912:
Creates a helm chart to manage kpack installations.

Fixes #685 

Still a draft as I wanted to discuss if the current parametrization is good enough and how to add this to the release process.

### DIFF
--- a/helm/.helmignore
+++ b/helm/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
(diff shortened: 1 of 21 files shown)


samj1912:
The second commit is because of #687 . I have currently kept the config map names as expected by kpack, but I may create a concurrent PR to make this name configurable.

samj1912:
> We should split the CRDs into their own Helm chart with the current handling of CRDs in Helm.
> 
> They will only be applied during install, not during update. Nor will they be removed during delete.

I guess the main issue with isolating the crds into a separate helm chart is that you could possibly have users install the kpack operator without the CRDs. 
I would leave it to the maintainers to decide which workflow they prefer.

loewenstein:
Well, isolating and not isolating CRDs both has potential pitfalls.

Keeping the CRDs in means people will have to invent their own ways to update CRDs.

PR #689, opened 2021-05-09
Make lifecycle config path configurable

### LINKED ISSUE #687: kpack hardcodes the lifecycle config map name
As a part of testing the helm chart release in #686 I realized that kpack currently hardcodes the lifecycle config at https://github.com/pivotal/kpack/blob/ed7fe5056666f0ec1aa4378c92db1d52a0a872c5/cmd/controller/main.go#L160

This makes it hard to provide arbitrary configmap names for it. We can fix this by passing the configmap name through an env var or keep this hardcoded in the helm chart. I think the former might be preferred to avoid conflicts with people accidentally changing this configmap even though the kpack lifecycle image is being tracked through a different config map name.

samj1912:
kpack currently hardcodes the name of the lifecycle image configmap. This commit makes this cconfigurable using a flag/env var and defaults it to the current value of "lifecycle-image" in the release config.

Fixes #687

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -40,7 +40,7 @@ import (
 	"github.com/pivotal/kpack/pkg/reconciler"
 	"github.com/pivotal/kpack/pkg/reconciler/build"
 	"github.com/pivotal/kpack/pkg/reconciler/builder"
-	"github.com/pivotal/kpack/pkg/reconciler/clusterbuilder"
+	clusterBuilder "github.com/pivotal/kpack/pkg/reconciler/clusterbuilder"
 	"github.com/pivotal/kpack/pkg/reconciler/clusterstack"
 	"github.com/pivotal/kpack/pkg/reconciler/clusterstore"
 	"github.com/pivotal/kpack/pkg/reconciler/image"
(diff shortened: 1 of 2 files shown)


cmd/controller/main.go
```diff
@@ -40,7 +40,7 @@ import (
 	"github.com/pivotal/kpack/pkg/reconciler"
 	"github.com/pivotal/kpack/pkg/reconciler/build"
 	"github.com/pivotal/kpack/pkg/reconciler/builder"
-	"github.com/pivotal/kpack/pkg/reconciler/clusterbuilder"
+	clusterBuilder "github.com/pivotal/kpack/pkg/reconciler/clusterbuilder"
```
samj1912, line 43: This  change aws made by gofmt.

matthewmcnew:
An additional complication on this is the [kp cli](https://github.com/vmware-tanzu/kpack-cli) is hardcoded to use the lifecycle image configmap in the kpack namespace. We could attempt to make the kp cli smarter but, that may require the end user of kp to have elevated privileges. Do you have additional examples on when this might be utilized?

matthewmcnew:
I added a needs discussion label to the outstanding question regarding the kp cli.

PR #690, opened 2021-05-09
Add support for project descriptor

### LINKED ISSUE #288: Add support for project.toml
Add support for the [Cloud Native Buildpacks Project descriptor RFC](https://github.com/buildpacks/rfcs/pull/32).

samj1912:
This PR adds support for the following attributes of a project descriptor - 

1. Build Env Variables
2. Includes and excludes

The implementation of the project descriptor is pretty similar to the pack implementation.
- The build env vars set by the project descriptor are overridden by the ones set by the user explicitly.
- The includes and excludes use the go-gitignore library which is also used by pack and required by the spec (the spec says that these values are to be interpreted in the same way as gitignore)

Fixes #288

### DIFF
--- a/pkg/cnb/project_descriptor.go
+++ b/pkg/cnb/project_descriptor.go
@@ -1,0 +1,88 @@
+package cnb
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/BurntSushi/toml"
+	ignore "github.com/sabhiram/go-gitignore"
+)
+
+const fileDescriptorName = "project.toml"
+
+func ProcessProjectDescriptor(appDir, platformDir string) error {
+	var d descriptor
+	file := filepath.Join(appDir, fileDescriptorName)
+	if _, err := os.Stat(file); os.IsNotExist(err) {
+		return nil
+	} else if err != nil {
+		return fmt.Errorf("unable to determine if project descriptor file exists: %w", err)
+	}
+	_, err := toml.DecodeFile(file, &d)
+	if err != nil {
+		return err
+	}
+	if err := processFiles(appDir, d); err != nil {
+		return err
+	}
+	return serializeEnvVars(d.Build.Env, platformDir)
+}
+
+func processFiles(appDir string, d descriptor) error {
+	fileFilter, err := getFileFilter(d)
+	if err != nil {
+		return err
+	}
+	if fileFilter == nil {
+		return nil
+	}
+	return filepath.Walk(appDir, func(path string, f os.FileInfo, fileError error) error {
+		if fileError != nil {
+			return fileError
+		}
+		relPath, err := filepath.Rel(appDir, path)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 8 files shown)


pkg/cnb/project_descriptor.go
```diff
@@ -0,0 +1,89 @@
...
+	fileFilter, err := getFileFilter(d)
+	if err != nil {
+		return err
+	}
+	if fileFilter == nil {
+		return nil
+	}
+	return filepath.Walk(appDir, func(path string, f os.FileInfo, err error) error {
+		if err != nil {
+			return nil
```
matthewmcnew, line 42: Should this return the `err`?

matthewmcnew:
Excellent work @samj1912! 

I wanted to call out the some questions I had: 
1. How should we handle project descriptor specified `build.buildpacks`? Is that something you plan on using with kpack? Would it make sense to rewrite the order if the provided buildpacks are in the running builder?
2. Should we do anything if a descriptor `build.builder` is specified? Perhaps log a warning to the user?
3. Would it make sense to validate the license similar to what [pack currently does](https://github.com/buildpacks/pack/blob/8e2a9282b1b99118eac1b4a0129c8b4d576073b6/project/project.go#L64-L70)?

pkg/cnb/project_descriptor.go
matthewmcnew: It seems like if we run into an error walking the directory we might not be able to accurately apply the includes/excludes and we should error out?

samj1912: Makes sense. Will update :)

samj1912: Should be fixed now :)

tylerphelan:
I think long term we should have a solution to the `build.buildpacks` use case but for now we could move forward with this! We just need make sure we document that `build.buildpacks` is not supported currently.

tylerphelan (approved):

tomkennedy513 (approved):

matthewmcnew (approved):

matthewmcnew:
To document the discussion from our May 24th, OSS WG: 

Questions 1/2 (How to handle build.buildpacks build.builder?). It is unclear how we will handle `build.buildpacks` due to the needs of a kpack "operator" to ensure that kpack image's are using specific buildpacks and the need to prevent buildpacks that are not blessed by the operator (For example: a buildpack that does not provide accurate BOM metadata). This conversation will likely be complicated by the introduction of inline buildpacks that users would likely expect to work on kpack similarly to pack. In the meantime, [we are going to just log an info message to the user if the supplied project.toml contains those keys](https://github.com/pivotal/kpack/issues/704).

Question 3 (Should we validate the project.toml license?). At this time due to the passive nature of scheduled kpack's builds we feel like that validating this unused field will lead to unnecessary build failures. 

Question 4 (Should we support setting an alternative path to the descriptor?) Yes, [we plan on providing this and it coinciding it with the release of the next api version.](https://github.com/pivotal/kpack/issues/705)

PR #691, opened 2021-05-10
Use go version 1.15 for CI

samj1912:
The final output images are built using go 1.15. CI should also be using the same version.


matthewmcnew:
Thanks @samj1912!

PR #694, opened 2021-05-12
RFC for sigstore/cosign integration with kpack.

stormqueen1990:
Initial proposal for the `kpack` integration with `cosign`.

[Readable](https://github.com/stormqueen1990/kpack/blob/cosign-integration/rfcs/0000-cosign-integration.md)

cc @rszumlakowski @matthewmcnew


matthewmcnew:
This is great! Thank you!

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,131 @@
...
+  flow must happen after the image has been pushed to the registry.
+
+- If `cosign` fails to sign an image, the build should fail and output an error
+  message specifying the failure.
+
+- Whenever `kpack` signs an image using `cosign`, it should add these
+  annotations:
+  - Build number.
+  - Cluster builder used.
+  - Buildpacks involved in the build process.
```
matthewmcnew, line 73: The buildpacks/stack images used is available on cnb built image labels. The digest of the image will encompass these labels and the digest will be signed by cosign, is is necessary to also include this information in the cosign metadata?

matthewmcnew: cc' @samj1912

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,131 @@
...
+## Actions to take
+
+### Enable image signing using `cosign`
+
+- Create a new configuration in the `Image` resource that allow users to
+  configure image signing with `cosign` as the signing mechanism, specifying
+  the secret reference to get the private key and passphrase. The user can also
+  specify signature annotations as part of configuration:
+```yaml
+cosign:
```
matthewmcnew, line 51: I assume this is a sibling to the notary key?

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,131 @@
...
+`Secret`s for sending private key and passphrase into the build pod:
+```yaml
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cosign-key
+type: Opaque
+data:
+  cosign-privatekey: # base64 encoded private key here
```
matthewmcnew, line 95: These key secret key names are hardcoded? Is there any prior art to this?

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,131 @@
...
+  passphrase: # base64 encoded private key passphrase here
+```
+
+### Authentication to a container registry
+
+`cosign` uses the `DefaultKeychain` from
+[`go-containerregistry`](https://github.com/google/go-containerregistry/blob/main/pkg/authn/README.md#tldr-for-consumers-of-this-package)
+to authenticate the command-line interface using the machine's Docker CLI
+credentials. For `kpack`, it may be possible to use the same authentication
```
matthewmcnew, line 104: kpack makes heavy use of go-containerregistry as well 🙂

rfcs/0000-cosign-integration.md
samj1912, line 51: Apart from having this on the image resource - my use case was to have this on the controller side. 

This was mainly to ensure that *every* image built by kpack is signed with the same cosign key that the cluster administrator controls. Currently this restricts the keys to be present in user namespace. Could we also add a similar field on the controller side?

rfcs/0000-cosign-integration.md
samj1912: I guess we could skip these parts and instead extract them from the image.

rfcs/0000-cosign-integration.md
rszumlakowski: Yes, we thought that the `cosign` key should be a sibling to the `notary` key on the `Image` resource.

rfcs/0000-cosign-integration.md
rszumlakowski: We called the private key name `cosign-privatekey` because the SSH key secret type uses the name `ssh-privatekey` as its key field name.  Cosign has its own key format so we thought it was important to call it out in the key field name.  e.g.: if someone used `ssh-keygen` to make an X509 or ECDSA private key and passed it to `cosign` then `cosign` would fail.  At this time, it only accepts keys made with `cosign generate-key-pair`.

rfcs/0000-cosign-integration.md
rszumlakowski: @matthewmcnew - can you please provide help here?  I haven't seen any mention of configuration for the controller itself anywhere in the `kpack` documentation.  This seems like an implementation detail.

rfcs/0000-cosign-integration.md
stormqueen1990: Would it be OK if I edited this point about metadata to leave only build number and build time?

rfcs/0000-cosign-integration.md
matthewmcnew: Cluster wide image signing is intriguing because it would allow you verify every image produced from a kpack install. However, It is also unprecedented within kpack. The current notary image signing functionality is configured per image. A cluster-wide image signing solution would be a challenge to implement because the build pods may exist in any namespace and would need a mechanism to mount the signing credentials across namespaces.

matthewmcnew: @samj1912 If cosign support was configured per image as specified in this RFC would that still be useful for you? It seems that you could enforce that kpack images are configured with cosign?

rfcs/0000-cosign-integration.md
matthewmcnew: Thanks for the explanation!

rfcs/0000-cosign-integration.md
samj1912: We could yes, but then the cosign secrets would be in each users namespace and they could use that to sign images generated elsewhere.

matthewmcnew: Could you use RBAC to prevent users from reading the cosign secrets?

samj1912: RBAC doesn't allow denying access to certain resources. You will have to instead only allow certain secret object names they they can use which doesn't seem great UX as users couldn't create arbitrary secrets anymore. You would also need a webhook to default these cosign secrets for each image. The other issue would be updating or managing these cosign secrets in each user namespace and syncing them together, which again is a non trivial issue.

samj1912: Another alternative is to create a CRD which is a cluster resource that can store cosign secrets and managing that through RBAC so that only the controller or the pods created by it could use/access it. That might actually be a good way to solve this problem.

samj1912: I would also prefer if the cosign field in the spec was an array as cosign allows multiple private keys to sign the same image. Alternatively, this cluster resource for cosign could be referenced in the cluster builder which will ensure that certain cluster builders if used generate images signed by a particular key.

matthewmcnew: Yeah, RBAC really only works if you prevent secret read.

matthewmcnew: I mentioned this in the OSS CNB slack but, I fear a cluster wide signing solution may be outside the scope of kpack. However, it would definitely be possible with an extension controller that runs in cluster and automatically signs every image that is available on a build.status.latestImage.

rszumlakowski: It sounds like, putting the `cosign` signing on the `Image` (feature 1) resource is helpful, but does not necessarily cover @samj1912 's entire use case (feature 2 - signing at the controller level).  I think we could approach both features separately.  Feature 1 is easier to implement and we could start sooner.  Feature 2 needs more design discussion before implementation could begin.  We could approach it after feature 1 is complete.  Agile programming for the win, right?

Additionally, it seems like a simple modification to make the `Image` resource take a list of `cosign` private keys and sign with each one.  We will make that change to the RFC.

stormqueen1990: Just edited it to accept a list. I believe this should address the multiple key use case.

tylerphelan:
this is amazing!

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,135 @@
...
+  - name: <secret-name n>
+  annotations: # optional map of key-value pairs the user wants to add to images
+    key1: value1
+    key2: value2
+```
+`cosign` will apply this optional map of annotations to the signature payload,
+which can then be verified with `cosign verify`. Verification is out of scope
+for this RFC.
+
+- Implement a flow that generates the signature payload for the image, then
```
tylerphelan, line 65: Does cosign have an sdk we can use for this? Otherwise we could use the binary but that comes with changes to the stacks that builds run with.

tylerphelan:
i recommend changing the `readable` link to use the branch link instead of the blob/commit link

rfcs/0000-cosign-integration.md
stormqueen1990: Would we be able to include their code somehow? I am not a Golang expert, but there's some code here that deals with the payload generation: https://github.com/sigstore/cosign/blob/bcc20b7197999b4a67cfafe68e5797d14531cb95/cmd/cosign/cli/sign.go#L164

stormqueen1990:
> i recommend changing the `readable` link to use the branch link instead of the blob/commit link

Sorry about that! Just updated the link to point to the branch.

rfcs/0000-cosign-integration.md
rszumlakowski: My assumption is that we could simply `import` the `cosign` source code directly into `kpack` as a Golang module.

rfcs/0000-cosign-integration.md
stormqueen1990: I updated this section to leave only build number and build time as required metadata.

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,132 @@
...
+
+This proposal aims to cover specifically the flow of signing an image produced
+by `kpack Image` resource without verification of base or builder images pulled
+in the process.
+
+## Actions to take
+
+### Enable image signing using `cosign`
+
+- Create a new configuration in the `Image` resource that allow users to
```
samj1912, line 46: Should we add an option to set the cosign registry or state that it will be supported?

It is currently controlled via `COSIGN_REPOSITORY`

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,132 @@
...
+  the secret reference to get the private key and passphrase. The user can also
+  specify signature annotations as part of configuration:
+```yaml
+cosign:
+  # a list of secret references holding the private keys and passphrases for signing images
+  secretRefs:
+  - name: <secret-name 1>
+  ...
+  - name: <secret-name n>
+  annotations: # optional map of key-value pairs the user wants to add to images
```
tylerphelan, line 57: would it make more sense for annotations to be configured outside of the image spec such as in a ConfigMap?

1. Allows shared annotations per namespace
2. It might be easier to update a configmap than this field for non-static annotations (ex. git commit)
  - kpack cli doesn't currently support configuring the cosign field

rfcs/0000-cosign-integration.md
stormqueen1990: Makes sense to me. Does a single repository URL configuration address this concern, or should we add it as a part of the secret structure?

rfcs/0000-cosign-integration.md
stormqueen1990: I believe it would be more flexible. In this case, we could have a list of `ConfigMap`s under the `annotations` key.

About `kpack` CLI not currently supporting the `cosign` field, can we create a separate RFC to address this concern?

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,132 @@
...
+
+- Create a new configuration in the `Image` resource that allow users to
+  configure image signing with `cosign` as the signing mechanism, specifying
+  the secret reference to get the private key and passphrase. The user can also
+  specify signature annotations as part of configuration:
+```yaml
+cosign:
+  # a list of secret references holding the private keys and passphrases for signing images
+  secretRefs:
+  - name: <secret-name 1>
```
matthewmcnew, line 54: Could we do without specifying the secrets here and just sign images when they are provided with "cosign" `secrets` on the configured service account? This would allow "cosign" secrets to be configured analogous to the secrets provided for registry and git access. This would make configuring multiple cosigned images per namespace easier?

rfcs/0000-cosign-integration.md
matthewmcnew: How commonly do we think annotations will be used, especially non-static ones?

ncdc: What if we support multiple levels of annotations?
1. Per image - highest priority & overrides any key from a lower priority
2. Per namespace - lower priority

We could implement 1 first, and add 2 later, if it's needed?

How would you imagine the workflow for annotating the git commit?

rfcs/0000-cosign-integration.md
ncdc: @samj1912 how do you envision using `COSIGN_REPOSITORY`?

rfcs/0000-cosign-integration.md
ncdc: > How commonly do we think annotations will be used, especially non-static ones?

I'd probably want to start with some automatically-created dynamic ones, e.g. the git commit/url for the source for the build, information about the build (number, trigger(s), etc.).

I'm not sure about what you might put for static ones?

stormqueen1990: Could the prior suggestion of having the `kpack` build number as an annotation replace the requirement of having git commits as annotations, for instance?

matthewmcnew: > I'd probably want to start with some automatically-created dynamic ones, e.g. the git commit/url for the source for the build, information about the build (number, trigger(s), etc.).

We could pretty easily do this without requiring the user to specify but, is are these git source annotations necessary? That metadata is already on image labels and the image digest (which includes image labels) will be signed by cosign.

matthewmcnew: > Per namespace - lower priority
Is it important to specify how this might be done now at the rfc stage?

ncdc: > We could pretty easily do this without requiring the user to specify but, are these git source annotations necessary? That metadata is already on image labels and the image digest (which includes image labels) will be signed by cosign.

Good point. Maybe we don't need any auto-generated annotations?

> Is it important to specify how this might be done now at the rfc stage?

This feels like the sort of thing to spec out now.... however, I wonder if it makes sense to start with a single thing: only support annotations at the image level? Then we don't have to talk about priorities, and we don't have to talk about global vs local annotations. WDYT?

stormqueen1990: > I'm not sure about what you might put for static ones?

Our thought was that users could want to put annotations with some static information such as a container image name, but it's possible that this is already addressed by other metadata pieces (such as build number).

rfcs/0000-cosign-integration.md
stormqueen1990: Makes sense. In this case, would we need to have a specific secret name to tell us whether `cosign` signing is being requested?

rfcs/0000-cosign-integration.md
samj1912: @ncdc currently my usecase is more around the flexibility of having application tags being in one repository and the cosign tag in another so that it doesn't pollute the original repository with cosign tags and also so that I can control/manage the lifecycle of the repository with cosign signatures  differently than the application one. This also allows me to store signatures of images in an "operator" repository whereas the application images could be saved in a "user" repository.

samj1912: > Makes sense to me. Does a single repository URL configuration address this concern, or should we add it as a part of the secret structure?

Technically cosign allows each signature to be stored in a separate repository. It might be most flexible if we keep it along with the secret structure. But upto you and the maintainers.

stormqueen1990: @matthewmcnew @tylerphelan would it make sense to combine this URL property with the annotations in a `ConfigMap`?

rfcs/0000-cosign-integration.md
rszumlakowski: More arguments why static annotations are useful:

I pictured the signature annotations to be used help an attestation system or as a part of a end users supply chain processes.

1. Someone could put a signature on a container image at build time with an annotation saying, "Built on my build system".
2. Later on they could run integration tests on the container image saying, "Passed integration tests".  Let's say both those processes occur at VMware and they would have the same signing key.  If I remember correctly, you can only sign an image with Cosign more than once if the two signatures have different annotations.
3. Now, let's say the end user has a policy that they have to bless every image that comes into their air-gapped environment.  After image relocation (which would preserve the original keys and their annotations) and some pass through a test environment they would put another signature, with another annotation that says "this image is allowed to be promoted to production".

Now, maybe only the first stage in this example would actually be signed with kpack, but I would like enable the rest of this chain of signatures by using kpack to allow its users to apply signatures with static annotations.

rfcs/0000-cosign-integration.md
stormqueen1990: @ncdc do you think it make more sense to move forward using `ConfigMap`s for the static annotations, taking into account @rszumlakowski's scenario, so we can combine static annotations and `COSIGN_REPOSITORY` URL in the same place?

ncdc: I think there is a valid use case for supporting `spec.cosign.signatureRepository` (`COSIGN_REPOSITORY`).

I also think that because a cosign signature supports annotations, it could make sense to have `spec.cosign.annotations`. (Unless there are good reasons not to support these per-`Image`?)

For static annotations that apply to multiple `Image`s, we can either try to design for that now, or we could defer while we continue to gather more use cases. My preference would be to defer.

rfcs/0000-cosign-integration.md
ncdc: >Someone could put a signature on a container image at build time with an annotation saying, "Built on my build system".

Would it be safer to rely on secure identity built into the systems themselves (@anvega can hopefully provide more detail here) rather than a user-supplied value for a build server's identity?

Also cc @joshuagl in case you've got thoughts 😄

As I wrote in another comment thread, given that signature annotations are a thing, it seems reasonable to support them at the Image level, unless folks have significant reasons not to.

lukehinds:
really nice to see this underway..tagging a fellow sigstorista @dlorenc

rfcs/0000-cosign-integration.md
dlorenc: cosign author here! I put the "annotations" into the tool to basically act as "lightweight" attestations. There are some other richer formats we're looking to support down the road (mostly the in-toto Attestation format: https://github.com/in-toto/attestation) but nothing is concrete yet.

I hope that helps!

anvega: Aside from the constraint pointed out of annotations meant to support lightweight attestations, secure identities are inherently short-lived. As such, there'd be limitations around how to validate those without additional moving parts or a broader framework. 

That said, consideration for secure identities is directionally correct. With support for in-toto down the line, it does open up the possibilities to integrate in-toto attestations with secure identities.

rfcs/0000-cosign-integration.md
samj1912: @matthewmcnew, @dlorenc and I were discussing a possible implementation that doesn't require cross namespace secret mounts.

See this thread - https://sigstore.slack.com/archives/C01PZKDL4DP/p1623002688061600?thread_ts=1623002688.061600&cid=C01PZKDL4DP

The gist of it is -

A pod runs in user namespace and outputs the image ref and digest (so that the user doesn't have access to the signing key)
The controller can take just that information and generate the signature and upload it to a separate repository.

This way, no secrets from the user are needed or used.

samj1912: Maybe something to discuss in the WG meeting tomorrow?

stormqueen1990: Sure thing!

rszumlakowski: We should discuss if it makes sense to allow both the `notary` and `cosign` keys in the same `Image` resource at the same time.  There is no technical reason to block anyone from using both tools to sign the same image but it be an unusual use case. (as discussed during the kpack community meeting on June 7, 2021)

stormqueen1990:
Updated the RFC file to incorporate feedback received.

rfcs/0000-cosign-integration.md
stormqueen1990: RFC updated to address this and explicitly state that Notary v1 configurations should not be affected by Cosign configurations.

rfcs/0000-cosign-integration.md
stormqueen1990: Updated the RFC to state that `kpack` will read the secret types to determine whether `cosign` should be used as the signing mechanism.

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,143 @@
...
+in the process.
+
+## Actions to take
+
+### Enable image signing using `cosign`
+
+- Implement a flow that generates the signature payload for the image, then
+  calculates its signature and pushes it either to the registry where the image
+  is located, using the same credentials that were used to push the image, or
+  [to the registry specified in the `COSIGN_REPOSITORY` environment variable](#key-generation-and-storage).
```
matthewmcnew, line 46: How would the `COSIGN_REPOSITORY` environment variable be set?

stormqueen1990: This would be an implementation detail. We could mount the secret into the build pod and consume it from there, or use other mechanisms to pass it in, depending on the preferred approach. `cosign` currently consumes it from the environment variable. However, I assume that there might be a flow where we don't necessarily have it set in the environment.

ncdc: Given the current code in cosign, which only supports this environment variable, kpack's code would need to set it before invoking the code that uploads signatures.

ncdc: The user will need to be able to specify the value for this. I think we previously discussed putting it in the Image's `cosign` section as a field?

stormqueen1990: > The user will need to be able to specify the value for this. I think we previously discussed putting it in the Image's `cosign` section as a field?

We indeed talked about it, but given that there might be a situation where a user could want to push different signatures to different repositories, this field was added to the secret instead. I linked the section where we are talking about it. Happy to reorder sections if it helps reading! 🙂

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,143 @@
...
+`kpack` will be able to determine that the user wants to sign images using
+`cosign` by scanning the types of the secrets attached to the service account,
+which in this use case would be a custom type:
+```yaml
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cosign-key
+type: kpack.io/cosign-credentials
```
stormqueen1990, line 87: With regards to secret types, this document states that any string is accepted as a secret type: https://kubernetes.io/docs/concepts/configuration/secret/#secret-types

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,143 @@
...
+stringData:
+  cosign-privatekey: <cosign-private-key>
+  passphrase: <cosign-private-key-passphrase>
+  # Optional repository URL, to be used as the environment variable
+  # COSIGN_REPOSITORY. If not set, pushes the signature collocated with
+  # the image.
+  cosign-repository-url: <repository-url>
```
samj1912, line 94: I think one more thing we might need to account for is the cosign variable `COSIGN_DOCKER_MEDIA_TYPES`
See https://github.com/sigstore/cosign#registry-support

There are some registries that work only with the variable set and some that work only without the variable. This should be exposed in the config somehow.

cc: @dlorenc who might know more about compat issues with registries.

dlorenc: Yep, that one would be nice to have but probably not critical. It's rare to see issues and we don't have a perfect understanding across registries unfortunately.

samj1912: I have sadly faced issues without that flag in certain private registries. It would be great if this was included.

samj1912:
Also relevant to the secret storage format https://github.com/sigstore/cosign/pull/345

I wonder if we should add a PR upstream to add the repository and docker media types fields if we want to consolidate the secret storage format and make it super easy for kpack users to use cosign?

dlorenc:
> I wonder if we should add a PR upstream to add the repository and docker media types fields if we want to consolidate the secret storage format and make it super easy for kpack users to use cosign?

We're not really opinionated on the secret storage format itself, we just wanted a secure way to get a usable secret into k8s somehow :) If there's anything we could change about the naming to make it easier for kpack please let us know! We were running into this type of thing over and over again: https://smallstep.com/blog/command-line-secrets/

I'm not super sold on plumbing the rest of the config env vars up into k8s through this path yet, what would it look like for a kpack/cosign user?

rfcs/0000-cosign-integration.md
tylerphelan, line 94: should this be `cosign-repository` to match the env var `COSIGN_REPOSITORY`

matthewmcnew: Is there any risk in configuring this on the secret directly?

stormqueen1990, line 94: @tylerphelan @matthewmcnew would there be any concerns from the `kpack` side if we changed the keys in this secret to match the keys generated by `cosign generate-key-pair -k8s <namespace>/<secret-name>`, as shown [here](https://github.com/sigstore/cosign/blob/9de98f9dca95acdb4f0a193d0305cd8a977f0dd1/pkg/cosign/kubernetes/secret.go#L79)?
This way we would be able to reuse the secret generated by `cosign` as suggested by @samj1912.

stormqueen1990: In case we decide to go down the route of using the same names as generated by `cosign`, would it be fair to use `cosign.repository` as the key here, to maintain consistency, instead of using dashes as a separator?

samj1912: I think it might be better to move the repository URL and the docker media types which I suggested above to be annotations on the secret. That way, by default users can use the `cosign generate-key-pair -k8s` and if needed they could add these annotations later to customize the behavior. (via `kubectl annotate` as suggested by @ncdc )

stormqueen1990: I'll update the example here to reflect these.

tomkennedy513: Ya I like that idea

rszumlakowski: If we do add this option I think it would make the most sense to add it to the secret directly since that's where the repository URL is.  The option is there to support a registry so the option should be located alongside the registry location you want to use.

One risk I see to putting this option in the secret is now that the secret has fields in it that are not really a secret; it's polluting a secret with non-secret things.  I'm not sure if this is idiomatic Kubernetes or not.

rszumlakowski: Using annotations for the secret metadata (repository URL, Docker media types) sounds super awesome.

stormqueen1990: I added this field to the proposal as an annotation, as discussed yesterday. 🙂

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,162 @@
...
+These secrets should follow the format generated by
+[`cosign generate-key-pair -k8s <namespace>/<name>`](https://github.com/sigstore/cosign/pull/345).
+
+The service account can have more than one `cosign` secret attached to it. For
+each `cosign` secret attached, a separate signature should be generated and
+stored.
+
+`kpack` will be able to determine that the user wants to sign images using
+`cosign` by scanning the annotations in each of the secrets. Secrets to be used
+by `cosign` signing should be annotated with `kpack.io/cosign-credentials`.
```
matthewmcnew, line 83: Is the `kpack.io/cosign-credentials` annotation necessary can kpack just detect the `cosign.*` keys?

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,162 @@
...
+
+`cosign` can use file-based keys, key management systems or hardware keys to
+ingest private keys, as well as a keyless flow that requires integration with
+[Rekor](https://github.com/sigstore/rekor).
+
+`kpack` should not generate the keys used for signing and verification. The user
+should pass them in using one of the mechanisms supported by `cosign`.
+In this RFC, we suggest that the first mechanism implemented be the use of
+Kubernetes `Secret`s for sending the relevant data into the build pod, by means
+of attaching secrets to the service account used by `kpack`.
```
matthewmcnew, line 73: Could we clarify exactly how these secrets are used? Any build that's service account has the cosign secrets will be signed? What about builders?

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,162 @@
...
+- [Issue](https://github.com/pivotal/kpack/issues/684) filed on May 4th.
+- [Use of `cosign` in Kubernetes](https://github.com/kubernetes/release/pull/2016)
+  for validation of distroless images.
+- [Use of `cosign` in Connaisseur](https://github.com/sse-secure-systems/connaisseur/pull/107)
+  for policy enforcement in container images.
+- [Notary v1 implementation pull request](https://github.com/pivotal/kpack/pull/541).
+
+## Alternatives
+
+- Use a CI/CD step to sign images using `cosign` instead of implementing it as
```
matthewmcnew, line 155: Could we mention the alternative of configuring the cosign configuration on the image spec?

stormqueen1990: Sure, I'll add it here!

rfcs/0000-cosign-integration.md
stormqueen1990: That would work, too. I originally thought of adding the specific annotation to prevent accidental signing, as a mean of making intention clear.

rfcs/0000-cosign-integration.md
stormqueen1990: Yes, the proposal would be that any build that has the service account with such a secret attached would get signed. Our original idea was to leave out builders and sign images only. I'll update the text to make these parts clearer.

matthewmcnew: I am a bit concerned that would require creating secrets with `cosign` and then altering them to support use in kpack. Would you object to being permissive and signing with any "cosign" secret?

rfcs/0000-cosign-integration.md
matthewmcnew: I am a bit concerned that would require creating secrets with cosign and then altering them to support use in kpack. Would you object to being permissive and signing with any "cosign" secret?

rfcs/0000-cosign-integration.md
matthewmcnew: I think we should also sign builders if we are using the same secret inside service account mechanism for consistency.

rfcs/0000-cosign-integration.md
stormqueen1990: I see. That's fair, I can update it to remove the requirement of having an empty annotation here. I also just recalled that my original rationale was that the user would need to manually update the secret in any case, in order to add the `COSIGN_REPOSITORY` and `COSIGN_DOCKER_MEDIA_TYPES` variables, but this is only true when the signature is being pushed to a different repository or legacy media types compatibility is required.

rfcs/0000-cosign-integration.md
matthewmcnew: Could you clarify this is in the RFC?

stormqueen1990: Sure thing! 😄

rfcs/0000-cosign-integration.md
stormqueen1990: I believe this is being addressed on #765. Is it OK if I resolve this conversation?

rfcs/0000-cosign-integration.md
stormqueen1990: I added a sentence in this section calling out the fact that any build that has a service account with cosign secrets attached to it will have signatures generated.

rszumlakowski: Although I agree strongly that kpack should sign the builder images that it produces I also feel strongly that signing and verifying builder images is a second feature.

Signing end-user application images is the base feature that we have been discussing in this RFC so far.  I think that's the first feature we should implement.  kpack does not run these images so it does not need to be responsible for verifying signatures on them.  This RFC already covers this feature.

Signing builder images is a feature that I think we should approach after the base feature (signing application images) is complete.  This RFC does not yet discuss verifying signatures on any images.  It feels incomplete to me if one part of kpack is signing builder images that another part of kpack consumes, but does not verify the signatures on the builder images.

Can we please defer the signing and verifying of builder images to a second RFC?

tylerphelan (approved):

tomkennedy513 (approved):

rfcs/0000-cosign-integration.md
matthewmcnew: Implementing builder signing can definitely be implemented independently of individual image signing (and perhaps even released in different release). 

However, I believe it would still be valuable to specify how builder signing would work in this rfc. A benefit of the rfc process is that it enables us to design and evolve the api holistically at a higher level than individual features.  In this case particularly, allowing builders to be signed with the same mechanism as images is a big benefit of the proposed approach. It seems like a missed opportunity if we didn't provide a specify a consistent image signing approach for all the images that kpack produces.

dumez-k (approved):

samj1912 (approved):

rfcs/0000-cosign-integration.md
matthewmcnew: Builds scheduled with a kpack built builder image will be scheduled with the exact digest that kpack itself is reporting so, verifying them within kpack might not be necessary. However, A builder image may be used outside of kpack and that is where image signing seems valuable to me. If we want to verify builder images within kpack it may influence this design so, it might be useful to discuss.

stormqueen1990: I updated the text to include signing of the builder image!

tomkennedy513 (approved):

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,167 @@
...
+involvement recently.
+
+## Outcome
+
+`kpack` integrates with `cosign` to sign images it builds and push the
+signatures to a registry so that users can ensure the chain of custody of a
+generated artifact.
+
+This proposal aims to cover specifically the flow of signing an image produced
+by the `kpack Image` resource, along with its builder image, without
```
matthewmcnew, line 36:
```suggestion
by the kpack `Image`, `Build`, `Builder` and `ClusterBuilder` resources, without
```
Because the the Builder resources are unique resources that may not be used with the Image resource.

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Actions to take
+
+### Enable image signing using `cosign`
+
+- Implement a flow that generates the signature payload for the image, then
+  calculates its signature and pushes it either to the registry where the image
+  is located, using the same credentials that were used to push the image, or
+  [to the registry specified in the `COSIGN_REPOSITORY` environment variable](#key-generation-and-storage).
+  `kpack` should sign both the user-requested image being generated and the
```
matthewmcnew, line 47:
```suggestion
  `kpack` should sign images produced by kpack builds and images produced by the Builder and ClusterBuilder resources, using their respective service accounts to
```

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,167 @@
...
+  builder image used for the build, using their respective service accounts to
+  find credentials. This flow must happen after each of the images have been
+  pushed to the registry.
+
+- If `cosign` fails to sign an image, the build should fail and output an error
+  message in the build log, so the operator can troubleshoot the issue. The
+  error messages should also be added in any other places where error messages
+  are presented.
+
```
matthewmcnew, line 56:
```suggestion
- If `cosign` fails to sign an image produced by a Builder resource the Builder should not enter a Ready status. 
```

rfcs/0000-cosign-integration.md
```diff
@@ -0,0 +1,167 @@
...
+- Whenever `kpack` signs an image using `cosign`, it should add these
```
matthewmcnew, line 57:
```suggestion
- Whenever `kpack` signs an image produced within a kpack build using `cosign`, it should add these
```

rfcs/0000-cosign-integration.md
stormqueen1990: @matthewmcnew Makes sense! Is it OK if I keep the preformatting for the `kpack` word?

matthewmcnew (approved):

samj1912 (approved):

tylerphelan (approved):

dumez-k (approved):

matthewmcnew:
Thanks! @stormqueen1990

Issue #704, opened 2021-05-25
Log info when `build.buildpacks` or `build.builder` is ignored in project.toml


matthewmcnew:
https://github.com/pivotal/kpack/pull/690 adds support for project.toml's env and include/exclude. However, `build.buildpacks` and `build.builder` are not used and this may be confusing if users expect them to behave identically to pack. 

We should log a simple info message during the `prepare` step that we are ignoring these keys if they are populated. 

Note: `pack` will ignore these project.toml properties if a builder or buildpack order is explicitly provided. All kpack images/builds are configured with a builder. However, external tooling like the `kp` cli may default a builder so, a configured kpack builder is not necessarily the equivalent of an explicitly configured pack builder.

PR #711, opened 2021-06-02
Log info when ignoring buildpacks or builder in descriptor

dumez-k:


pkg/cnb/project_descriptor.go
```diff
@@ -23,6 +24,13 @@ func ProcessProjectDescriptor(appDir, platformDir string) error {
 	if err != nil {
 		return err
 	}
+	if d.Build.Buildpacks != nil {
+		logger.Println("warning: buildpacks provided in project descriptor file will be ignored")
```
matthewmcnew, line 28: I think @samj1912 was suggesting `info` as a message because it is may not be a `warning` situation?

matthewmcnew:
@samj1912 It would be great if you could review this PR 🙂

pkg/cnb/project_descriptor.go
dumez-k: I guess my only concern was elevating the message to the user because this may be unexpected behavior, i.e. if they explicitly provided builder/buildpacks in the descriptor expecting them to be consumed I wanted to call it out a little more aggressively.

But you're right it would still probably be an INFO level message not a WARN.

pkg/cnb/project_descriptor.go
```diff
@@ -77,10 +85,18 @@ func getFileFilter(d descriptor) (func(string) bool, error) {
 	return nil, nil
 }
 
+type buildpack struct {
+	Id      string `json:"id" toml:"id"`
+	Version string `json:"version" toml:"version"`
+	Uri     string `json:"uri" uri:"uri"`
```
tylerphelan, line 91: `uri:"uri"` -> `toml:"uri"` ?

pkg/cnb/project_descriptor_test.go
```diff
@@ -18,16 +20,20 @@ func TestProcessProjectDescriptor(t *testing.T) {
...
 	var (
 		appDir, platformDir, projectToml string
 	)
 
 	it.Before(func() {
 		var err error
+		buf = new(bytes.Buffer)
+		logger = log.New(buf, "", 0)
 		appDir, err = ioutil.TempDir("", "appDir")
-		require.NoError(t, err)
```
tylerphelan, line 28: I think `NoError` reads better but it's not a big deal

tylerphelan (approved):

tomkennedy513 (approved):

pkg/cnb/project_descriptor.go
samj1912: We just had a discussion around project descriptor portability across different platforms in the CNB WG meeting today. I think we (CNB) want to define some sort of guidance for what platforms are expected to do when they find project descriptor or features from it they can't/don't want to support.

Suggestions in the CNB WG meeting ranged from straight up build errors for unsupported features or warnings. 
---

The reason I had suggested info as the log level instead of warning is because in the pack world, flags that are explicitly passed in through the CLI override the project descriptor and I believe that the resource definition is the closest equivalent to explicit flags here in the kpack world. If the user explicitly passed in a builder, that should override the builder in the project toml.

Looking back at it, I think we may want to warn for the buildpack key but keep it to an info level for the builder key?

Thoughts?

samj1912 (approved):

pkg/cnb/project_descriptor.go
dumez-k: oh woops good catch lol

pkg/cnb/project_descriptor.go
dumez-k: The resource definition explicit flag equivalence makes sense to me. I personally don't love differing the levels of the two log messages though. Would rather keep them in sync because it could give users the wrong impression of severity, like buildpacks being ignored is somehow more severe than builders being ignored and neither are *actually* problems technically.

That being said would would be happy to drop both to `info` level if that is good with you!

pkg/cnb/project_descriptor_test.go
dumez-k: makes sense

dumez-k:
Resolves https://github.com/pivotal/kpack/issues/704

PR #713, opened 2021-06-03
decompose k8ssecret keychain interface

tomkennedy513:
- rename methods to be easier to understand
- enable ability to get service account secrets only, ignoring volume secrets

Co-authored-by: Tyler Phelan <tylerp@vmware.com>

### DIFF
--- a/pkg/dockercreds/k8sdockercreds/k8s_keychain.go
+++ b/pkg/dockercreds/k8sdockercreds/k8s_keychain.go
@@ -32,7 +32,7 @@ func NewSecretKeychainFactory(client k8sclient.Interface) (*k8sSecretKeychainFac
 	return &k8sSecretKeychainFactory{client: client, volumeKeychain: volumeKeychain}, nil
 }
 
-func (f *k8sSecretKeychainFactory) KeychainForSecretRef(ctx context.Context, ref registry.SecretRef) (authn.Keychain, error) {
+func (f *k8sSecretKeychainFactory) MultiKeychainFromServiceAccountRef(ctx context.Context, ref registry.ServiceAccountRef) (authn.Keychain, error) {
 	if !ref.IsNamespaced() {
 		keychain, err := k8schain.New(ctx, nil, k8schain.Options{})
 		if err != nil {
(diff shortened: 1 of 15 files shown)


pkg/dockercreds/k8sdockercreds/k8s_keychain.go
```diff
@@ -41,35 +41,24 @@ func (f *k8sSecretKeychainFactory) KeychainForSecretRef(ctx context.Context, ref
...
-func toStringPullSecrets(secrets []corev1.LocalObjectReference) []string {
-	var stringSecrets []string
-	for _, s := range secrets {
-		stringSecrets = append(stringSecrets, s.Name)
-	}
-	return stringSecrets
-}
-
-func keychainFromServiceAccount(ctx context.Context, secretRef registry.SecretRef, fetcher *secret.Fetcher) (authn.Keychain, error) {
+func (f *k8sSecretKeychainFactory) KeychainFromServiceAccountSecrets(ctx context.Context, ref registry.ServiceAccountRef) (authn.Keychain, error) {
```
matthewmcnew, line 57: It looks like we are exposing a public api here to avoid the k8schain which I fear might be exposing too much implementation detail. Could we instead provide a keychain that treats the k8schain as the lowest priority?

pkg/dockercreds/k8sdockercreds/k8s_keychain.go
```diff
@@ -32,7 +32,7 @@ func NewSecretKeychainFactory(client k8sclient.Interface) (*k8sSecretKeychainFac
 	return &k8sSecretKeychainFactory{client: client, volumeKeychain: volumeKeychain}, nil
 }
 
-func (f *k8sSecretKeychainFactory) KeychainForSecretRef(ctx context.Context, ref registry.SecretRef) (authn.Keychain, error) {
+func (f *k8sSecretKeychainFactory) MultiKeychainFromServiceAccountRef(ctx context.Context, ref registry.ServiceAccountRef) (authn.Keychain, error) {
```
matthewmcnew, line 35: Is `multi` necessary in the name?

pkg/registry/keychain_factory.go
```diff
@@ -7,23 +7,23 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
-type SecretRef struct {
+type ServiceAccountRef struct {
```
matthewmcnew, line 10: This is super minor but, I still feel like it is a `SecretRef` because that is the ultimate thing that is getting pointed to.

PR #724, opened 2021-06-11
Use until for image waiting

tylerphelan:
This includes the change to use UntilWithSync for build

### DIFF
--- a/pkg/logs/wait_for_image.go
+++ b/pkg/logs/wait_for_image.go
@@ -8,6 +8,7 @@ import (
 	"github.com/pkg/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/cache"
 	watchTools "k8s.io/client-go/tools/watch"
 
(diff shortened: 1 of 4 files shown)


pkg/logs/wait_for_image.go
```diff
@@ -74,10 +59,9 @@ func imageUpdateHasResolved(generation int64) func(event watch.Event) (bool, err
 			return false, errors.New("unexpected object received")
 		}
 
-		//space shuttle style
-		if image.Status.ObservedGeneration == generation {
-			if !image.Status.GetCondition(corev1alpha1.ConditionReady).IsUnknown() {
-				return true, nil // Ready=False or Ready=True
+		if image.Status.ObservedGeneration == generation { // image updated
+			if condition := image.Status.GetCondition(corev1alpha1.ConditionReady); condition.IsFalse() {
```
matthewmcnew, line 63: Why restrict this to image failed?

matthewmcnew: I think it will skip no-op changes if we ignore Ready=True here.

matthewmcnew (approved):

PR #737, opened 2021-06-17
Remove lifecycle building binary

tylerphelan:
- Lifecycle is built out of band of the kpack repo now


tomkennedy513 (approved):

matthewmcnew:
I think it is still useful to include how the lifecycle image is built within the kpack repo so that the kpack project can be completely built on its own. @tylerphelan I believe we were looking at utilizing the kpack repo to build the lifecycle image in the kpack pipelines? Could we close this?

PR #755, opened 2021-07-03
Image spec add new tagging strategy: GitRevision

zhengzhou1992:
Hi,

It may be a common usage that associate the git revision with the container image revision, so I extend the ImageTaggingStrategy with the **GitRevision** option. 

for example, use git with revision `ewfge4w`  as Image source,
when use BuildNumber image tagging strategy:
`docker.io/my-project/myapp:mytag-b1.20210703.151213`
when use **GitRevision** image tagging strategy:
`docker.io/my-project/myapp:mytag-ewfge4w`

Please let me know if any suggestions and comments, I will try to adjust  : 0

### DIFF
--- a/pkg/apis/build/v1alpha1/image_builds.go
+++ b/pkg/apis/build/v1alpha1/image_builds.go
@@ -158,25 +158,43 @@ func (im *Image) SourceResolver() *SourceResolver {
 }
 
 func (im *Image) generateTags(buildNumber string) []string {
-	if im.disableAdditionalImageNames() {
+	tag, err := name.NewTag(im.Spec.Tag, name.WeakValidation)
+
+	switch im.Spec.ImageTaggingStrategy {
+	case BuildNumber:
+		if err != nil {
+			// We assume that if the Image Name cannot be parsed the image will not be successfully built
+			// in this case we can just ignore any additional image names
+			return nil
+		}
+
+		now := time.Now()
+		tagName := tag.TagStr() + "-"
+		if tagName == "latest-" {
+			tagName = ""
+		}
+		return []string{
+			im.Spec.Tag,
+			tag.RegistryStr() + "/" + tag.RepositoryStr() + ":" + tagName + "b" + buildNumber + "." + now.Format("20060102") + "." + fmt.Sprintf("%02d%02d%02d", now.Hour(), now.Minute(), now.Second())}
+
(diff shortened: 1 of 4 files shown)


matthewmcnew:
Hi @zhengzhou1992, Thanks for the PR!

Utilizing the git commit for an additional image seems to be useful functionality but, I fear we may be conflating the objective of the build number tag with the desire for meaningful tags. 

The original objective of the the `BuildNumber` tag was to ensure that each built image was tagged to prevent images built from previous builds from being garbage collected. A single git commit could be built multiple times so, a git commit tag would not necessarily prevent images from being garbage collected. 

I wonder if we should simply provide the git commit as a tag in addition to the build number tag or allow multiple tags to be provided in the `image.spec.tag`?

Curious other's thoughts.

matthewmcnew:
Hi @zhengzhou1992,
Thanks for the context! It sounds like additionally tagging all builds with the git revision alongside the build number strategy would be helpful and/or allowing you provide additional tags that are mutable. 

Do you have any automation that runs after kpack builds a new image? Or do developers manually deploy kpack built images? We commonly recommend adding additional tags after kpack builds an image in a CI/CD tool perhaps with the [crane cli](https://github.com/google/go-containerregistry/blob/main/cmd/crane/README.md).

v-zhidu:
I have similar problems in my company. A readable git revison tag would be helpful for some workflows. We should providing multiple strategies for users to choose is not really bad idea. 

> We commonly recommend adding additional tags after kpack builds an image in a CI/CD tool perhaps with the crane cli.
> #802 

There two methods are not suitable for our scenario. I intend to automate the process of building images in one tool. I prefer manage some inmutable configuration, and determine tags when kpack building.

samj1912:
@v-zhidu kpack now natively supports additional tags - https://github.com/pivotal/kpack/pull/802

v-zhidu:
> 
@samj1912 thanks for the reply. I have seen this PR. But the additional tags are reassigned each time the code is committed in my scenario. I prefer the image resource is immutable. As far as I know，spring-boot-maven-plugin is already integrated with buildpacks, and it can determine the image tag by version in pom.xml (same as git tag).

matthewmcnew:
Closing this as it needs more discussion. Could we open an issue or rfc to track the discussion on this going forward?

PR #763, opened 2021-07-09
Fix changed pack command

wiggitywhitney:
Hello!  The pack command `pack suggest-stacks` has been deprecated and replaced with `pack stack suggest` so I have updated the kpack docs accordingly.  I hope you find this helpful!  

Have a wonderful day.


matthewmcnew:
Good Catch! Thanks, @wiggitywhitney

matthewmcnew (approved):

PR #768, opened 2021-07-12
Unfocus Access Checker Tests and update error messages to match

jromero:
I came across this focused test. Once I unfocused it they appear to have been some outdated tests. I updated the expected error messages to match. 

_Full disclosure: I didn't dig into why the messages changed but I assumed that it might simply be a client library update._


tomkennedy513 (approved):
Nice find!

matthewmcnew (approved):

Issue #741, opened 2021-06-22
Support buildpacks with the wildcard stack


matthewmcnew:
Currently, buildpacks with a wildcard stack will not be properly recognized by kpack. 

RFC: https://github.com/buildpacks/rfcs/pull/97

mgibson1121:
Is there any way to validate this?  Correct me if I'm wrong but seems like it could be validated by the UBI stories when they are finished.

matthewmcnew:
This can be validated with any of the heroku jvm buildpacks: https://github.com/heroku/buildpacks-jvm/releases.

matthewmcnew:
You can also package a simple buildpack locally with the wildcard stack with `pack buildpack new` & edit buildpack.toml & `pack buildpack package`

PR #776, opened 2021-07-19
add support for anystack buildpacks

tomkennedy513:


tylerphelan (approved):

pkg/cnb/buildpack_validation.go
```diff
@@ -16,7 +17,9 @@ func (bl BuildpackLayerInfo) supports(buildpackApis []string, id string, mixins
 	}
 
 	for _, s := range bl.Stacks {
-		if s.ID == id {
+		iaAnystack := s.ID == "*" && api.MustParse(bl.API).Compare(api.MustParse("0.5")) >= 0
```
matthewmcnew, line 20: MustParse will panic on a not parseable input which is actually an underlying user input. `api.NewVersion` is probably safer.

matthewmcnew: We already pull in semver to parse buildpack versions. Could we use that instead of adding another dependency on the lifecycle?

tomkennedy513: ya good call

matthewmcnew (approved):

pkg/cnb/buildpack_validation.go
```diff
@@ -15,8 +16,19 @@ func (bl BuildpackLayerInfo) supports(buildpackApis []string, id string, mixins
 		return errors.Errorf("unsupported buildpack api: %s, expecting: %s", bl.API, strings.Join(buildpackApis, ", "))
 	}
 
+	anyStackMinimumVersion, err := semver.NewVersion("0.5")
```
matthewmcnew, line 19: You could MustParse this.

matthewmcnew: And pull it out as var

pkg/cnb/buildpack_validation.go
```diff
@@ -46,3 +54,7 @@ func present(haystack []string, needle string) bool {
 	}
 	return false
 }
+
+func isAnystack(stackId string, buildpackVersion *semver.Version) bool {
```
matthewmcnew, line 58: Nice!

matthewmcnew:
Closes https://github.com/pivotal/kpack/issues/741

PR #780, opened 2021-07-21
Utilize release-1.19 of kube-openapi to generate openapi definitions

matthewmcnew:
- This makes the openapi definition consistent across v1alpha1 and v1alpha2
 - Fix pipeline check that verifies ./hack/openapi-codegen.sh is consistent with committed definition


dumez-k (approved):
nice!

tomkennedy513 (approved):

PR #783, opened 2021-07-23
Migrate webhook configs to v1 api version

tylerphelan:
- v1beta1 will be removed in k8s 1.22

I opted to explicitly set all values that have new defaults to their previous values so nothing should change.

I added the v1 admissionreview version to webhook configs.

### DIFF
--- a/config/webhook.yaml
+++ b/config/webhook.yaml
@@ -6,7 +6,6 @@ metadata:
 webhooks:
 - admissionReviewVersions:
   - v1
-  - v1beta1
   clientConfig:
     service:
       name: kpack-webhook
@@ -14,7 +13,6 @@ webhooks:
   failurePolicy: Fail
   matchPolicy: Exact
   sideEffects: None
-  timeoutSeconds: 30
   name: defaults.webhook.kpack.io
 ---
 apiVersion: admissionregistration.k8s.io/v1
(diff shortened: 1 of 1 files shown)


tomkennedy513 (approved):

dumez-k (approved):

config/webhook.yaml
```diff
@@ -1,32 +1,38 @@
 #@ load("@ytt:data", "data")
-apiVersion: admissionregistration.k8s.io/v1beta1
+apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
   name: defaults.webhook.kpack.io
 webhooks:
 - admissionReviewVersions:
+  - v1
```
matthewmcnew, line 8: Any reason to have v1beta1 now?

config/webhook.yaml
```diff
@@ -1,32 +1,38 @@
...
   - v1beta1
   clientConfig:
     service:
       name: kpack-webhook
       namespace: kpack
   failurePolicy: Fail
+  matchPolicy: Exact
   sideEffects: None
+  timeoutSeconds: 30
```
matthewmcnew, line 17: Seems like if the new default is `10` and other projects are going with that: https://github.com/tektoncd/pipeline/blob/main/config/500-webhooks.yaml we should maybe just take the default?

config/webhook.yaml
tylerphelan: i dont think we need it actually

PR #795, opened 2021-08-11
Validate that image cache size does not change when the storage class is immutable

### LINKED ISSUE #775: kpack reconciliation loop gets stuck if storage class does not allow dynamic resizing
If the storage class does not allow resizing (both up and down) -- kpack images currently get stuck and don't trigger builds on source changes etc.

We could optionally check for an annotation on the stroageclass `is-mutable` or `is-resizeable` to allow the webhook to prevent users for re-sizing the images.

samj1912:
Fixes #775

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -104,6 +104,9 @@ func withCheckDefaultStorageClass(ctx context.Context, storageClassLister lister
 
 		if val, ok := sc.Annotations["storageclass.kubernetes.io/is-default-class"]; ok && val == "true" {
 			ctx = context.WithValue(ctx, buildapi.HasDefaultStorageClass, true)
+			if val, ok := sc.Annotations["storageclass.kubernetes.io/is-immutable"]; ok && val == "true" {
+				ctx = context.WithValue(ctx, buildapi.IsImmutable, true)
+			}
 			break
 		}
 	}
--- a/pkg/apis/build/v1alpha1/image_validation.go
+++ b/pkg/apis/build/v1alpha1/image_validation.go
@@ -16,6 +16,7 @@ type ImageContextKey string
 
 const (
 	HasDefaultStorageClass ImageContextKey = "hasDefaultStorageClass"
+	IsImmutable            ImageContextKey = "isImmutable"
 )
 
 var (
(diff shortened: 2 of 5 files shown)


cmd/webhook/main.go
```diff
@@ -104,6 +104,9 @@ func withCheckDefaultStorageClass(ctx context.Context, storageClassLister lister
 
 		if val, ok := sc.Annotations["storageclass.kubernetes.io/is-default-class"]; ok && val == "true" {
 			ctx = context.WithValue(ctx, buildapi.HasDefaultStorageClass, true)
+			if val, ok := sc.Annotations["storageclass.kubernetes.io/is-immutable"]; ok && val == "true" {
```
samj1912, line 107: Happy to change this annotation to whatever others feel best. Maybe kpack io domain might be preferable.

phil9909: Maybe we could check `sc.AllowVolumeExpansion` instead of inventing an annotation. See https://kubernetes.io/docs/concepts/storage/storage-classes/#allow-volume-expansion

samj1912: @phil9909 thanks for the suggestion! much cleaner!

pkg/apis/build/v1alpha1/image_validation.go
```diff
@@ -16,6 +16,7 @@ type ImageContextKey string
 
 const (
 	HasDefaultStorageClass ImageContextKey = "hasDefaultStorageClass"
+	IsExpandable           ImageContextKey = "IsExpandable"
```
tylerphelan, line 19: `isExpandable` for consistency?

pkg/apis/build/v1alpha1/image_validation.go
```diff
@@ -96,11 +97,19 @@ func (is *ImageSpec) validateCacheSize(ctx context.Context) *apis.FieldError {
 
 	if apis.IsInUpdate(ctx) {
 		original := apis.GetBaseline(ctx).(*Image)
-		if original.Spec.CacheSize != nil && is.CacheSize.Cmp(*original.Spec.CacheSize) < 0 {
-			return &apis.FieldError{
-				Message: "Field cannot be decreased",
-				Paths:   []string{"cacheSize"},
-				Details: fmt.Sprintf("current: %v, requested: %v", original.Spec.CacheSize, is.CacheSize),
+		if original.Spec.CacheSize != nil {
+			if ctx.Value(IsExpandable) == false && is.CacheSize.Cmp(*original.Spec.CacheSize) != 0 {
```
tylerphelan, line 101: I think checking first if the cache size is decreased and throw "Field cannot be decreased" then checking for expansion is more precise error handling

pkg/apis/build/v1alpha1/image_validation.go
```diff
@@ -96,11 +97,19 @@ func (is *ImageSpec) validateCacheSize(ctx context.Context) *apis.FieldError {
...
+				return &apis.FieldError{
+					Message: "Field cannot be changed",
```
tylerphelan, line 103: would some context to the error be valuable? -> "Field cannot be changed, storageClass cannot be expanded"

tylerphelan (changes requested):
nice – some small comments.

PR #796, opened 2021-08-11
Add support for platform API 0.6

### LINKED ISSUE #767: Support platform api 0.6
The lifecycle now supports the [0.6 platform api](https://github.com/buildpacks/spec/releases/tag/platform%2Fv0.6). When it is  available kpack should utilize it. 

There are some notable improvements which kpack should utilize: 
- The  image manifest size in the export report should be used by notary image signing in the completion container.   
- kpack no longer needs to specify the default process type is "web"

samj1912:
Fixes #767

### DIFF
--- a/pkg/apis/build/v1alpha1/build_pod.go
+++ b/pkg/apis/build/v1alpha1/build_pod.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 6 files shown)


tylerphelan (approved):

tomkennedy513 (approved):

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -760,10 +766,10 @@ func (b *Build) setupBindings() ([]corev1.Volume, []corev1.VolumeMount) {
 
 func (bc *BuildPodBuilderConfig) highestSupportedPlatformAPI(b *Build) (string, error) {
```
matthewmcnew, line 767: Could we instead have this method return a parsed platform api?

pkg/apis/build/v1alpha1/build_pod.go
```diff
@@ -390,12 +391,17 @@ func (b *Build) BuildPod(images BuildPodImages, secrets []corev1.Secret, taints
 							"-project-metadata=/layers/project-metadata.toml"},
 							cacheArgs,
 							func() []string {
-								if platformAPI == "0.3" {
+								platformAPIVersion := semver.MustParse(platformAPI)
```
matthewmcnew, line 394: A bit concerned about the "MustParse" here a bad builder with an invalid metadata could lead to a panic.

PR #799, opened 2021-08-11
Fix e2e tests to work with locally hosted docker registry

jromero:
This set of changes also includes updates to the documentation with instructions on local development using docker registry and ngrok.

Full description of changes:

1. Delete images by sha instead of tags due to [lack of widespread standardization](https://github.com/google/go-containerregistry/issues/999#issuecomment-828701797)
2. Fix: Images being deleted prematurely when using nested `it.Run`.
3. Move `it.Before` and `it.After` to `beforeAll` and `afterAll` to preserve desired behaviour.
4. Fix: Make `spec` library happy by using deterministic nested `it` tests.


DEVELOPMENT.md
```diff
@@ -0,0 +1,83 @@
+## Setup
+
+### Prerequisites
+
+  * [libgit2](https://libgit2.org/)
+    * macOS: `brew instal libgit2`
```
samj1912, line 6: note: I have had issues with versions of libgit2. kpack only seems to support v1.1.0

DEVELOPMENT.md
```diff
@@ -0,0 +1,83 @@
...
+#### Prerequisites
+
+  * [docker](https://docs.docker.com/get-docker/)
+  * [ngrok](https://ngrok.com/) - or any public routing solution
+
+1. In one terminal, start a docker registry
+    ```shell
+    docker run -p 5000:5000 --rm -e REGISTRY_STORAGE_DELETE_ENABLED=true registry:2
+    ```
+2. In another terminal, route port 5000:
```
samj1912, line 27: Is ngrok needed? For local k8s clusters like k3s etc it seems to work with localhost as well. Should we add this as an optional step?

jromero: Does k3s also handle SSL for you? I use ngrok (and recommend it) because it handles all of that for you. If you know what you are doing then you would probably be able to skip these steps altogether.

tylerphelan (approved):

Makefile
```diff
@@ -1,12 +1,17 @@
...
+.PHONY: unit
 unit:
 	$(GOCMD) test -v -count=1 -parallel=1 -timeout=0 ./pkg/...
 
+.PHONY: unit-ci
 unit-ci:
 	$(GOCMD) test -v -count=1 -parallel=1 -timeout=0 ./pkg/... -coverprofile=coverage.txt -covermode=atomic
 
-.PHONY: unit
+.PHONY: e2e
```
matthewmcnew, line 15: Nice!

test/execute_build_test.go
```diff
@@ -34,391 +35,403 @@ import (
...
+	}, metav1.CreateOptions{})
+	require.NoError(t, err)
 
-		waitUntilReady(t, ctx, clients, builder, clusterBuilder)
-	})
+	waitUntilReady(t, ctx, clients, builder, clusterBuilder)
+}
 
-	it("builds and rebases git, blob, and registry based images", func() {
+func afterAll(t *testing.T, cfg config) {
```
matthewmcnew, line 273: Can we just use a waitgroup on the part of the test that runs in pararallel?

DEVELOPMENT.md
jromero: Interesting. I've been running 1.1.1. What issues have you noticed?

samj1912: I meant 1.1.x. Ubuntu bionic for example ships with an older version of libgit2 which results in compilation errors.

jromero: Ahh, I see. I'll add a note.

jromero:
@matthewmcnew, I've resolved a few iterations of merge conflicts + broke out the tests into two files as they appear to be growing. The hope is to make them more manageable. This just being one step in that direction. Please provide feedback before I need to do yet another merge conflict. 😢

tomkennedy513 (approved):

matthewmcnew:
Closing as this has gone stale.

PR #802, opened 2021-08-11
Add support for additional tags

samj1912:
Adds support for additional tags to be specified as a part of the image spec. This would allow users to manually patch and specify additional tags to be pushed out. This can allow kpack users to push semver tags or other tags like git commit shas by a utility like `kp cli`.

Example use cases - 

`kp image save --additional-tag myimage:v1.2 --additional-tag myimage:$(git rev-parse HEAD)`

where the above would override the additionalTags array entirely with the newly supplied values

to patch and tag images with commit shas or version numbers.

Fixes #802 and possibly #755


docs/image.md
```diff
@@ -5,7 +5,8 @@ kpack will monitor the inputs to the image configuration to rebuild the image wh
 
 The following defines the relevant fields of the `image` resource spec in more detail:
 
-- `tag`: The image tag.
+- `tag`: The image tag. This is the primary image tag that is immutable for a given image and is re-used during subsequent builds for certain cached data.
+- `additionalTags`: Any additional list of image tags that should be published. This list of tags is mutable.
```
samj1912, line 9: I am a bit conflicted if we should only allow additional tags to be tag names or fully fledged docker image tags. The former would be easier in terms of secret management etc. and ensuring that pushes succeed where as the latter would be more flexible.  I am tempted to switch to the former but I have currently implemented the latter.

jromero: This seems relevant: https://github.com/buildpacks/community/discussions/128

samj1912:
@matthewmcnew this is ready for a review

matthewmcnew (approved):

tomkennedy513 (approved):

dumez-k (approved):

samj1912:
@matthewmcnew can we merge this?

matthewmcnew:
@samj1912 We plan on merging today. 

Thank you for your patience :)

PR #804, opened 2021-08-14
Add CD to kpack using Github actions

samj1912:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,13 +1,14 @@
-name: CI
+name: CI/CD
 
 on:
   push:
     branches:
-    - main
+      - main
+    tags:
+      - "v*"
   pull_request: {}
 
 jobs:
-
   test:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


hack/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM gcr.io/cf-build-service-public/ci/kpack-builder
+USER root
+RUN chmod -R 755 /cnb/lifecycle
+USER 1000:1000
```
samj1912, line 4: I had to do this since the latest builder seems incorrect. The /cnb/lifecycle directory has incorrect permissions that prevents pack from running the appropriate lifecycle binary once it drops permissions (even in trusted workflows during export).

hack/common.sh
```diff
@@ -4,10 +4,9 @@ function pack_build() {
     image=$1
     target=$2
     pack_args=${@:3}
-    builder="gcr.io/cf-build-service-public/ci/kpack-builder" # builder used ci
-
-    pack build ${image} --builder ${builder} -e BP_GO_TARGETS=${target} ${pack_args} --publish --trust-builder
-
+    (cd hack && docker build -t kpack-builder .)
+    builder="kpack-builder" # builder used ci
+    pack build ${image} --builder ${builder} -e BP_GO_TARGETS=${target} ${pack_args} --publish --trust-builder -t ${image}:${kpack_version} --cache-image ${image}-cache
```
samj1912, line 9: Added a cache image flag here to speed up GH rebuilds. I also had to add an extra tag here so that we can get proper versioning on the output image and so that github doesn't GC the untagged releases once latest has been pushed.

hack/common.sh
```diff
@@ -27,6 +26,7 @@ function compile() {
   # e.g. IMAGE_PREFIX=username/kpack-
   IMAGE_PREFIX=${IMAGE_PREFIX:-"${registry}/"}
 
+  kpack_version=${kpack_version:="dev"}
```
samj1912, line 29: The kpack version is set to 

dev-${COMMIT_SHA}

on push to main and `vX.Y.Z` when a matching version tag is published.

.github/workflows/ci.yaml
```diff
@@ -18,3 +19,56 @@ jobs:
...
+      - if: ${{ github.event_name != 'pull_request' || ! github.event.pull_request.head.repo.fork }}
+        name: Create release artifacts
+        env:
+          kpack_version: ${{ steps.get_version.outputs.version }}
+          kpack_repo: ghcr.io/${{ github.repository_owner }}/kpack
+        run: |
+          #!/usr/bin/env bash
+          set -euo pipefail
+          ./hack/release.sh ${kpack_repo} release-${kpack_version}.yaml
+      - uses: "marvinpinto/action-automatic-releases@latest"
```
samj1912, line 56: This creates a nightly release on each push to main

The release looks like - https://github.com/samj1912/kpack/releases/tag/latest and is updated on each push to main

.github/workflows/ci.yaml
```diff
@@ -18,3 +19,56 @@ jobs:
...
+        if: startsWith(steps.get_version.outputs.version, 'dev')
+        name: "Nightly release"
+        with:
+          repo_token: "${{ secrets.GITHUB_TOKEN }}"
+          automatic_release_tag: "latest"
+          prerelease: true
+          title: "Nightly kpack Build"
+          files: |
+            release-*.yaml       
+      - uses: "marvinpinto/action-automatic-releases@latest"
```
samj1912, line 66: This creates an actual release and adds a changelog by comparing the last available semver tag.

It looks like https://github.com/samj1912/kpack/releases/tag/v2.0.1

In order to trigger this workflow we need to push a tag to the kpack repo without creating a release. (This can also be done via the GH UI by quickly creating a release and tagging it to a tag and immediately deleting the release but not the tag. We have to go through such a round-about method from the UI since the UI doesn't support creating tags without releases unlike the git command line.)

.github/workflows/ci.yaml
```diff
@@ -18,3 +19,56 @@ jobs:
...
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name != 'pull_request' || ! github.event.pull_request.head.repo.fork }}
+    steps:
+      - id: checkout
+        name: Checkout code
+        uses: actions/checkout@v2
+      - name: Login to GitHub Package Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
```
samj1912, line 33: I chose to use ghcr here but this can be replaced with any repository of choice. ghcr seemed easiest since it requires no secret setup. For other container repositories we will need to add an appropriate token to login to that registry.

hack/Dockerfile
samj1912: This can be removed once the bug in the kpack builder is fixed.

matthewmcnew:
I'm going to close this tidy up our repo. Can we create an issue to discuss making CI/CD more public?

PR #805, opened 2021-08-15
Add support for project descriptor v0.2

samj1912:


pkg/cnb/project_descriptor.go
```diff
@@ -12,32 +12,58 @@ import (
...
 	} else if err != nil {
 		return fmt.Errorf("unable to determine if project descriptor file exists: %w", err)
 	}
-	_, err := toml.DecodeFile(file, &d)
+	d, err := parseProjectDescriptor(file, logger)
 	if err != nil {
 		return err
 	}
-	if d.Build.Buildpacks != nil {
+	if d.Buildpacks != nil || d.Group != nil {
```
samj1912, line 53:
```suggestion
	if d.Buildpacks != nil {
```

pkg/cnb/project_descriptor_test.go
```diff
@@ -42,6 +43,19 @@ func testProcessProjectDescriptor(t *testing.T, when spec.G, it spec.S) {
 		os.RemoveAll(platformDir)
 	})
 
+	when("unsupported project descriptor version", func() {
+		it.Before(func() {
+			ioutil.WriteFile(projectToml, []byte(`
+[_]
+schema-version = "0.3"
```
tylerphelan, line 50: maybe make this 0.99 or something because 0.3 will be supported next and this will need to be changed

pkg/cnb/project_descriptor.go
```diff
@@ -95,10 +121,24 @@ type build struct {
...
+	Project project `toml:"_"`
+	IO      ioTable `toml:"io"`
+}
+
+type project struct {
+	SchemaVersion string `toml:"schema-version"`
+}
+
+type ioTable struct {
+	Buildpacks build `toml:"buildpacks"`
```
tylerphelan, line 143: is this missing a level? isn't it `io.buildpacks.build` ?

pkg/cnb/project_descriptor_test.go
```diff
@@ -207,5 +221,223 @@ uri = "check-this-out.com"
...
+			it.Before(func() {
+				ioutil.WriteFile(projectToml, []byte(`
+[[build.groups]]
+id = "cool-buildpack"
+version = "v4.2"
+uri = "check-this-out.com"
+				`), 0644)
+			})
+			it("it ignores the table", func() {
+				assert.Nil(t, cnb.ProcessProjectDescriptor(appDir, platformDir, logger))
```
tylerphelan, line 234: should we log that the key is being ignored similar to buildpacks and builders?

pkg/cnb/project_descriptor.go
samj1912: It was changed later in the spec https://github.com/buildpacks/spec/blob/extensions/project-descriptor/0.2/extensions/project-descriptor.md#iobuildpacksgroup-optional

pkg/cnb/project_descriptor_test.go
samj1912: I have purposefully not logged it since it's an artifact of our implementation where I have both buildpacks and groups in the same struct and they are used in one version but not the other.

For builders and others we purposefully ignore a valid specced project descriptor because we don't support certain functionality.

tomkennedy513 (approved):

Issue #766, opened 2021-07-12
Add webhook conversion for api versions


matthewmcnew:
https://github.com/pivotal/kpack/pull/676 is adding the v1alpha2 api version but, does not provide support to convert across api versions. Before we release v1alpha2 to a wide audience we will need to add support to convert across versions.

PR #806, opened 2021-08-18
Add conversion webhook for crd apiversion migration

tylerphelan:
- Only used for images and builds which we know will require conversion
- Refactor reusable api types to pkg/apis/core

fix https://github.com/pivotal/kpack/issues/766

Co-authored-by: Tom Kennedy <ktom@vmware.com>


pkg/apis/build/v1alpha2/build_conversion.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"context"
+
+	"knative.dev/pkg/apis"
+
+	"github.com/pivotal/kpack/pkg/apis/build/v1alpha1"
+)
+
+func (b *Build) ConvertTo(_ context.Context, to apis.Convertible) error {
+	toBuild := to.(*v1alpha1.Build)
```
tomkennedy513, line 12: i think in the future we should look at doing something like this with the type switch https://github.com/tektoncd/pipeline/blob/67b318ba62d9dfee4e9ec639571754f42d9702d7/pkg/apis/pipeline/v1alpha1/cluster_task_conversion.go#L31-L50

tomkennedy513 (approved):

pkg/apis/build/v1alpha2/build_conversion.go
tylerphelan: yea we should probably have this, updated

tomkennedy513 (approved):

dumez-k (approved):
Looks great! 

Talking with @tylerphelan a little bit and ideally somewhere down the line it might be nice to refactor `corev1alpha1` into it's own package just called `core` rather than versioning it as I find that it leads to confusion a little bit between the two api versions. 

I recognize though that this would be a fairly expensive refactor for the value-add. Just a call-out.

pkg/apis/build/v1alpha1/image_types.go
```diff
@@ -105,4 +96,12 @@ func (i *Image) NamespacedName() types.NamespacedName {
 	return types.NamespacedName{Namespace: i.Namespace, Name: i.Name}
 }
 
+func (i *Image) ConvertTo(_ context.Context, _ apis.Convertible) error {
+	return errors.New("called convertTo in v1alpha1")
```
matthewmcnew, line 100: v1alpaha2 is the hub?

cmd/webhook/main.go
```diff
@@ -90,6 +93,37 @@ func validatingAdmissionController(ctx context.Context, _ configmap.Watcher) *co
...
 }
 
+func conversionController(ctx context.Context, _ configmap.Watcher) *controller.Impl {
+
+	conversions := map[schema.GroupKind]conversion.GroupKindConversion{
+		buildapi.Kind("Image"): {
+			DefinitionName: "images.kpack.io",
+			HubVersion:     buildapi.SchemeGroupVersion.Version,
+			Zygotes: map[string]conversion.ConvertibleObject{
+				buildapi.SchemeGroupVersion.Version: &buildapi.Image{},
```
tylerphelan, line 103: use `v1alpha2` to be more explicit about conversions

pkg/apis/build/v1alpha1/image_types.go
tylerphelan: update error message for more context

matthewmcnew (approved):
Excellent.

PR #815, opened 2021-08-30
Test conversions

tylerphelan:
- No longer test by converting to and back due to new fields that can be lost
- Add test to call out fields that cannot be converted back to v1alpha1


tomkennedy513 (approved):

pkg/apis/build/v1alpha2/build_conversion_test.go
```diff
@@ -113,48 +188,72 @@ func testBuildConversion(t *testing.T, when spec.G, it spec.S) {
...
-			convertedBackBuild := &Build{}
-			err = convertedBackBuild.ConvertFrom(context.TODO(), v1alpha1Build)
+			testV1alpha2Build := &Build{}
+			err = testV1alpha2Build.ConvertFrom(context.TODO(), v1alpha1Build)
 			require.NoError(t, err)
+			require.Equal(t, testV1alpha2Build, v1alpha2Build)
+		})
 
-			require.Equal(t, build, convertedBackBuild)
+		it("can't convert new fields to v1alpha1", func() {
```
matthewmcnew, line 227: What is this actually testing? That those fields don't cause it to error? Has the test ever failed?

tylerphelan: yea it doesn't really do anything. I just wanted to acknowledge the fields that don't convert to v1alpha1 and are dropped. I don't think this really tests anything.

matthewmcnew (approved):

PR #817, opened 2021-08-31
Cosign Signing Integration

DennyHoang:
This PR integrates cosign for image signing as specified in RFC #694

When an operator attaches a cosign formatted secret (contains at least cosign.key data key) to the service account used for building the image, the image will then attempt to be signed after it has been built and pushed to a repository.

# Status: Ready for Review
Currently, we are opening this draft PR to get an early review of our current implementation. There are still features that have not been implemented as listed in the `Features` checklist below. Also, open for discussion, would it be alright to have `Signing/Verifying builder/clusterBuilder images` as a separate PR such that the next version of kpack could have the initial image signing first? Then, we will work on implementing the `Signing/Verifying builder/clusterBuilder images` portion afterwards as the upcoming feature cutoff is nearing for the next release candidate for kpack. 

# Features:
- [x] Signing when keys are mounted for buildPod scenario
- [x] Signing when keys are mounted for rebasePod scenario
- [x] Add static annotations to signature for build number and build timestamp
- [x] Support custom annotations to be added to signature
- [x] Support for alternative registry for built image vs cosign signature
- [x] Support for legacy docker media type
- [x] ~Signing/Verifying builder/clusterBuilder images~ This will be addressed in a separate PR. Currently requires ability to provide a different keychain. Issue raised to cosign [here](https://github.com/sigstore/cosign/issues/657) and [here](https://github.com/sigstore/cosign/issues/666) for signing builder images as mentioned [here](https://github.com/pivotal/kpack/pull/817#discussion_r704818018)

## We had some questions which are added as "Review" items to help organize discussion into threads.


DennyHoang:
Asking for assistance and insight:
- `go.mod`: with the introduction of `cosign`, there are different version of dependencies that were indirectly imported. We currently are using a `replace` to force them to a version. When we run `go mod tidy` or anything else, it generally updates the versions in the `go.mod`. For example, `k8s.oi/api` is `v0.20.7`, however the previous command updates it to `v0.21.3`. Is there a way to prevent or improve this?

DennyHoang:
Asking for assistance and insight:
- `pkg/cosigner/image_signer.go`: For testing purposes, we used `cliSignCmd = cli.SignCmd` so that we can override this function during testing to validate parameter input while ignoring its functionality. Is this an appropriate approach or is there another best practice? We didn't see or find any particular mocking/mock generating libraries being included or other examples in the code when we first looked.

go.mod
```diff
@@ -6,34 +6,41 @@ require (
...
+	k8s.io/api v0.21.3
+	k8s.io/apimachinery v0.21.3
+	k8s.io/client-go v0.21.3
 	k8s.io/code-generator v0.20.7
-	k8s.io/kube-openapi v0.0.0-20210113233702-8566a335510f
+	k8s.io/kube-openapi v0.0.0-20210305001622-591a79e4bda7
 	knative.dev/pkg v0.0.0-20210819054404-bda81c029160
 )
+
+replace github.com/prometheus/common => github.com/prometheus/common v0.26.0
```
DennyHoang, line 40: Asking for assistance and insight:

go.mod: with the introduction of cosign, there are different version of dependencies that were indirectly imported. We currently are using a replace to force them to a version. When we run go mod tidy or anything else, it generally updates the versions in the go.mod. For example, k8s.oi/api is v0.20.7, however the previous command updates it to v0.21.3. Is there a way to prevent or improve this?

pkg/cosigner/image_signer.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/buildpacks/lifecycle"
+	"github.com/sigstore/cosign/cmd/cosign/cli"
+)
+
+type ImageSigner struct {
+	Logger *log.Logger
+}
+
+var (
+	cliSignCmd     = cli.SignCmd
```
DennyHoang, line 22: Asking for assistance and insight:

pkg/cosigner/image_signer.go: For testing purposes, we used cliSignCmd = cli.SignCmd so that we can override this function during testing to validate parameter input while ignoring its functionality. Is this an appropriate approach or is there another best practice? We didn't see or find any particular mocking/mock generating libraries being included or other examples in the code when we first looked.

docs/image.md
```diff
@@ -141,6 +142,56 @@ To create the notary secret used by kpack for image signing, run the following c
 - `<password>`: The password provided to encrypt the private key.
 - `<hash>.key`: The private key file.
 
+### <a id='cosign-config'></a>Cosign Configuration
```
DennyHoang, line 145: Would it be alright to have `Signing/Verifying builder/clusterBuilder images` as a separate PR such that the next version of kpack could have the initial image signing first? Then, we will work on implementing the `Signing/Verifying builder/clusterBuilder images` portion of the RFC as the upcoming feature cutoff is near for the next release candidate for kpack.

go.mod
tylerphelan: I think we will need the replace directive as you've done here to force:

```
k8s.io/api v0.20.7
k8s.io/apimachinery v0.20.7
k8s.io/client-go v0.20.7
```

but you can group the `replace` like `require`:

```
replace (
  k8s.io/client-go => k8s.io/client-go v0.20.7
  replace k8s.io/api => k8s.io/api v0.20.7
  replace github.com/tj/assert => github.com/tj/assert v0.0.3
)
```

cmd/completion/main.go
```diff
@@ -16,52 +19,94 @@ const (
...
 	flag.Parse()
 
-	if notaryV1URL != "" {
-		creds, err := dockercreds.ParseMountedAnnotatedSecrets(registrySecretsDir, dockerCredentials)
+	creds, err := dockercreds.ParseMountedAnnotatedSecrets(registrySecretsDir, dockerCredentials)
+	if err != nil {
+		logger.Fatal(err)
+	}
+
+	for _, c := range append(dockerCfgCredentials, dockerConfigCredentials...) {
```
matthewmcnew, line 56: I realize this was the existing behavior in completion but, it seems unnecessary to rebuild the keychain to just write to  `docker/config.json` again. Can we just mount the same credentials from export into completion?

pkg/cosigner/image_signer.go
```diff
@@ -0,0 +1,91 @@
...
+
+		if err := cliSignCmd(ctx, ko, annotations, refImage, "", true, "", false, false); err != nil {
+			return fmt.Errorf("unable to sign image with %s: %v", cosignKeyFile, err)
+		}
+	}
+
+	return nil
+}
+
+func findCosignFolders(dir string) []string {
```
matthewmcnew, line 74: Can we just mount all cosign creds at their own path and then do a directory list to avoid needing to creatively find the cosign secrets?

pkg/cosigner/image_signer.go
```diff
@@ -0,0 +1,91 @@
...
+
+func NewImageSigner(logger *log.Logger) *ImageSigner {
+	return &ImageSigner{
+		Logger: logger,
+	}
+}
+
+func (s *ImageSigner) Sign(reportFilePath string, annotations map[string]interface{}) error {
+	var report lifecycle.ExportReport
+	_, err := toml.DecodeFile(reportFilePath, &report)
```
matthewmcnew, line 34: Both Notary and Cosign are unmarshalling the report. Can we extract that to completion and pass in the report here?

cmd/completion/main.go
DennyHoang: Did you mean that I remount the `homeVolume` and set the Env `HOME` to `/builder/home`?

I noticed that `build-init` would save the config to `/builder/home/docker/config.json` and could therefore be reused by completion during the regular `buildPod` flow. However, during the `rebasePod` flow, as the only initContainer is the "rebase" container. It does load some docker credentials but it does not save it. I could introduce the "Save" portion to the rebase image and therefore the completion container could leverage a shared volume for both buildPod and rebasePod scenarios. Is it ideal to introduce this logic into "rebase" when it itself doesn't need the file to be saved just so we can leverage the byproduct of that outside of that container in completion?

The keychain still needed to be built or loaded into a variable to be passed into Notary while cosign only requires the config to be writing to a file named `config.json`. The keychain could be built from the `config.json` though instead of from each secret individually if mounting a shared volume is the way to go

pkg/cosigner/image_signer.go
DennyHoang: I've adjusted the buildPod to use a different mountPath when it is a cosign secret.
For finding the directory, I still have to loop through the folder at least and store secret folder names which is leveraged for COSIGN_REPOSITORY and COSIGN_DOCKER_MEDIA_TYPES information. Then using the folder name, I believe I still have to reconstruct the folder path to access the potential "cosign.key" and "cosign.password"

cmd/completion/main.go
DennyHoang: @matthewmcnew

pkg/cosigner/image_signer.go
DennyHoang: The thing I'm trying to wrap my head around right now is how would I use dependency injection if the function I am calling is not from an object. It is a public function, SignCmd, found [here](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/sign.go)
In the example you provided, the factory object is set during instantiation so that it can be any object that matches the interface. However, the function in question for the above issue is a public function not attached to an object. I'm relatively newer to GoLang so please let me know if this understanding is wrong and that dependency injection can still work for this.

This function is what we wanted functionally, and I didn't want to re-implement it or copy/paste it into the code. 

Regarding your second point, I'll continue investigating

pkg/notary/image_signer_test.go
```diff
@@ -17,6 +19,7 @@ import (
 	"github.com/theupdateframework/notary/storage"
 	"github.com/theupdateframework/notary/tuf/data"
 	"github.com/theupdateframework/notary/tuf/signed"
+	"github.com/tj/assert"
```
matthewmcnew, line 22: We use https://github.com/stretchr/testify instead of tj/assert.

pkg/cosigner/image_signer.go
matthewmcnew: Can't we just assume that each directory in `/var/builder/cosign/` properly provides cosign.key and cosign.password?

pkg/cosigner/image_signer.go
matthewmcnew: In golang you can define a type which is itself a function.

That would look something like this: 
```go
type SignFunc func(ctx context.Context, ko cli.KeyOpts, annotations map[string]interface{}, imageRef string, certPath string, upload bool, payloadPath string, force bool, recursive bool) error

type ImageSigner struct {
	Logger *log.Logger
	signFunc SignFunc
}

func NewImageSigner(logger *log.Logger, signFunc SignFunc) *ImageSigner {
	return &ImageSigner{
		Logger: logger,
		signFunc: signFunc,
	}
}
```


Then the cosign func can be dependency injected like this in main:

```go
imageSigner := NewImageSigner(logger, cli.SignCmd)
```

matthewmcnew: At this point the image signer is doing very little (which is a good thing) but, a test with a mocked out cosign is not giving us very much confidence that image signer is actually being invoked and used correctly. So, I would like to explore actually invoking the cosign library in the test. We test registry interactions in [client_test](https://github.com/pivotal/kpack/blob/main/pkg/registry/client_test.go) with a test http server and I imagine a cosign variation might look very similar. Happy to jump on a zoom to work on it together!

cmd/completion/main.go
matthewmcnew: I forgot about the complexities introduced by the rebase container. I think it would eventually be a good idea for the rebase container to "save" the credentials to a volume that can be loaded in completion. I don't think you need to introduce this in your PR.

pkg/cosigner/image_signer.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	refImage := report.Image.Tags[0]
+
+	ctx := context.Background()
+	for _, cosignFolder := range cosignFolders {
+		cosignKeyFile := fmt.Sprintf("%s/cosign.key", cosignFolder)
+		cosignPasswordFile := fmt.Sprintf("%s/cosign.password", cosignFolder)
+
+		ko := cli.KeyOpts{KeyRef: cosignKeyFile, PassFunc: func(bool) ([]byte, error) {
```
matthewmcnew, line 58: It is unfortunate that cosign does not seem to support providing an alternative keychain. It works to rely on the the default keychain here but, when we want to sign builder images we will need to provide the ggcr keychain to cosign.

matthewmcnew: Might be worth raising on issue on: https://github.com/sigstore/cosign/issues

cmd/completion/main.go
DennyHoang: Since that is the case, I will leave the current credential generation and saving inside the completion image. When we introduce and find a more uniform way of saving credentials, we can adjust the completion container then. For cosign portion in the completion container, with the shared volume during non rebase case, it seemed to work from my quick testing when mounting the shared home volume. For the notary portion, it is still likely that the keychain would need to be loaded to be passed into the notary sign function. Is there a place I can put stories or track this concern? Do I just make a Github issue saying to standardize docker credential saving using shared volumes to be more consistent between rebasePod/buildPod/initContainers/containers?

pkg/cosigner/image_signer.go
DennyHoang: I am assuming as such.

I modified the findCosignFolders function to return the list of `<cosign-secret-name>` where it looks into `/var/builder/cosign/<cosign-secret-name>`.

I could change this to be a walk function such that the signing occurs while it is recursing through the directories. The only requirement was that i needed the `cosign-secret-name` because I was utilizing that to access annotation values in the passed in maps for cosign repository and cosign docker media types.

pkg/notary/image_signer_test.go
DennyHoang: Oops, the VScode auto imported the wrong thing. That explains why go.mod started to act up for the github action for running the test coverage. Will fix this, thanks for catching it :)

pkg/cosigner/image_signer.go
DennyHoang: It looks like sigstore/cosign has an [e2e_test.go](https://github.com/sigstore/cosign/blob/main/test/e2e_test.go) which runs a fake registry locally, pushes an image, then runs the cosign signCmd and verify on it. Would extracting the necessary stuff from this and re-implementing it in our testing be sufficient to validating that cosign's signCmd works?

DennyHoang: Going to make a commit first where I validate the real SignCmd once following the e2e_test.go sample. 
A second commit, which can be reverted if we see fit, where I update all the existing signCmd mocks to also run the real signCmd after validating the arguments. Afterwards, doing an additional validation that the expected items were pushed to the registry.

go.mod
samj1912: created https://github.com/sigstore/cosign/issues/651

pkg/cosigner/image_signer.go
matthewmcnew: https://github.com/sigstore/cosign/issues/666

pkg/cosigner/image_signer.go
```diff
@@ -0,0 +1,107 @@
...
+func NewImageSigner(logger *log.Logger, signFunc SignFunc) *ImageSigner {
+	return &ImageSigner{
+		Logger:   logger,
+		signFunc: signFunc,
+	}
+}
+
+func (s *ImageSigner) Sign(report lifecycle.ExportReport, annotations map[string]interface{}, cosignRepositories map[string]interface{}, cosignDockerMediaTypes map[string]interface{}) error {
+	if len(report.Image.Tags) < 1 {
+		s.Logger.Println("no image tag to sign")
```
matthewmcnew, line 39: Do we actually think this would occur? It seems like an error case?

pkg/cosigner/image_signer.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	if len(cosignSecrets) == 0 {
+		s.Logger.Println("no keys found for cosign signing")
+		return nil
+	}
+
+	refImage := report.Image.Tags[0]
+
+	ctx := context.Background()
```
matthewmcnew, line 56: Can we pass in a context from main?

pkg/cosigner/image_signer.go
DennyHoang: Realistically, probably not going to occur. I will remove this

pkg/cosigner/image_signer.go
DennyHoang: Sure, we can pass it in :)

pkg/cosigner/image_signer.go
DennyHoang: Oh, they made it the day after my issue was raised :) Perfect! 
Thanks for letting me know

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -707,6 +667,16 @@ func (b *Build) setupSecretVolumesAndArgs(secrets []corev1.Secret, filter func(s
...
 			args = append(args, fmt.Sprintf("-ssh-%s=%s=%s", "git", secret.Name, annotatedUrl))
+		case string(secret.Data[COSIGNSecretDataCosignKey]) != "":
+			if cosignRepository := secret.ObjectMeta.Annotations[COSIGNRespositoryAnnotationPrefix]; cosignRepository != "" {
+				args = append(args, fmt.Sprintf("-cosign-repositories=%s=%s", secret.Name, cosignRepository))
+			}
+
+			if cosignDockerMediaType := secret.ObjectMeta.Annotations[COSIGNDockerMediaTypesAnnotationPrefix]; cosignDockerMediaType != "" {
+				args = append(args, fmt.Sprintf("-cosign-docker-media-types=%s=%s", secret.Name, cosignDockerMediaType))
+			}
+
```
stormqueen1990, line 678: For readability purposes, could this be extracted to a separate function/method?

cmd/completion/main.go
```diff
@@ -31,46 +42,96 @@ func init() {
...
+		homeDir, err := os.UserHomeDir()
+		if err != nil {
+			logger.Fatalf("error obtaining home directory: %v", err)
+		}
 
-			creds, err = creds.Append(dockerCfgCreds)
-			if err != nil {
-				logger.Fatal(err)
-			}
+		err = creds.Save(path.Join(homeDir, ".docker", "config.json"))
```
tylerphelan, line 87: this maybe should be `filepath.Join`

pkg/cosigner/image_signer.go
```diff
@@ -0,0 +1,101 @@
...
+	return &ImageSigner{
+		Logger:   logger,
+		signFunc: signFunc,
+	}
+}
+
+func (s *ImageSigner) Sign(ctx context.Context, report lifecycle.ExportReport, annotations map[string]interface{}, cosignRepositories map[string]interface{}, cosignDockerMediaTypes map[string]interface{}) error {
+	cosignSecrets, err := findCosignSecrets()
+	if err != nil {
+		s.Logger.Printf("no keys found for cosign signing: %v\n", err)
```
tylerphelan, line 40: should this return an error?

pkg/cosigner/image_signer.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil
+	}
+
+	if len(cosignSecrets) == 0 {
+		s.Logger.Println("no keys found for cosign signing")
+		return nil
+	}
+
+	refImage := report.Image.Tags[0]
```
tylerphelan, line 49: is this 100% safe to index? is it worth it to do a length check first?

pkg/cosigner/image_signer.go
```diff
@@ -0,0 +1,101 @@
...
+
+	for _, cosignSecret := range cosignSecrets {
+		cosignKeyFile := fmt.Sprintf("%s/%s/cosign.key", secretLocation, cosignSecret)
+		cosignPasswordFile := fmt.Sprintf("%s/%s/cosign.password", secretLocation, cosignSecret)
+
+		ko := cli.KeyOpts{KeyRef: cosignKeyFile, PassFunc: func(bool) ([]byte, error) {
+			content, err := ioutil.ReadFile(cosignPasswordFile)
+			if err != nil {
+				return []byte(""), nil
```
tylerphelan, line 58: should this return the err?

pkg/cosigner/image_signer.go
```diff
@@ -0,0 +1,101 @@
...
+			}
+
+			return content, nil
+		}}
+
+		if cosignRepository, ok := cosignRepositories[cosignSecret]; ok {
+			os.Setenv(cosignRepositoryEnv, fmt.Sprintf("%s", cosignRepository))
```
tylerphelan, line 65: is there a reason we don't handle the `os.Setenv` errors?

cmd/completion/main.go
```diff
@@ -31,46 +42,96 @@ func init() {
...
-		if err := signer.Sign(notaryV1URL, notarySecretDir, reportFilePath, creds); err != nil {
+		if err := signer.Sign(notaryV1URL, notarySecretDir, report, creds); err != nil {
 			logger.Fatal(err)
 		}
 	}
 
 	logger.Println("Build successful")
 }
+
+func mapKeyValueArgs(args flaghelpers.CredentialsFlags) map[string]interface{} {
```
tylerphelan, line 120: can this return `map[string]string` ?

DennyHoang: cosign's `cli.signCmd` takes in `map[string]interface{}` for it's annotation parameter. I didn't want two different functions, one returning `map[string]interface{}` and one returning `map[string]string` or having to convert between those two map types. It was easier to just do a `fmt.Sprintf("%s", <variable>)` cast when the `cosignRepository` and `cosignDockerMediaTypes` maps were used.

Would you recommend I still do a `map[string]string` and do a loop copy to a `map[string]interface{}` for annotations later?
If the current casting option is fine, is there a better way to cast an `interface{}` to a string? I was also looking at `variable.(string)` instead of using `fmt`.

pkg/cosigner/image_signer.go
DennyHoang: I'll update to handle

pkg/cosigner/image_signer.go
DennyHoang: It shouldnt error. "cosign.password" is an optional file or field in the secret. I will add a comment saying so here

pkg/cosigner/image_signer.go
DennyHoang: It should be very safe, but I'll add a check to be 100% safe

pkg/cosigner/image_signer.go
DennyHoang: Yes, I had confused this with the following length check on cosignSecrets.

pkg/cosigner/image_signer.go
stormqueen1990: We could, however, output a warning message just in case the password is missing but the key requires it (that would be a configuration error that can be hard to spot without logs).

pkg/cosigner/image_signer.go
DennyHoang: Trying to write the tests for this now

cmd/completion/main.go
```diff
@@ -16,13 +22,18 @@ const (
 	registrySecretsDir = "/var/build-secrets"
 	reportFilePath     = "/var/report/report.toml"
 	notarySecretDir    = "/var/notary/v1"
+	buildNumberKey     = "buildNumber"
```
tylerphelan, line 25: are these 2 `const`s used?

DennyHoang: We are currently using those for annotating cosign signatures so that each signature contains the `buildNumber` and `buildTimestamp`

pkg/apis/build/v1alpha2/cosign_types.go
```diff
@@ -0,0 +1,13 @@
+package v1alpha2
+
+// +k8s:openapi-gen=true
+type CosignConfig struct {
+	// +listType
+	Annotations []CosignAnnotation `json:"annotations"`
```
tylerphelan, line 6: I think `Annotations` and the `CosignAnnotation.Name` and `CosignAnnotation.Value` can be `omitempty` ?

pkg/cosigner/image_signer.go
```diff
@@ -0,0 +1,101 @@
...
+	signFunc SignFunc
+}
+
+const (
+	cosignRepositoryEnv       = "COSIGN_REPOSITORY"
+	cosignDockerMediaTypesEnv = "COSIGN_DOCKER_MEDIA_TYPES"
+)
+
+var (
+	secretLocation = "/var/build-secrets/cosign"
```
tylerphelan: can this be a const?

pkg/cosigner/image_signer.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+func NewImageSigner(logger *log.Logger, signFunc SignFunc) *ImageSigner {
+	return &ImageSigner{
+		Logger:   logger,
+		signFunc: signFunc,
```
tylerphelan, line 30: is `signFunc` only used for tests?

pkg/cosigner/image_signer.go
tylerphelan: maybe something like this just to be safe:

```
content, err := ioutil.ReadFile(cosignPasswordFile)
  if err == nil {
    return content, nil
  } else if errors.Is(err, os.ErrNotExist) {
    return []byte(""), nil
  } else {
    return []byte(""), err
}
```

pkg/cosigner/image_signer.go
tylerphelan: it may be hard to test this behavior, I wouldn't make a specific test for these error paths

cmd/completion/main.go
tylerphelan: I see, I think that's a good enough reason to keep it as `map[string]interface{}`

tylerphelan: `fmt` works fine to cast

pkg/cosigner/image_signer.go
DennyHoang: It isnt a const because I actually override this value during testing so I dont need this particular folder path

pkg/cosigner/image_signer.go
DennyHoang: Yes, I only override for testing purposes. Otherwise, it should be using cli.SignCmd

pkg/cosigner/image_signer.go
DennyHoang: I think it is possible to test if I change the `cosignRepositoryEnv` to be a variable. Then we would override it to "something=" which is invalid and would fail. Then we can validate these error messages.

We will give it a try for a bit. if it fails, we can possibly ignore because it is very unlikely to occur

stormqueen1990: @tylerphelan according to the [Linux manual page on `setenv()`](https://man7.org/linux/man-pages/man3/setenv.3.html), this should only fail when the variable names contain `=` or is empty/null. Would you say it's fair to not have specific tests to check the error path here?
We currently see very little value in writing a test, since we would have to force the flow by changing the environment variable name, which in this case should just be a constant.

pkg/cosigner/image_signer.go
DennyHoang: So the file could not exist or the file could be empty. When it is empty, the password would be an empty string based on the file content.

cmd/completion/main.go
DennyHoang: resolving

pkg/apis/build/v1alpha2/cosign_types.go
DennyHoang: done, resolving

pkg/cosigner/image_signer.go
tylerphelan: > Would you say it's fair to not have specific tests to check the error path here?

yea, I don't think testing the behavior of `os.Setenv` is worthwhile

pkg/cosigner/image_signer.go
tylerphelan: is `main` updated with `cli.SignCmd` or am I just not seeing it?

tylerphelan: nevermind, I see it now

cmd/completion/main.go
tylerphelan: hm, they are still showing as unused in my editor

pkg/cosigner/image_signer.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+}
+
+func (s *ImageSigner) Sign(ctx context.Context, report lifecycle.ExportReport, annotations map[string]interface{}, cosignRepositories map[string]interface{}, cosignDockerMediaTypes map[string]interface{}) error {
```
tylerphelan, line 37: this can also be written as:
```
Sign(ctx context.Context, report lifecycle.ExportReport, annotations, cosignRepositories, cosignDockerMediaTypes map[string]interface{}) error
```

pkg/cosigner/image_signer.go
tylerphelan: could this value be injected instead of reset in the tests then? I think it would be a worthwhile improvement.

pkg/cosigner/image_signer.go
```diff
@@ -0,0 +1,116 @@
+package cosigner
```
tomkennedy513, line 1: seems like this package could be called "cosign" to line up with the notary package naming

pkg/cosigner/image_signer.go
tomkennedy513: it could be passed in to the sign function or put inside the signer

cmd/completion/main.go
DennyHoang: OHHH this is the one in completion/main. My bad, I thought it was in build_pod... And yes, weird, VScode normally shows unused but you are right, I am no longer using those values. Will delete

pkg/cosigner/image_signer.go
DennyHoang: done, resolving

tomkennedy513 (approved):
Looks great! Would you mind squashing these commits and rebasing when everything is approved?

dumez-k (approved):

pkg/cosign/image_signer.go
```diff
@@ -0,0 +1,112 @@
...
+	return &ImageSigner{
+		Logger:   logger,
+		signFunc: signFunc,
+	}
+}
+
+func (s *ImageSigner) Sign(ctx context.Context, report lifecycle.ExportReport, secretLocation string, annotations, cosignRepositories, cosignDockerMediaTypes map[string]interface{}) error {
+	cosignSecrets, err := findCosignSecrets(secretLocation)
+	if err != nil {
+		return fmt.Errorf("error finding cosign signing keys: %v", err)
```
matthewmcnew, line 36: Can we use the `errors/pkg` instead of fmt.errors?
```
return errors.Errorf...
```

pkg/cosign/image_signer.go
```diff
@@ -0,0 +1,112 @@
...
+			}
+		}
+
+		if cosignDockerMediaType, ok := cosignDockerMediaTypes[cosignSecret]; ok {
+			if err := os.Setenv(cosignDockerMediaTypesEnv, fmt.Sprintf("%s", cosignDockerMediaType)); err != nil {
+				return fmt.Errorf("failed setting COSIGN_DOCKER_MEDIA_TYPES env variable: %v", err)
+			}
+		}
+
+		// Separate error catching because each function should be attempted
```
matthewmcnew, line 77: Is this to ensure that the env vars are unset?

tylerphelan: what about this?

```
func (s *ImageSigner) Sign(ctx context.Context, report lifecycle.ExportReport, secretLocation string, annotations, cosignRepositories, cosignDockerMediaTypes map[string]interface{}) error {
	cosignSecrets, err := findCosignSecrets(secretLocation)
	if err != nil {
		return fmt.Errorf("error finding cosign signing keys: %v", err)
	}

	if len(cosignSecrets) == 0 {
		s.Logger.Println("no keys found for cosign signing")
		return nil
	}

	if len(report.Image.Tags) == 0 {
		s.Logger.Println("no image found in report to sign")
		return nil
	}

	refImage := report.Image.Tags[0]

	for _, cosignSecret := range cosignSecrets {
		if err := s.sign(ctx, refImage, cosignSecret, secretLocation, annotations, cosignRepositories, cosignDockerMediaTypes); err != nil {
			return err
		}
	}

	return nil
}

func (s *ImageSigner) sign(ctx context.Context, refImage, cosignSecret, secretLocation string, annotations, cosignRepositories, cosignDockerMediaTypes map[string]interface{}) error {
	cosignKeyFile := fmt.Sprintf("%s/%s/cosign.key", secretLocation, cosignSecret)
	cosignPasswordFile := fmt.Sprintf("%s/%s/cosign.password", secretLocation, cosignSecret)

	ko := cli.KeyOpts{KeyRef: cosignKeyFile, PassFunc: func(bool) ([]byte, error) {
		content, err := ioutil.ReadFile(cosignPasswordFile)
		// When password file is not available, default empty password is used
		if err != nil {
			return []byte(""), nil
		}

		return content, nil
	}}

	if cosignRepository, ok := cosignRepositories[cosignSecret]; ok {
		if err := os.Setenv(cosignRepositoryEnv, fmt.Sprintf("%s", cosignRepository)); err != nil {
			return fmt.Errorf("failed setting %s env variable: %v", cosignRepositoryEnv, err)
		}
		defer os.Unsetenv(cosignRepositoryEnv)
	}

	if cosignDockerMediaType, ok := cosignDockerMediaTypes[cosignSecret]; ok {
		if err := os.Setenv(cosignDockerMediaTypesEnv, fmt.Sprintf("%s", cosignDockerMediaType)); err != nil {
			return fmt.Errorf("failed setting %s env variable: %v", cosignDockerMediaTypesEnv, err)
		}
		defer os.Unsetenv(cosignDockerMediaTypesEnv)
	}

	if err := s.signFunc(ctx, ko, annotations, refImage, "", true, "", false, false); err != nil {
		return fmt.Errorf("unable to sign image with %s: %v\n", cosignKeyFile, err)
	}

	return nil
}
```

go.mod
```diff
@@ -6,34 +6,40 @@ require (
...
+	k8s.io/client-go v0.21.3
 	k8s.io/code-generator v0.20.7
-	k8s.io/kube-openapi v0.0.0-20210113233702-8566a335510f
+	k8s.io/kube-openapi v0.0.0-20210305001622-591a79e4bda7
 	knative.dev/pkg v0.0.0-20210819054404-bda81c029160
 )
+
+replace (
+	github.com/prometheus/common => github.com/prometheus/common v0.26.0
+	github.com/tj/assert => github.com/tj/assert v0.0.3
```
matthewmcnew, line 42: We can drop this dep

pkg/cosign/image_signer.go
matthewmcnew: I really like the defers.

pkg/cosign/image_signer_test.go
```diff
@@ -0,0 +1,575 @@
...
+			})
+
+			it("signs images", func() {
+				cliSignCmdCallCount := 0
+				password1Count := 0
+				password2Count := 0
+
+				cliSignCmd := func(ctx context.Context, ko cli.KeyOpts, annotations map[string]interface{}, imageRef, certPath string, upload bool, payloadPath string, force, recursive bool) error {
+					t.Helper()
+					assert.Equal(t, testCtx, ctx)
```
matthewmcnew, line 101: I really like that these tests actually validate the signature. I think the additional validation of the values passed to the `cliSignCmd` here might be unnecessary (and may make it hard to use new api versions).

I think it is fine for now but, perhaps something to think about in the future.

matthewmcnew (approved):

pkg/cosign/image_signer.go
DennyHoang: Makes sense, I'll make this change

go.mod
DennyHoang: Sure, will try to remove. Last time, github action test failed pulling in this dependency for some reason and introducing this replace fixed it. Ill see what happens after next commit

pkg/cosign/image_signer.go
DennyHoang: sure, ill make this change. But for my own learning, what is the reason this is preferred?

tomkennedy513: it gives better stack traces

DennyHoang:
Please dont merge yet

tylerphelan (approved):

DennyHoang:
Squashed. Rebased. Thanks a bunch @tomkennedy513 for the help!

Tests Passed. Manually looked over diff and didn't see anything wrong with the squash.
Manually tested building an Image with cosign features and it worked (with a small adjustment to `build_pod.go` with a nil check and creation of NodeSelector map).

Should be good to merge at this moment :D 🎉

DennyHoang:
Hmm, small change is required.
If I build an image without cosign, it is trying to look into the `cosign` secret folder but it doesnt exist because no secrets were mounted there.

Going to make a small adjustment to log there instead of returning an error

DennyHoang:
Apart from that change, thanks @tomkennedy513 for your help with the squash!

If people are okay with that adjustment I made in the above comment/commit, then this PR is otherwise good to go!

cmd/completion/main.go
```diff
@@ -96,8 +97,10 @@ func main() {
 	cosignDockerMediaTypesOverrides := mapKeyValueArgs(cosignDockerMediaTypes)
 
 	ctx := context.Background()
-	if err := cosignSigner.Sign(ctx, report, cosignSecretLocation, annotations, cosignRepositoryOverrides, cosignDockerMediaTypesOverrides); err != nil {
-		logger.Fatalf("cosignSigner sign: %v\n", err)
+	if _, err := os.Stat(cosignSecretLocation); !os.IsNotExist(err) {
```
matthewmcnew, line 100: Can we update the signer to exit with an error now? https://github.com/pivotal/kpack/pull/817/files#diff-2b751ca77925350f424134c6ca29134ce0cb23dff9b4439c90dffdfcd27c407eR38

pkg/cosign/image_signer.go
```diff
@@ -0,0 +1,111 @@
...
+
+func (s *ImageSigner) Sign(ctx context.Context, report lifecycle.ExportReport, secretLocation string, annotations, cosignRepositories, cosignDockerMediaTypes map[string]interface{}) error {
+	cosignSecrets, err := findCosignSecrets(secretLocation)
+	if err != nil {
+		return errors.Errorf("no keys found for cosign signing: %v\n", err)
+	}
+
+	if len(cosignSecrets) == 0 {
+		s.Logger.Println("no keys found for cosign signing")
+		return nil
```
matthewmcnew, line 42: Could we error out here as well?

pkg/cosign/image_signer.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	if len(report.Image.Tags) == 0 {
+		s.Logger.Println("no image found in report to sign")
```
matthewmcnew, line 46: And error out here?

tylerphelan (approved):

dumez-k (approved):

PR #818, opened 2021-09-01
Add default process to image and build specs

dumez-k:


pkg/apis/build/v1alpha2/image_types.go
```diff
@@ -51,6 +51,7 @@ type ImageSpec struct {
 	ProjectDescriptorPath    string                            `json:"projectDescriptorPath,omitempty"`
 	Build                    *corev1alpha1.ImageBuild          `json:"build,omitempty"`
 	Notary                   *corev1alpha1.NotaryConfig        `json:"notary,omitempty"`
+	DefaultProcess           string
```
jromero, line 54: Curious to know when this object gets marshalled and why this new field is missing json format annotations (or is it just overlooked)?

pkg/apis/build/v1alpha2/build.go
```diff
@@ -141,6 +141,13 @@ func (b *Build) NotaryV1Config() *corev1alpha1.NotaryV1Config {
 	return b.Spec.Notary.V1
 }
 
+func (b *Build) DefaultProcess() string {
+	if b.Spec.DefaultProcess == "" {
```
jromero, line 145: I see a few other methods that have `b == nil` checks making me think that there are cases where a `nil` Build object is used and may be called. Should this method also have the same check?

FWIW, I only bring it up because of the existing pattern. In reality it would be more ideal that a `nil` Build object isn't called at all.

pkg/apis/build/v1alpha2/build_types.go
```diff
@@ -59,6 +59,7 @@ type BuildSpec struct {
 	Resources             corev1.ResourceRequirements `json:"resources,omitempty"`
 	LastBuild             *LastBuild                  `json:"lastBuild,omitempty"`
 	Notary                *corev1alpha1.NotaryConfig  `json:"notary,omitempty"`
+	DefaultProcess        string
```
jromero, line 62: Same question about missing json annotation.

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -410,11 +409,12 @@ func (b *Build) BuildPod(images BuildPodImages, secrets []corev1.Secret, taints
 									return nil
```
matthewmcnew, line 409: I think in all three cases we want to set a process type if the process type is set.

pkg/apis/build/v1alpha2/build.go
```diff
@@ -141,6 +141,13 @@ func (b *Build) NotaryV1Config() *corev1alpha1.NotaryV1Config {
...
+		return "web"
```
matthewmcnew, line 146: I think we only want to fall back to web if we are using the `0.4` and `0.5` platform api. On `0.6` we should not set a default process type as that would override the default process type set by builpacks?

matthewmcnew:
Looking at this PR made me realize we should probably add it to `kp` as well?

pkg/apis/build/v1alpha2/build_types.go
tylerphelan, line 62: I kinda like `DefaultProcessType` for this

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -7,12 +7,11 @@ import (
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/google/go-containerregistry/pkg/name"
+	corev1alpha1 "github.com/pivotal/kpack/pkg/apis/core/v1alpha1"
```
tylerphelan, line 10: we usually keep the internal packages at the bottom of the import list

pkg/apis/build/v1alpha2/build.go
matthewmcnew, line 146: I think we only want to fall back to web if we are using the `0.4` and `0.5` platform api. On `0.6` we should not set a default process type as that would override the default process type set by builpacks?

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -213,7 +212,7 @@ func (b *Build) BuildPod(images BuildPodImages, secrets []corev1.Secret, taints
 				step(corev1.Container{
 					Name:    "completion",
 					Image:   images.completion(config.OS),
-					Command: []string{"/cnb/process/web"},
+					Command: []string{fmt.Sprintf("/cnb/process/%s", b.DefaultProcess())},
```
matthewmcnew, line 215: This is just the completion's image default process type. It should not be affected by the build's process.

pkg/apis/build/v1alpha2/image_types.go
dumez-k: Overlooked, good catch!

pkg/apis/build/v1alpha2/build.go
dumez-k: Makes sense, added nil check to be safe.

pkg/apis/build/v1alpha2/build_types.go
dumez-k: Overlooked as well.

pkg/apis/build/v1alpha2/build.go
matthewmcnew: FWIW. The nil check is not necessary here because DefaultProcess will always be called on an existing build.

pkg/apis/build/v1alpha2/build.go
```diff
@@ -142,9 +142,10 @@ func (b *Build) NotaryV1Config() *corev1alpha1.NotaryV1Config {
 }
 
 func (b *Build) DefaultProcess() string {
-	if b.Spec.DefaultProcess == "" {
-		return "web"
+	if b == nil {
```
matthewmcnew, line 145: I think this nil check is not necessary now especially that this method has no logic.

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -404,20 +405,21 @@ func (b *Build) BuildPod(images BuildPodImages, secrets []corev1.Secret, taints
...
+								if b.DefaultProcess() == "" {
+									if platformAPI.Equal(lowestSupportedPlatformVersion) || platformAPI.Equal(highestSupportedPlatformVersion) {
+										return nil
+									} else {
+										return []string{fmt.Sprintf("-process-type=web")}
 									}
 								}
+								return []string{fmt.Sprintf("-process-type=%s", b.DefaultProcess())}
+							}(),
+							func() []string {
```
matthewmcnew, line 417: Nice!

pkg/apis/build/v1alpha2/build_types.go
```diff
@@ -59,7 +59,7 @@ type BuildSpec struct {
 	Resources             corev1.ResourceRequirements `json:"resources,omitempty"`
 	LastBuild             *LastBuild                  `json:"lastBuild,omitempty"`
 	Notary                *corev1alpha1.NotaryConfig  `json:"notary,omitempty"`
-	DefaultProcess        string
+	DefaultProcess        string                      `json:"DefaultProcess,omitempty"`
```
matthewmcnew, line 62: I think we want that to be `defaultProcess` to match the other json annotations.

pkg/apis/build/v1alpha2/build.go
dumez-k: Following the pattern of the other functions in `build.go`. For example `NotaryV1Config` on line 137 has a similar setup just in case a nil build object gets passed in. 

I agree it's probably unnecessary though but for the sake of consistency maybe we should still keep it.

matthewmcnew: Lets clean up other methods vs reimplementing other logic for consistency

dumez-k: 👍

tylerphelan (approved):

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -405,19 +405,21 @@ func (b *Build) BuildPod(images BuildPodImages, secrets []corev1.Secret, taints
...
-									return nil
-								case platformAPI.Equal(highestSupportedPlatformVersion):
-									return []string{
-										"-report=/var/report/report.toml",
-									}
-								default:
-									return []string{
-										"-process-type=web",
-										"-report=/var/report/report.toml",
+								if b.DefaultProcess() == "" {
```
matthewmcnew, line 408: Love this.

matthewmcnew (approved):
So good! 😭

matthewmcnew:
@dumez-k Can you rebase and merge 🙂

tomkennedy513 (approved):
great work!

tomkennedy513:
> @dumez-k Can you rebase and merge 🙂

can we squash too :)

PR #823, opened 2021-09-10
add support for kubernetes service bindings

tomkennedy513:
- deprecate cnb bindings and block users from adding them to v1alpha2 images and builds
- v1alpha1 images that are converted to v1alpha2 will still be able to use the cnb bindings but the bindings cannot be modified

Co-authored-by: Tyler Phelan <tylerp@vmware.com>


pkg/apis/core/v1alpha1/build_types.go
```diff
@@ -23,11 +23,11 @@ type BuildBuilderSpec struct {
 
 // +k8s:openapi-gen=true
 // +k8s:deepcopy-gen=true
-type Bindings []Binding
+type CnbBindings []CnbBinding
 
 // +k8s:openapi-gen=true
 // +k8s:deepcopy-gen=true
-type Binding struct {
+type CnbBinding struct {
```
tylerphelan, line 30: I kinda like `CNBBinding` for the type

pkg/apis/core/v1alpha1/service_bindings.go
```diff
@@ -0,0 +1,24 @@
+package v1alpha1
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+type ServiceBinding struct {
```
tylerphelan, line 7: should this go in `build/v1alpha2` ?

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -846,6 +810,80 @@ func builderSecretVolume(bbs corev1alpha1.BuildBuilderSpec) corev1.Volume {
 	}
 }
 
+func setupBindingVolumesAndMounts(bindings []ServiceBinding) ([]corev1.Volume, []corev1.VolumeMount) {
+	volumes := make([]corev1.Volume, 0)
+	volumeMounts := make([]corev1.VolumeMount, 0)
+
+	for _, binding := range bindings {
+		switch b := binding.(type) {
```
tylerphelan, line 818: can this be achieved with the interface instead of a type switch?

pkg/apis/core/v1alpha1/service_bindings.go
tomkennedy513: i moved this to core because the buildpod package was also using it, but i can explore other options

pkg/apis/build/v1alpha2/build_pod.go
tomkennedy513: it could, but i was liking the simplicity of the type switch and not having to have any intermediate service binding types

pkg/apis/core/v1alpha1/service_bindings.go
tylerphelan: sorry i had a comment here but deleted it

tylerphelan (approved):
love it

matthewmcnew:
This should close: https://github.com/pivotal/kpack/pull/555

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -846,6 +810,80 @@ func builderSecretVolume(bbs corev1alpha1.BuildBuilderSpec) corev1.Volume {
...
+		case *corev1alpha1.ServiceBinding:
+			if b.SecretRef != nil {
```
matthewmcnew, line 820: Is this necessary?

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -846,6 +810,80 @@ func builderSecretVolume(bbs corev1alpha1.BuildBuilderSpec) corev1.Volume {
...
+				volumeMounts = append(volumeMounts,
+					corev1.VolumeMount{
+						Name:      secretVolume,
+						MountPath: fmt.Sprintf("%s/bindings/%s/secret", platformVolume.MountPath, b.Name),
+						ReadOnly:  true,
+					},
+				)
+			}
+		default:
+			break
```
matthewmcnew, line 880: Is this actually an error case?

pkg/apis/build/v1alpha2/build_validation.go
```diff
@@ -25,9 +28,27 @@ func (bs *BuildSpec) Validate(ctx context.Context) *apis.FieldError {
...
+
+	return bs.CnbBindings.Validate(ctx)
+}
+
+func resourceCreatedByKpackController(info *authv1.UserInfo) bool {
+	if info == nil {
+		return false
+	}
+
+	return info.Username == "system:serviceaccount:kpack:controller"
```
matthewmcnew, line 51: Elegant.

pkg/apis/build/v1alpha2/build_validation.go
```diff
@@ -25,9 +28,27 @@ func (bs *BuildSpec) Validate(ctx context.Context) *apis.FieldError {
...
 		Also(bs.LastBuild.Validate(ctx).ViaField("lastBuild")).
-		Also(bs.validateImmutableFields(ctx))
+		Also(bs.validateImmutableFields(ctx)).
+		Also(bs.validateCnbBindings(ctx).ViaField("cnbBindings"))
+}
+
+func (bs *BuildSpec) validateCnbBindings(ctx context.Context) *apis.FieldError {
+	//only allow the kpack controller to create resources with cnb bindings
+	if !resourceCreatedByKpackController(apis.GetUserInfo(ctx)) && len(bs.CnbBindings) > 0 {
+		return apis.ErrDisallowedFields("")
```
matthewmcnew, line 40: Should we return a deprecation notice?

pkg/apis/build/v1alpha2/image_builds.go
```diff
@@ -34,7 +34,7 @@ type BuildReason string
 func (im *Image) Build(sourceResolver *SourceResolver, builder BuilderResource, latestBuild *Build, reasons, changes string, nextBuildNumber int64) *Build {
 	buildNumber := strconv.Itoa(int(nextBuildNumber))
 
-	return &Build{
+	build := &Build{
```
matthewmcnew, line 37: Any reason to not just return this?

pkg/apis/build/v1alpha2/image_conversion.go
```diff
@@ -46,8 +45,14 @@ func (is *ImageSpec) convertTo(to *v1alpha1.ImageSpec) {
 	to.SuccessBuildHistoryLimit = is.SuccessBuildHistoryLimit
 	to.ImageTaggingStrategy = is.ImageTaggingStrategy
 	to.Source = is.Source
-	to.Build = is.Build
 	to.Notary = is.Notary
+
+	if fromBuild := is.Build; fromBuild != nil {
```
matthewmcnew, line 50: Why fromBuild?

pkg/apis/core/v1alpha1/service_bindings.go
```diff
@@ -0,0 +1,24 @@
...
+	Name      string
+	SecretRef *corev1.LocalObjectReference
```
matthewmcnew, line 9: Does this need to be a pointer?

pkg/buildpod/generator.go
```diff
@@ -84,13 +91,64 @@ func (g *Generator) buildAllowed(ctx context.Context, build BuildPodable) error
 		}
 	}
 
-	for _, binding := range build.Bindings() {
-		if binding.SecretRef != nil && forbiddenSecrets[binding.SecretRef.Name] {
-			return fmt.Errorf("binding %q uses forbidden secret %q", binding.Name, binding.SecretRef.Name)
+	bindings := make([]buildapi.ServiceBinding, 0)
+
+	if services := build.Services(); len(services) != 0 {
+		for _, s := range build.Services() {
```
matthewmcnew, line 97:
```suggestion
		for _, s := range services {
```

pkg/buildpod/generator.go
```diff
@@ -84,13 +91,64 @@ func (g *Generator) buildAllowed(ctx context.Context, build BuildPodable) error
...
+			var b corev1alpha1.ServiceBinding
+			if s.Kind == "Secret" {
+				b = corev1alpha1.ServiceBinding{
+					Name:      s.Name,
+					SecretRef: &corev1.LocalObjectReference{Name: s.Name},
+				}
+			} else {
+				gvr, _ := meta.UnsafeGuessKindToResource(s.GroupVersionKind())
```
matthewmcnew, line 105: Can we extract this to provisonedService, err := readProvisonedServiceDuckType()

pkg/buildpod/generator.go
```diff
@@ -84,13 +91,64 @@ func (g *Generator) buildAllowed(ctx context.Context, build BuildPodable) error
...
+				ps := duckprovisionedserviceable.ProvisionedServicable{}
+				if err := duck.FromUnstructured(unstructured, &ps); err != nil {
+					return nil, err
+				}
+				b = corev1alpha1.ServiceBinding{
+					Name:      s.Name,
+					SecretRef: ps.Status.Binding,
+				}
+			}
+			if forbiddenSecrets[b.SecretRef.Name] {
```
matthewmcnew, line 119: Could we check forbidden secrets in its own method?

pkg/buildpod/generator.go
```diff
@@ -84,13 +91,64 @@ func (g *Generator) buildAllowed(ctx context.Context, build BuildPodable) error
...
+				return nil, fmt.Errorf("build rejected: service %q uses forbidden secret %q", b.Name, b.SecretRef.Name)
```
matthewmcnew, line 120: Lets use pkg/errors to get those sweet stack traces.

pkg/buildpod/generator.go
```diff
@@ -84,13 +91,64 @@ func (g *Generator) buildAllowed(ctx context.Context, build BuildPodable) error
...
+			}
+
+			secret, err := g.K8sClient.CoreV1().Secrets(build.GetNamespace()).Get(ctx, b.SecretRef.Name, metav1.GetOptions{})
+			if err != nil {
+				return nil, err
+			}
+
+			if t, ok := secret.StringData["type"]; !ok || secret.Type == "" || fmt.Sprintf("service.binding/%s", t) != string(secret.Type) {
```
matthewmcnew, line 128: Do we actually care?

pkg/apis/build/v1alpha2/build_pod.go
tomkennedy513: probably not, since the status.binding field needs to be present for the duck type so the secret ref shouldn't be nil

tomkennedy513: i'll take it out

pkg/apis/build/v1alpha2/build_pod.go
tomkennedy513: i don't see a case where we would actually ever hit this point, but we can through an error

pkg/apis/build/v1alpha2/image_builds.go
tomkennedy513: I pulled it out to a variable to debug something and never put it back. good catch

pkg/apis/build/v1alpha2/image_conversion.go
tomkennedy513: I reused it from the convertFrom func, I'll rename it

pkg/buildpod/generator.go
tomkennedy513: are you thinking we just let it through and let the build potentially fail?

pkg/apis/core/v1alpha1/service_bindings.go
tomkennedy513: probably not, but the binding in the provisioned service is nilable, so a pointer made sense

pkg/apis/build/v1alpha2/build_pod.go
tomkennedy513: actually I think there is a case where the bindings are nil in the provisioned service status so the secret ref would be nil as well

pkg/apis/build/v1alpha2/build_validation.go
```diff
@@ -25,9 +28,27 @@ func (bs *BuildSpec) Validate(ctx context.Context) *apis.FieldError {
...
+
+	return bs.CnbBindings.Validate(ctx)
+}
+
+func resourceCreatedByKpackController(info *authv1.UserInfo) bool {
+	if info == nil {
+		return false
+	}
+
+	return info.Username == "system:serviceaccount:kpack:controller"
```
tylerphelan, line 51: I think we should pull `"system:serviceaccount:kpack:controller"` to a `const`

tomkennedy513:
> Curious about the name `CnbBinding`. What's CNB about it? As far as I can tell this is based on the [k8s service binding spec](https://github.com/servicebinding/spec#service-binding-specification-for-kubernetes). Is there something the Cnb prefix is trying to differentiate?

the cnb bindings are supposed to represent the bindings that kpack originally supported (https://github.com/buildpacks/rfcs/blob/main/text/0012-service-binding.md). The name is trying to differentiate from the kubernetes service bindings (https://github.com/servicebinding/spec) that v1alpha2 is supporting

dumez-k (approved):

pkg/reconciler/image/build_required.go
```diff
@@ -85,15 +85,15 @@ func configChange(img *buildapi.Image, lastBuild *buildapi.Build, srcResolver *b
 		old = buildchange.Config{
 			Env:       lastBuild.Spec.Env,
 			Resources: lastBuild.Spec.Resources,
-			Bindings:  lastBuild.Spec.Bindings,
+			Services:  lastBuild.Spec.Services,
```
matthewmcnew, line 88: Should we also include old cnb Bindings in here?

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -840,6 +807,80 @@ func builderSecretVolume(bbs corev1alpha1.BuildBuilderSpec) corev1.Volume {
...
+				volumeMounts = append(volumeMounts,
+					corev1.VolumeMount{
+						Name:      secretVolume,
+						MountPath: fmt.Sprintf("%s/bindings/%s/secret", platformVolume.MountPath, b.Name),
+						ReadOnly:  true,
+					},
+				)
+			}
+		default:
+			return nil,nil, errors.New(fmt.Sprintf("unsupported binding type: %T", b))
```
matthewmcnew, line 877:
```suggestion
			return nil,nil, errors.Errorf("unsupported binding type: %T", b)
```

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -840,6 +807,80 @@ func builderSecretVolume(bbs corev1alpha1.BuildBuilderSpec) corev1.Volume {
 	}
 }
 
+func setupBindingVolumesAndMounts(bindings []ServiceBinding) ([]corev1.Volume, []corev1.VolumeMount, error) {
+	volumes := make([]corev1.Volume, 0)
```
matthewmcnew, line 811: This is minor but, if we are initalizing a nil slice the general pattern should be `var volumes []corev1.Volume` if we want the performance optimization of make we should use the size we know that we will have `make([]corev1.Volume, 0, len(bindings))`

pkg/apis/build/v1alpha2/build_validation.go
```diff
@@ -2,13 +2,18 @@ package v1alpha2
...
+	"regexp"
 
+	authv1 "k8s.io/api/authentication/v1"
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/kmp"
 
 	"github.com/pivotal/kpack/pkg/apis/validate"
 )
 
+const KpackControllerServiceAccountUsername = "system:serviceaccount:kpack:controller"
```
matthewmcnew, line 15:
```suggestion
const kpackControllerServiceAccountUsername = "system:serviceaccount:kpack:controller"
```
I don't think we want to export this field?

pkg/apis/build/v1alpha2/build_validation_test.go
```diff
@@ -125,75 +126,132 @@ func testBuildValidation(t *testing.T, when spec.G, it spec.S) {
...
 				{Name: "&", MetadataRef: &corev1.LocalObjectReference{Name: "metadata"}},
 			}
 
-			assertValidationError(build, apis.ErrInvalidValue("&", "spec.bindings[0].name"))
-		})
+			assertValidationError(build, ctx, apis.ErrInvalidValue("&", "spec.cnbBindings[0].name"))
 
-		it("validates bindings have metadata", func() {
-			build.Spec.Bindings = []corev1alpha1.Binding{
+			// validates cnb bindings have metadata
```
matthewmcnew, line 228: Should these be subtests?

pkg/apis/core/v1alpha1/source_types.go
```diff
@@ -52,9 +49,9 @@ func (g *Git) BuildEnvVars() []corev1.EnvVar {
 	}
 }
 
-func (in *Git) ImagePullSecretsVolume() corev1.Volume {
+func (in *Git) ImagePullSecretsVolume(name string) corev1.Volume {
 	return corev1.Volume{
-		Name: imagePullSecretsDirName,
+		Name: name,
```
matthewmcnew, line 54: Subtle change but, a good one!

pkg/duckprovisionedserviceable/fake/fakeprovisionedservice.go
```diff
@@ -0,0 +1,38 @@
+/*
+Copyright 2020 VMware, Inc.
```
matthewmcnew, line 2: It is not 2020 and I think we have not been placing this in kpack. Can we drop it?

PR #826, opened 2021-09-13
Handle nil-able cache fields in Image/Build Conversions

matthewmcnew:


tomkennedy513 (approved):

dumez-k (approved):

pkg/apis/build/v1alpha2/build_conversion.go
```diff
@@ -55,10 +57,12 @@ func (bs *BuildSpec) convertTo(to *v1alpha1.BuildSpec) {
 func (bs *BuildSpec) convertFrom(from *v1alpha1.BuildSpec) {
 	bs.Env = from.Env
 	bs.Source = from.Source
-	bs.Cache = &BuildCacheConfig{
-		Volume: &BuildPersistentVolumeCache{
-			ClaimName: from.CacheName,
-		},
+	if from.CacheName != "" {
```
tylerphelan, line 60: is this necessary?

tylerphelan: oh, to keep `Cache` `nil` if `CacheName==""`

tylerphelan (approved):

pkg/apis/build/v1alpha2/build_conversion.go
matthewmcnew: Yeah I waffled on this. I don't think it is really functionally necessary but, It seemed more accurate to not create a populated `spec.cache.volume` object if the cache name is not even set.

PR #833, opened 2021-09-27
Support creating keychain from multiple volume secrets

tylerphelan:
- continue supporting historical env var used to set the path to the secret dir


pkg/dockercreds/volume_secrets_keychain.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"os"
+	"strings"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+)
+
+const (
+	SecretFilePathEnv  = "CREDENTIAL_PROVIDER_SECRET_PATH" // historically only support one path
+	SecretFilePathsEnv = "CREDENTIAL_PROVIDER_SECRET_PATHS"
```
dumez-k, line 12: should we document this new env variable somewhere as well?

tomkennedy513 (approved):

pkg/dockercreds/volume_secrets_keychain.go
```diff
@@ -0,0 +1,40 @@
...
+		if err != nil {
+			return nil, err
+		}
+		secrets.Append(newSecrets)
+	}
+
+	return secrets, nil
+}
+
+func getSecretFolders() []string {
```
matthewmcnew, line 29: Could we use the `directory` term over `folders` to be consisent?

pkg/dockercreds/volume_secrets_keychain.go
```diff
@@ -0,0 +1,40 @@
...
+	secretFoldersEnvVal, ok := os.LookupEnv(SecretFilePathsEnv)
+	if !ok {
+		secretFolderEnvVal, ok := os.LookupEnv(SecretFilePathEnv)
+		if !ok {
+			return []string{}
+		}
+		return []string{secretFolderEnvVal}
+	}
+
+	return strings.Split(secretFoldersEnvVal, ",")
```
jromero, line 39: Wouldn't it be more expected to split paths by colon (`:`) as opposed to commas? (prior art: [PATH](https://en.wikipedia.org/wiki/PATH_(variable)))

matthewmcnew:
I would like to propose that we merge https://github.com/pivotal/kpack/pull/837 instead of growing the scope of credentials that get mounted into the kpack controller. With https://github.com/pivotal/kpack/pull/837 and the new v1alpha2 api all clusterstores/clusterstacks/lifecycle can configure a service account with image pull secrets directly in the api.

PR #846, opened 2021-10-04
use v1alpha1 in image waiter

tomkennedy513:
this is needed to for [kp](https://github.com/vmware-tanzu/kpack-cli) and other consumers of the logs package to maintain backwards compatibility with earlier versions of kpack. This can be bumped to the next lowest supported version once v1alpha1 is no longer supported.

### DIFF
--- a/pkg/logs/wait_for_image.go
+++ b/pkg/logs/wait_for_image.go
@@ -11,7 +11,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	watchTools "k8s.io/client-go/tools/watch"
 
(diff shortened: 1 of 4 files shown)


dumez-k (approved):

matthewmcnew (changes requested):
Why do you think this should be done? Can you update the PR/Commit to explain your thoughts here?

tomkennedy513:
> Why do you think this should be done? Can you update the PR/Commit to explain your thoughts here?

updated the commit and pr

matthewmcnew (approved):

PR #847, opened 2021-10-04
Make the Image reconciliation generate deterministic build names

matthewmcnew:
kpack appears to periodically reconcile the Image before its internal Build lister is updated with the previous reconcile's Build. This leads to subsequent reconciles creating duplicate builds.

By using deterministic build names the reconcile of an image will result in an error attempting to create a new build if a image reconcile starts before the informer has the previously created build.


tomkennedy513 (approved):

matthewmcnew:
The risk here would be that the build name for the next build has already been created manually outside of the image reconciler loop the image reconciler would never be able to successfully reconcile.

dumez-k (approved):

PR #855, opened 2021-10-07
Add docs for k8s service bindings

tylerphelan:


docs/servicebindings.md
```diff
@@ -0,0 +1,123 @@
...
+1. Natively using a Secret (most common use case for kpack users)
+2. Using ProvisionedServices as described in the [specification](https://github.com/k8s-service-bindings/spec#provisioned-service) (requires additional operator configuration external to kpack)
+
+## Create a Service Binding with a Secret
+
+Requirements:
+
+* A Secret containing the service binding data
+  * The Secret `stringData` field **must** contain a key-value pairs of `type:<binding type>`. The buildpacks will use read this type
+  * The Secret `type` (not `stringData.type`) **must** be set to `service.binding/<binding type>` where `<binding type>` is the value of the key set in the above bullet.
```
matthewmcnew, line 18: We don't validate this and I don't believe it is a hard requirement in the spec. Should we drop it to a **should**?

PR #857, opened 2021-10-07
Disambiguate an image resource from a built OCI image

tylerphelan:


dumez-k (changes requested):
dammit, looks like my comments were on the commit itself(?) instead of the PR. didn't know that was possible, sorry!

docs/tutorial.md
```diff
@@ -39,7 +39,7 @@ This tutorial will walk through creating a kpack [builder](builder.md) resource
    
    > Note: Learn more about kpack secrets with the [kpack secret documentation](secrets.md) 
 
-1. Create a service account that references the registry secret created above 
+2. Create a service account that references the registry secret created above
```
matthewmcnew, line 42: Is there a reason to move to numbering explicitly? Mardown will number correctly with 1.s

docs/tutorial.md
```diff
@@ -240,16 +241,16 @@ This tutorial will walk through creating a kpack [builder](builder.md) resource
...
     
-1. kpack rebuilds
+9. kpack rebuilds
     
-   We recommend updating the kpack image configuration with a CI/CD tool when new commits are ready to be built.
+   We recommend updating the kpack image resource with a CI/CD tool when new commits are ready to be built.
    > Note: You can also provide a branch or tag as the `spec.git.revision` and kpack will poll and rebuild on updates!  
 
-   We can simulate an update from a CI/CD tool by updating the `spec.git.revision` on the image configured in step #6.
+   We can simulate an update from a CI/CD tool by updating the `spec.git.revision` on the image resourceured in step #6.
```
AidanDelaney, line 249:
```suggestion
   We can simulate an update from a CI/CD tool by updating the `spec.git.revision` on the image resource used in step #6.
```

docs/tutorial.md
tylerphelan: my preference is usually just to explicitly number them to keep it simple

docs/tutorial.md
tylerphelan: thank you!

PR #863, opened 2021-10-21
Fix compilation error in ./test/

phil9909:
We noticed that `go test ./test/` failed with a compilation error:

```                                                                                                                                                      
# github.com/pivotal/kpack/test [github.com/pivotal/kpack/test.test]                                                                                                                           
test/execute_build_test.go:218:5: unknown field 'ServiceAccount' in struct literal of type "github.com/pivotal/kpack/pkg/apis/build/v1alpha2".NamespacedBuilderSpec                            
```

**Note:**
Maybe you should consider at least compiling the e2e tests in the CI.
This can be done with:

```bash
go test -c -o /tmp/test-bin ./test/
```

### DIFF
--- a/test/execute_build_test.go
+++ b/test/execute_build_test.go
@@ -215,7 +215,7 @@ func testCreateImage(t *testing.T, when spec.G, it spec.S) {
 						},
 					},
 				},
-				ServiceAccount: serviceAccountName,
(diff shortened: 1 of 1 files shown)


phil9909:
Probably introduced by #861, therefore mentioning @matthewmcnew

PR #864, opened 2021-10-21
update e2e test to use ServiceAccountName

tomkennedy513:


tomkennedy513:
> Please consider my note in #863 to prevent this next time.

That's a good idea. Also, I didn't see your PR when I opened this one, that's my fault.

PR #889, opened 2021-11-19
Add support for platform API 0.7 and 0.8

### LINKED ISSUE #886: Support Platform api 0.7 / platform api 0.8
**Why**
To support the full feature set of the Lifecycle `v0.13.x` including sBOM support, kpack builds need to support platform api 0.8. 

**What**
Kpack builds should support and elect to use [platform api 0.8](https://github.com/buildpacks/spec/releases) if available on the executed lifecycle. If the 0.8 platform api is not available the build should utilize the 0.7 platform api. The included version of the lifecycle in the kpack release should be updated to the most recent release of the lifecycle. 

**Complications**
* This will be a slightly more complicated update because the analyze phase moves before the detect phase.

samj1912:
This PR adds support for 0.7 and 0.8.

Resolves #886

### DIFF
--- a/pkg/apis/build/v1alpha2/build_pod.go
+++ b/pkg/apis/build/v1alpha2/build_pod.go
@@ -235,14 +235,19 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
 					return []string{}
 				}
 				tags := []string{}
-				for _, tag := range b.Spec.Tags {
-					tags = append(tags, "-tag="+tag)
+				if len(b.Spec.Tags) > 1 {
+					for _, tag := range b.Spec.Tags[1:] {
+						tags = append(tags, "-tag="+tag)
+					}
 				}
 				return tags
 			}(),
 			func() []string {
 				if b.Spec.LastBuild != nil && b.Spec.LastBuild.Image != "" {
-					return []string{b.Spec.LastBuild.Image}
+					if platformAPILessThan07 {
+						return []string{b.Spec.LastBuild.Image}
+					}
+					return []string{"-previous-image=" + b.Spec.LastBuild.Image, b.Tag()}
 				}
 				return []string{b.Tag()}
 			}(),
--- a/pkg/apis/build/v1alpha2/build_pod.go
+++ b/pkg/apis/build/v1alpha2/build_pod.go
@@ -189,24 +189,109 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
 		MountPath: sourceVolume.MountPath,
 		SubPath:   b.Spec.Source.SubPath, // empty string is a nop
 	}
-
+	platformAPILessThan07 := platformAPI.LessThan(semver.MustParse("0.7"))
 	var genericCacheArgs []string
+	var analyzerCacheArgs []string = nil
 	var exporterCacheArgs []string
 	var cacheVolumes []corev1.VolumeMount
 
 	if b.Spec.NeedVolumeCache() && buildContext.os() != "windows" {
 		genericCacheArgs = []string{"-cache-dir=/cache"}
 		cacheVolumes = []corev1.VolumeMount{cacheVolume}
+		if platformAPILessThan07 {
+			analyzerCacheArgs = genericCacheArgs
+		}
 		exporterCacheArgs = genericCacheArgs
 	} else if b.Spec.NeedRegistryCache() {
 		useCacheFromLastBuild := (b.Spec.LastBuild != nil && b.Spec.LastBuild.Cache.Image != "")
 		if useCacheFromLastBuild {
 			genericCacheArgs = []string{fmt.Sprintf("-cache-image=%s", b.Spec.LastBuild.Cache.Image)}
 		}
+		analyzerCacheArgs = genericCacheArgs
 		exporterCacheArgs = []string{fmt.Sprintf("-cache-image=%s", b.Spec.Cache.Registry.Tag)}
 	} else {
 		genericCacheArgs = nil
 	}
+
+	analyzeContainer := corev1.Container{
+		Name:      "analyze",
+		Image:     b.Spec.Builder.Image,
+		Command:   []string{"/cnb/lifecycle/analyzer"},
+		Resources: b.Spec.Resources,
+		Args: args([]string{
+			"-layers=/layers",
+			"-analyzed=/layers/analyzed.toml"},
+			analyzerCacheArgs,
+			func() []string {
+				if platformAPILessThan07 {
+					return []string{
+						"-group=/layers/group.toml",
+					}
+				}
+				return []string{}
+			}(),
+			func() []string {
+				if platformAPILessThan07 {
+					return []string{}
+				}
+				tags := []string{}
+				for _, tag := range b.Spec.Tags {
+					tags = append(tags, "-tag="+tag)
+				}
+				return tags
+			}(),
+			func() []string {
+				if b.Spec.LastBuild != nil && b.Spec.LastBuild.Image != "" {
+					return []string{b.Spec.LastBuild.Image}
+				}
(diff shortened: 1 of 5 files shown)


pkg/apis/build/v1alpha2/build_pod_test.go
```diff
@@ -460,6 +541,50 @@ func testBuildPod(t *testing.T, when spec.G, it spec.S) {
...
+					corev1.VolumeMount{
+						Name:      "binding-secret-some-v1alpha1-binding",
+						MountPath: "/platform/bindings/some-v1alpha1-binding/secret",
+						ReadOnly:  true,
+					},
+				)
+			}
+		})
+
+		it("configures the v1alpha1bindings", func() {
```
natalieparellano, line 561:
```suggestion
		it("configures the v1alpha1bindings for older apis", func() {
```

pkg/apis/build/v1alpha2/build_pod_test.go
```diff
@@ -670,22 +811,46 @@ func testBuildPod(t *testing.T, when spec.G, it spec.S) {
...
 			require.NoError(t, err)
 
+			assert.Equal(t, pod.Spec.InitContainers[1].Name, "analyze")
+			assert.Contains(t, pod.Spec.InitContainers[1].Env, corev1.EnvVar{Name: "CNB_PLATFORM_API", Value: "0.8"})
+			assert.Equal(t, pod.Spec.InitContainers[1].Image, builderImage)
+			assert.Equal(t, []string{
+				"layers-dir",
+				"workspace-dir",
+				"home-dir",
+				"cache-dir",
```
natalieparellano, line 837: Technically it is not necessary to mount in the cache directory, because the analyzer (for newer platform api) will not use it.

pkg/apis/build/v1alpha2/build_pod_test.go
```diff
@@ -718,13 +908,16 @@ func testBuildPod(t *testing.T, when spec.G, it spec.S) {
 				"home-dir",
 				"cache-dir",
 			}, names(pod.Spec.InitContainers[2].VolumeMounts))
+			tags := []string{}
+			for _, tag := range build.Spec.Tags {
+				tags = append(tags, "-tag="+tag)
+			}
```
natalieparellano, line 914:
```suggestion
```
`-tag` is unsupported by the analyzer for older platform api.

natalieparellano: The code [here](https://github.com/pivotal/kpack/pull/889/files#diff-796b50408de7dbf79786503ffc9edfbc1eb29abc4c36902f37c6c5714606333bR233-R242) looks correct, so maybe it's at odds with this expectation.

pkg/apis/build/v1alpha2/build_pod_test.go
samj1912: Yup - I just left it there since it made the code simple and it didn't do any harm. But can move it out if needed.

samj1912: @natalieparellano removed it conditionally based on platform api.

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -251,6 +341,7 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
 			}),
 			SecurityContext: podSecurityContext(buildContext.BuildPodBuilderConfig),
 			InitContainers: steps(func(step func(corev1.Container, ...stepModifier)) {
+				// analyze,
```
tylerphelan, line 344: intentional?

samj1912: Nope - removed.

tylerphelan (approved):

tylerphelan:
do we want to include the `stack.toml` or `-run-image` for analyze? https://github.com/buildpacks/spec/blob/main/platform.md#inputs (@natalieparellano)

tylerphelan (approved):

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -189,24 +189,114 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
...
+					return []string{}
+				}
+				tags := []string{}
+				for _, tag := range b.Spec.Tags {
+					tags = append(tags, "-tag="+tag)
+				}
+				return tags
+			}(),
+			func() []string {
+				if b.Spec.LastBuild != nil && b.Spec.LastBuild.Image != "" {
```
matthewmcnew, line 244: This is providing the "previous image" and falling back to the destination tag if the previous image is not available. With platform api 0.7+ and the addition of the previous-image flag we should provide the previous image explicitly.

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -189,24 +189,114 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
...
+					}
+				}
+				return []string{}
+			}(),
+			func() []string {
+				if platformAPILessThan07 {
+					return []string{}
+				}
+				tags := []string{}
+				for _, tag := range b.Spec.Tags {
```
matthewmcnew, line 238: Should this just be tags[1+] because we will provide a tag as the only positional arg to the analyzer?

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -189,24 +189,114 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
...
+		userprofileHomeEnv(),
+	)
+	detectContainer := corev1.Container{
+		Name:      "detect",
+		Image:     b.Spec.Builder.Image,
+		Command:   []string{"/cnb/lifecycle/detector"},
+		Resources: b.Spec.Resources,
+		Args: []string{
+			"-app=/workspace",
+			"-group=/layers/group.toml",
```
matthewmcnew, line 283: It would be nice to be explicit about `/layers/analyzed.toml`.

matthewmcnew:
Do we want to also relax our mixin validation logic within that platform api?

pkg/apis/build/v1alpha2/build_pod.go
samj1912: Not sure if I follow you? Detector doesn't take in analyzed as in input flag. It is specified for analyzer, restorer and exporter currently.

matthewmcnew:
Added an additional issue to track mixin validation relaxation: https://github.com/pivotal/kpack/issues/891

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -305,69 +400,32 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
...
-						ImagePullPolicy: corev1.PullIfNotPresent,
-						Env: []corev1.EnvVar{
-							{
-								Name:  platformAPIEnvVar,
-								Value: platformAPI.Original(),
-							},
-						},
-					},
-					ifWindows(buildContext.os(), addNetworkWaitLauncherVolume(), useNetworkWaitLauncher(dnsProbeHost))...,
+					func() corev1.Container {
```
matthewmcnew, line 403: I'd like to explore moving the instantiation of these containers within the pod spec (with the accepted possibility of some duplication). The step pattern was originally added to allow all the containers to be explicitly provided in the pod spec with some logic. 

I think we can have something like: 
```go
if platformAPILessThan07 {
    step(detect...)
    step(analyze..)
}else {
    step(analyze...)
    step(detect..)
}
						
```

matthewmcnew: Happy to explore that in a future refactor though :)

matthewmcnew (approved):

PR #898, opened 2021-12-01
Support http proxy env vars for git fetch

tylerphelan:
- Supports HTTP_PROXY, HTTPS_PROXY, http_proxy, https_proxy, no_proxy

### DIFF
--- a/pkg/git/proxy.go
+++ b/pkg/git/proxy.go
@@ -1,0 +1,30 @@
+package git
+
+import (
+	git2go "github.com/libgit2/git2go/v31"
+	giturls "github.com/whilp/git-urls"
+	"golang.org/x/net/http/httpproxy"
+)
+
+func proxyFromEnv(gitURL string) (git2go.ProxyOptions, error) {
+	u, err := giturls.Parse(gitURL)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/git/proxy.go
```diff
@@ -0,0 +1,30 @@
+package git
+
+import (
+	git2go "github.com/libgit2/git2go/v31"
+	giturls "github.com/whilp/git-urls"
+	"golang.org/x/net/http/httpproxy"
+)
+
+func proxyFromEnv(gitURL string) (git2go.ProxyOptions, error) {
+	u, err := giturls.Parse(gitURL)
```
matthewmcnew, line 10: Should we use `gitUrls` [here](https://github.com/pivotal/kpack/blob/ccb2f75404c088b8c2c9f9f585f1e1ec2c7ffa57/pkg/git/git_keychain.go#L180)?

matthewmcnew (approved):
😲

tomkennedy513 (approved):

pkg/git/proxy.go
tylerphelan: nice!

PR #901, opened 2021-12-02
Upgrade git2go to v33 and libgit2 to 1.3

tylerphelan:
- run goimports


tomkennedy513 (approved):

matthewmcnew:
This will require an update the to kpack/git2go-buildpack

tylerphelan:
current issues:
- basic auth over https is broken. PR for fix here: https://github.com/libgit2/git2go/pull/870
- when no credentials match git url, build results in: `fetching remote: no credentials found for https://github.com/cloudfoundry-samples/cf-sample-app-nodejs`

matthewmcnew:
Should we bump this as well for the github action? https://github.com/pivotal/kpack/blob/main/.github/actions/run-tests/Dockerfile#L1

tylerphelan:
> Should we bump this as well for the github action? https://github.com/pivotal/kpack/blob/main/.github/actions/run-tests/Dockerfile#L1

good call

pkg/git/certificate_check_callback.go
```diff
@@ -1,40 +1,44 @@
...
+			return nil
 		}
 
 		if cert.Kind == git2go.CertificateX509 {
 			if cert.X509 != nil {
 				err := cert.X509.VerifyHostname(hostname)
 				if err != nil {
-					logger.Println("host name could not be verified")
-					return git2go.ErrAuth
+					msg := "host name could not be verified"
```
matthewmcnew, line 20: Why pull these out?

pkg/git/certificate_check_callback.go
```diff
@@ -1,40 +1,44 @@
...
+					logger.Println(msg)
```
matthewmcnew, line 21: Do we need a logger here now that the error message is being returned? As in will it still be printed?

matthewmcnew (approved):

PR #908, opened 2021-12-24
Add priority classes to builds based on change reason

### LINKED ISSUE #774: kpack operator triggered rebuilds lead to cluster starvation
If the operator triggers a rebuild across the cluster due to a change in a buildpack, kpack currently schedules a lot of builds which may cause user submitted image/builds to be queued/starved for a long time. We should ideally implement rate-limiting or priority queues for rebuilds triggered automatically v/s by a user.

samj1912:

Fixes #774


config/controller.yaml
```diff
@@ -65,13 +65,16 @@ spec:
         app: kpack-controller
         version: #@ data.values.version
     spec:
+      priorityClassName: system-cluster-critical
```
AidanDelaney, line 68: Where is the priority class `system-cluster-critical` defined?  I can only see `kpack-build-high-priority` and `kpack-build-low-priority` defined in `priority.yaml`.

AidanDelaney (commented):
I have one question, otherwise LGTM.

config/controller.yaml
samj1912: It's a default class. See https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#how-to-use-priority-and-preemption

AidanDelaney (approved):
@samj1912 provided the link to https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/#marking-pod-as-critical in answer to my question.  LGTM.

pkg/buildchange/buildpack_change.go
```diff
@@ -36,3 +36,5 @@ func (b buildpackChange) IsBuildRequired() (bool, error) {
 func (b buildpackChange) Old() interface{} { return b.old }
 
 func (b buildpackChange) New() interface{} { return b.new }
+
+func (s buildpackChange) Priority() buildapi.BuildPriority { return buildapi.BuildPriorityLow }
```
tylerphelan, line 40: receiver `s`

samj1912:
@matthewmcnew @tylerphelan @tomkennedy513 could you please review?

tomkennedy513 (approved):

cmd/controller/main.go
```diff
@@ -63,6 +73,7 @@ var (
 	rebaseImage            = flag.String("rebase-image", os.Getenv("REBASE_IMAGE"), "The image used to perform rebases")
 	completionImage        = flag.String("completion-image", os.Getenv("COMPLETION_IMAGE"), "The image used to finish a build")
 	completionWindowsImage = flag.String("completion-windows-image", os.Getenv("COMPLETION_WINDOWS_IMAGE"), "The image used to finish a build on windows")
+	enablePriorityClasses  = flag.Bool("enable-priority-classes", getEnvBool("ENABLE_PRIORITY_CLASSES", false), "if set to true, enables different pod priority classes for normal builds and automated builds")
```
matthewmcnew, line 76: This is the first user provided configuration for kpack via a flag. The other flags are not really designed to be modified by the end user. In the past we have discussed a kpack-config config map that would allow users to modify the behavior of kpack (defaults, etc). Would this be a good impetus to do this now?

pkg/apis/build/v1alpha2/image_builds.go
```diff
@@ -31,9 +32,15 @@ const (
...
-func (im *Image) Build(sourceResolver *SourceResolver, builder BuilderResource, latestBuild *Build, reasons, changes string, nextBuildNumber int64) *Build {
+func (im *Image) Build(sourceResolver *SourceResolver, builder BuilderResource, latestBuild *Build, reasons, changes string, nextBuildNumber int64, priorityClass string) *Build {
 	buildNumber := strconv.Itoa(int(nextBuildNumber))
-
+	buildAnnotations := map[string]string{
+		BuildReasonAnnotation:  reasons,
+		BuildChangesAnnotation: changes,
+	}
+	if priorityClass != "" {
+		buildAnnotations[BuildPriorityAnnotation] = priorityClass
```
matthewmcnew, line 42: It seems like this field is actionably modifying the behavior of the build. Could it be configured on the build.spec?

matthewmcnew: Obviously, that introduces complexity with the apis.

cmd/controller/main.go
tylerphelan: do we definitely need the feature flag? maybe if it's requested?

pkg/apis/build/v1alpha2/image_builds.go
tylerphelan: +1 spec

samj1912:
@matthewmcnew updated to a spec field in commit https://github.com/pivotal/kpack/pull/908/commits/79dce7a8e067b0370f31b1838c2174e19a9356fb

Can't get openapi gen to work correctly and it is producing a large diff. Is there a specific version I should use?

tomkennedy513 (approved):

samj1912:
@tylerphelan thanks for that. codegen and openapigen should both be in sync now. I noticed some of the codegen changes were not applied from previous PRs so updated those as well.

tylerphelan:
code gen seems to be correct, I think we can merge

matthewmcnew (approved):

PR #911, opened 2022-01-04
add ytt config for configuring kp config at install

tomkennedy513:
The reason for this change is to allow users to template the kp_default_repository creds as well as proxy values by using the carvel ytt tool.


matthewmcnew:
@tomkennedy513 I'm about to give this a look. Could you provide more context on this PR for why you want to do this and what this helps accomplish?

config/overlays/kapp-config.yaml
```diff
@@ -0,0 +1,31 @@
+---
+apiVersion: kapp.k14s.io/v1alpha1
```
matthewmcnew, line 2: I understand that some folks may use kapp to install kpack but, is this appropriate in the kpack repo?

config/overlays/ca-cert.yaml
```diff
@@ -0,0 +1,20 @@
+#@data/values
+#@ load("@ytt:data", "data")
+#@ load("@ytt:assert", "assert")
+
+#@ def find_cert():
```
matthewmcnew, line 5: I know @cppforlife discouraged us from doing this. I wonder if we should just provide ca cert as a regular value.

config/overlays/kp-config.yaml
```diff
@@ -0,0 +1,43 @@
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name:  kp-config
+  namespace: kpack
+data:
+  default.repository: #@ values.kp_default_repository
+  default.repository.serviceaccount: kp-default-repository-serviceaccount
+  canonical.repository: #@ values.kp_default_repository
```
matthewmcnew, line 11: When can we drop this value again?

tomkennedy513: i think it makes sense for us to remove now

config/overlays/ca-cert.yaml
tomkennedy513: ya that sounds less confusing anyways

config/overlays/overlay.yaml
```diff
@@ -0,0 +1,119 @@
...
+            #@ end
+            #@ if values.no_proxy != None:
+            #@overlay/append
+            - name: NO_PROXY
+              value: #@ values.no_proxy
+          #@ end
+          #@overlay/match missing_ok=True
+          volumeMounts:
+            - name: credentials
+              mountPath: /var/kpack/credentials
```
matthewmcnew, line 71: I wonder if this could be excluded because these creds can now be provided in the v1alpha2 api.

matthewmcnew: I suppose that depends on `kp`'s ability to set them.

tomkennedy513: i think we can remove this once kp supports v1alpha2 service accounts

config/overlays/values.star
```diff
@@ -0,0 +1,25 @@
...
+    end
+elif len(parts) == 1:
+	assert.fail("kp_default_repository must be a valid writeable repository and must include a '/'")
+end
+
+kp_default_docker_auth = base64.encode("{}:{}".format(data.values.kp_default_repository_username, data.values.kp_default_repository_password))
+kp_default_docker_creds = {"username": data.values.kp_default_repository_username, "password": data.values.kp_default_repository_password, "auth": kp_default_docker_auth}
+kp_default_docker_configjson = base64.encode(json.encode({"auths": {kp_default_registry: kp_default_docker_creds}}))
+
+values = data.values
```
matthewmcnew, line 25: Instead of exporting this "values" object could we just export `kp_default_repository` etc? Similar to what we do with kp_default_docker_configjson?

matthewmcnew:
Can I close this @tomkennedy513

PR #914, opened 2022-01-19
update slack info in README

tomkennedy513:


tylerphelan (approved):

matthewmcnew:
Do we want to encourage k8s slack over the CNCF slack (when buildpacks move)?

tomkennedy513:
> Do we want to encourage k8s slack over the CNCF slack (when buildpacks move)?

My understanding is that kpack would not be able to move to the cncf slack

matthewmcnew:
I think they just wouldn't automatically migrate it but, we could create a kpack channel.

tomkennedy513:
Oh I must have misunderstood then. I thought kpack had to be in the cncf to have a channel there

matthewmcnew:
My take would be wait until the CNB slack is migrated and then decide where we would prefer the "official" channel to live.

samj1912:
You should be able to create a slack channel in the cncf slack. It's just that we can't migrate it along with the rest of the project I believe.

dumez-k (approved):

matthewmcnew (approved):

Issue #921, opened 2022-02-10
Binding to a ProvisionedService object swallows errors and proceeds with no logging


julian-hj:
While attempting to use the [ProvisionedService service binding](https://github.com/pivotal/kpack/blob/main/docs/servicebindings.md#create-a-service-binding-with-a-provisionedservice), we found that the binding was missing from our build pod, with no log messages explaining why.

Eventually we figured out that the kpack admin role lacked access to our custom type, but to do that, we had to rebuild the controller with additional trace messages injected into the code.

We found that [this method](https://github.com/pivotal/kpack/blob/ba7ecb34d1d5f7733307298c843e82507d8f297b/pkg/buildpod/generator.go#L140) returns blank object if the target object cannot be found (for any reason) or if it doesn't fulfill the ProvisionedService duck type. 

It's very easy to get the binding wrong--either because the roles aren't set up, or because the type is specified incorrectly, or because some controller hasn't yet set the status on the object, so it doesn't yet conform to the duck type. So it would be much better if there were log messages in the kpack controller logs and/or a status message on the Image object to help diagnose the issue.

matthewmcnew:
@julian-hj Does the ProvisonedService define a ClusterRole with a label matching `servicebinding.io/controller=true`? If so, [this PR](https://github.com/pivotal/kpack/pull/922) should give the kpack controller rbac access.

matthewmcnew:
Before this issue is closed we should add better error handling here.

julian-hj:
Hi @matthewmcnew yeah, that PR would definitely help some, but I think it would still be better to error & requeue the `Image` when the ProvisionedService can't be fully resolved, rather than proceeding.

PR #922, opened 2022-02-10
Allow kpack controller to aggregate rbac rolebindings to access ProvisionedServices resources

matthewmcnew:
https://github.com/pivotal/kpack/issues/921

### DIFF
--- a/config/controllerrole.yaml
+++ b/config/controllerrole.yaml
@@ -9,6 +9,10 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: kpack-controller-admin
+aggregationRule:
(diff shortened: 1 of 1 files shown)


julian-hj (commented):
I think it would be useful to also add a sentence or two to [the documentation](https://github.com/pivotal/kpack/blob/main/docs/servicebindings.md#create-a-service-binding-with-a-provisionedservice) to state that there must be a ClusterRole with the `servicebinding.io/controller` label. Although it is called out in the [k8s docs](https://github.com/servicebinding/spec#considerations-for-role-based-access-control-rbac) it is listed as a "should" and is pretty easy to miss.

tylerphelan (approved):

PR #931, opened 2022-03-08
RFC: Introduce Buildpack/ClusterBuildpack Resources 

matthewmcnew:
[Readable](https://github.com/pivotal/kpack/blob/buildpack-resource/rfcs/0000-buildpack-resource.md)


rfcs/0000-buildpack-resource.md
```diff
@@ -0,0 +1,104 @@
...
+spec:
+  tag: gcr.io/sample/builder
+  stack:
+    name: bionic-stack
+    kind: ClusterStack
+  order:
+  - group:
+    - id: paketo-buildpacks/java      
+```
+The `paketo-buildpacks/java` buildpack referenced in the Builder order would be sourced from the Buildpack resource within the `my-builder` namespace or the ClusterBuildpack resource that provides the `paketo-buildpacks/java` with the highest version number as defined by semver. If multiple Buildpack resources provide the same selected Buildpack the Builder should reconcile with an error.
```
tomkennedy513, line 59: Could this get you into a state where someone else creates a buildpack or clusterbuildpack providing the same id and version that would inadvertently cause your builder to error out

rfcs/0000-buildpack-resource.md
```diff
@@ -0,0 +1,104 @@
...
+    kind: ClusterStack
+  order:
+  - group:
+    - id: paketo-buildpacks/java      
+    - name: my-custom-app-buildpack-0.3.1
+      kind: Buildpack
+```
+
+**Open Questions:**
+* Should ClusterBuilders be able to use Buildpacks or only ClusterBuildpacks?
```
tylerphelan, line 81: clusterbuildpacks for simplicity

rfcs/0000-buildpack-resource.md
```diff
@@ -0,0 +1,104 @@
...
+
+
+**Prior Art:**
+* [kapp controller](https://carvel.dev/kapp-controller/) models available Package versions as indepdent Package resources.
```
tylerphelan, line 85: `indepdent`

rfcs/0000-buildpack-resource.md
```diff
@@ -0,0 +1,104 @@
...
+
+**Risks:**
+This will require introducing a new api version within kpack. Future kpack releases will need to maintain support for ClusterStore resources until the kpack apis which contain the ClusterStore are removed.
```
tylerphelan, line 88: What does backwards compatibility look like?

rfcs/0000-buildpack-resource.md
```diff
@@ -0,0 +1,104 @@
...
+
+**Actions to take:**
+
+Introduce a new ClusterBuildpack and Buildpack resource that provides a buildpack to be available within a Cluster or a singular namespace and be used in ClusterBuilders and Builders.
+
+The Buildpack resource will allow for an individual buildpack to be configured with a single image reference.
+
+Sample Cluster Buildpack Resource
+
+```yaml
```
tylerphelan, line 21: what does the status look like? Probably should have version and id. Supported stack ids?

rfcs/0000-buildpack-resource.md
```diff
@@ -0,0 +1,104 @@
...
+
+The Buildpack resource will allow for an individual buildpack to be configured with a single image reference.
+
+Sample Cluster Buildpack Resource
+
+```yaml 
+apiVersion: kpack.io/v1alpha3
+kind: ClusterBuildpack
+metadata:
+  name: paketo-java-buildpack-5.9.1
```
tylerphelan, line 25: naming conventions are kinda important here, having version in the name would probably be standard

rfcs/0000-buildpack-resource.md
tylerphelan: >If multiple Buildpack resources provide the same selected Buildpack the Builder should reconcile with an error.
 
Maybe buildpacks should fail to reconcile if a buildpack with the same version/id combo already exists.

tylerphelan, line 59: Use case:
I have a builder that I want to keep confined to my controlled Buildpack versions within my namespace. My builder uses `paketo-buildpacks/java` but the `Buildpack` that provides java is out of sync with the `ClusterBuildpack` that provides `paketo-buildpacks/java`. How to I ensure my `Builder` only uses the `Buildpacks` I'm managing in my ns instead of picking up the latest `paketo-buildpacks/java` `ClusterBuildpack`?

Is this use case relevant? We could combine the use of `id` and `kind` on `Builders` to specify they want to only the `Buildpack + paketo-buildpacks/java` combo or `ClusterBuildpack + paketo-buildpacks/java`

rfcs/0000-buildpack-resource.md
tylerphelan: Are we creating an undesired workflow where users create source of truth `Buildpack` resource named `paketo-java-buildpack` and then simply update that buildpack instead of create new ones?

rfcs/0000-buildpack-resource.md
```diff
@@ -0,0 +1,104 @@
...
+
+Sample Buildpack Resource
+
+```yaml
+apiVersion: kpack.io/v1alpha3
+kind: Buildpack
+metadata:
+  name: my-custom-app-buildpack-0.3.1
+spec:
+  image: gcr.io/my-org/my-custom-app-buildpack:0.3.1
```
tylerphelan, line 38: should this be immutable? might at safety

rfcs/0000-buildpack-resource.md
matthewmcnew: I think this is a risk but, inherent if a builder is letting its versions float without direct references to a buildpack.

rfcs/0000-buildpack-resource.md
matthewmcnew: 👍

rfcs/0000-buildpack-resource.md
matthewmcnew: > Maybe buildpacks should fail to reconcile if a buildpack with the same version/id combo already exists.

I think that would be difficult to enforce.

rfcs/0000-buildpack-resource.md
matthewmcnew: I think kpack should be un-opinionated about the workflow here. If users want to continually update a Buildpack instead of creating a new buildpack that should work!

rfcs/0000-buildpack-resource.md
matthewmcnew: I think kpack should be un-opinionated about the workflow here.

matthewmcnew: In the future, we may add `kapp-controller` esque tag selection to enable kpack to poll for buildpack updates.

rfcs/0000-buildpack-resource.md
matthewmcnew: You can explicitly reference a Buildpack resource in the `spec.order`. Do we want to introduce other mechanisms to restrict where buildpacks can be sourced from on the `spec`?

rfcs/0000-buildpack-resource.md
```diff
@@ -0,0 +1,218 @@
...
+kind: Buildpack
+metadata:
+  name: my-custom-app-buildpack-0.3.1
+  namespace: my-kpack-using-namespace
+spec:
+  image: gcr.io/my-org/my-custom-app-buildpack:0.3.1
+  serviceAccountName: my-service-account
+```
+
+The status of Buildpack resource should include the same metadata to the ClusterStore. This is to enable kpack to calculate the exact expected digest of the Builder image without needing to communicate with the registry. Additionally, this metadata may be used by external tooling such as the [kp cli](https://github.com/pivotal/kpack). The status must include the id, buildpack api, diffID, digest, size, homepage, supported stacks, and order for [meta-buildpacks](https://buildpacks.io/docs/concepts/components/buildpack/#meta-buildpack) of each buildpack within the configured buildpackage image.
```
ncarlson, line 46: > This is to enable kpack to calculate the exact expected digest of the Builder image without needing to communicate with the registry.

Are there any potential scenarios in which an image digest calculated by kpack might differ from the digest reported by a registry? If so, could scenarios which are either probable or damaging be mitigated?

rfcs/0000-buildpack-resource.md
```diff
@@ -0,0 +1,218 @@
...
+spec:
+  tag: gcr.io/sample/builder
+  stack:
+    name: bionic-stack
+    kind: ClusterStack
+  order:
+  - group:
+    - id: paketo-buildpacks/java      
+```
+The `paketo-buildpacks/java` buildpack referenced in the Builder order would be sourced from the Buildpack resource within the `my-builder` namespace or the ClusterBuildpack resource that provides the `paketo-buildpacks/java` with the highest version number as defined by semver. Buildpacks IDs defined in the builder order may be a "implementation buildpack" within a meta-buildpack. If multiple Buildpack resources provide the same selected Buildpack the Builder should reconcile with an error.
```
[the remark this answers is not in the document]
matthewmcnew: Yes, this behavior actually already exists in kpack and is implemented by https://github.com/Masterminds/semver/
See: https://github.com/pivotal/kpack/blob/d9fe3e513eecd9a7bb8e483f6d582758ea25d991/pkg/cnb/store_buildpack_repository.go#L104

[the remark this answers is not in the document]
matthewmcnew: Buildpack Versions should adhere to Semver: https://github.com/buildpacks/rfcs/blob/main/text/0035-buildpack-versions.md

rfcs/0000-buildpack-resource.md
matthewmcnew: This behavior already exists in kpack so, we know that https://github.com/google/go-containerregistry can accurately calculate the digest of the builder image.

samj1912 (approved):

rfcs/0000-buildpack-resource.md
```diff
@@ -0,0 +1,218 @@
...
+
+These distinct personas may also exist in non-enterprise settings where buildpacks and builders have independent lifecycles. For example, the kpack build cluster is a multitenant kpack cluster that provides up-to-date dependencies for multiple teams by continually updating new paketo dependencies. The kpack components are built with a kpack specific namespace Builder that utilizes the [Paketo Go Buildpack](https://github.com/paketo-buildpacks/go) alongside a custom buildpack to build libgit2. The kpack namespace scoped Builder continually pulls in new versions of the Paketo Go buildpack. If the buildpack images were configured on the Builder resource the process that updated the Paketo Go Buildpack would also need to update the kpack Custom Builder. This would be unnecessarily complicated and may not even be possible if the cluster operator does not know all the uses of a Buildpack within the cluster.
```
ncarlson, line 218: >  This would be unnecessarily complicated and may not even be possible if the cluster operator does not know all the uses of a Buildpack within the cluster.

Good point. Thank you for enumerating the drawbacks of having the Builder reference the buildpack images.

ncarlson (approved):

tomkennedy513 (approved):
This is fantastic. Great work @matthewmcnew

stormqueen1990 (approved):
LGTM! 🎉

rfcs/0000-buildpack-resource.md
jjbustamante, line 218: Similar comment to this [one](https://github.com/pivotal/kpack/pull/931/files#r885700667). Does it mean CNB lifecycle or the Buildpack lifetime?

jjbustamante (approved):
Seems good to me!

rfcs/0000-buildpack-resource.md
```diff
@@ -0,0 +1,218 @@
...
+spec:
+  image: privateregistry.com/paketo-buildpacks/java:5.9.1
+  serviceAccountRef:
+    name: private-sa
+    namespace: private-ns
+```
+
+The status of the ClusterBuildpack should match the status of the Buildpack resource.
+
+##### ClusterStore Resources
```
mgibson1121, line 112: Will there be any migration concerns?  One problem that jumped to mind is that this may increase complexity for getting started compared to configuring a ClusterStore.  Although the user has to configure one less resource, they can pretty much copy and paste several lines of paketo buildpacks in one yaml file to configure the clusterstore.  From this RFC it seems like the user would have to configure a buildpack/clusterbuildpack for each buildpackage they want to reference in a builder/clusterbuilder.

mgibson1121 (approved):

pviraj59 (approved):

mvalliath (approved):

rfcs/0000-buildpack-resource.md
matthewmcnew: > One problem that jumped to mind is that this may increase complexity for getting started compared to configuring a ClusterStore

I can add this to the RFC but, I hope that this will actually help simplify onboarding. Currently, because the Clusterstore is a monolithic resource users essentially need to use the  `kp` cli and a streamlined bootrap experience would require a published descriptor to be used alongside the cli. However, a Buildpack resource enables the Buildpacks to be managed independently which means a descriptor is not needed. Instead users can directly apply a preconfigured set of Buildpack/Builder/Stack resources with `kubectl apply` to get started fast. (This does require this WIP RFC to be implemented as well https://github.com/pivotal/kpack/pull/932)

rfcs/0000-buildpack-resource.md
```diff
@@ -0,0 +1,218 @@
...
+    - name: my-custom-app-buildpack-0.3.1
+      kind: Buildpack
+```
+
+
+**Prior Art:**
+* [kapp controller](https://carvel.dev/kapp-controller/) models available Package versions as independent Package resources.
+
+**Risks:**
+This will require introducing a new api version within kpack. Future kpack releases will need to maintain support for ClusterStore resources until the kpack apis which contain the ClusterStore are removed.
```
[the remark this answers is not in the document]
matthewmcnew: This is really great question. I was assuming that ClusterStores can be used similar to ClusterBuildpacks as essentially a large deprecated ClusterBuildpack. Do you think this makes sense or the ClusterStore should be reference-able directly?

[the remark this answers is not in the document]
matthewmcnew: > Moreover, can we translate the old resource to the new one instead of having them both live side by side for a few releases?
I don't believe we have an easy way to translate them. Existing cluster with kpack installed may have the resources saved into etcd. We would still need to support those resources.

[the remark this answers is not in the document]
tylerphelan: I think the best migration path is:
- v1alpha3:
  - ClusterStores can be used in builders and clusterbuilders via store key
  - If store key is unset Buildpacks and ClusterBuildpacks are used
- v1alpha4
  - ClusterStores are removed entirely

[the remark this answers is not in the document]
yaelharel: @tylerphelan, regarding your suggestion about v1alpha3:
Are you ok that ClusterStores will be used even if store key is set?

Also, what will happen in v1alpha4 with the ClusterStores? It's a breaking change. What will be the process of upgrading to this version?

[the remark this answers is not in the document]
tylerphelan: > Are you ok that ClusterStores will be used even if store key is set?
I think this provides a migration path for the api change

> Also, what will happen in v1alpha4 with the ClusterStores? It's a breaking change. What will be the process of upgrading to this version?

I think this will be a true breaking change. We can keep the legacy behavior if we want but I supposed at some api version we should remove it.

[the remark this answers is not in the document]
matthewmcnew: > Are we sure it'll be a big problem to make a translation from ClusterStores to ClusterBuildpacks?

I think attempting to map and migrate existing ClusterStores to ClusterBuildpacks could be useful but, It doesn't seem like there is an obvious strategy for how to do the mapping. Any large api change like this needs to allow sometime for users to migrate so, I don't believe we could remove support for ClusterStores. I think *if and when* Clusterstores are formally removed would be a good time to explore migration tooling. Is it okay if we leave that outside the scope of this RFC?

yaelharel (approved):

tcdowney:
Just want to comment saying that I am excited for this update. As a user of kpack / "platform author" of platforms using kpack, these two points really resonate with me.

> It is common for a buildpack to be used within multiple Builders or ClusterBuilders on a kpack cluster. Externalizing the specification of a Buildpack location within one resource on the Cluster removes unnecessary duplication. Updating a buildpack can occur with the create/update of a single resource which will propagate out to multiple Builder resources on a cluster.

> Some kpack clusters may be operated in an environment where different personas manage available buildpacks from the persona that manages the builder config. This is especially likely in an enterprise environment where only blessed buildpacks are available to developers but, developers can modify their buildpack usage by owning namespace scoped Builders. This can be restricted via RBAC with a distinct Buildpack resource but, would not be possible if Builders directly referenced buildpack images as developers could then use any Buildpack.

We had a use case where one team managed a set of the "blessed" buildplacks for the platform and we wanted to provide a few ancillary buildpacks around them. We didn't want to just copy the "blessed" `ClusterStore` wholesale into our supertset `ClusterStore` because we wanted to be able to track and consume updates to those buildpacks as they happened. I was originally thinking we needed someway to specify `ClusterStore` on a per-buildpack basis in the `(Cluster)Builder`, but this solution is more elegant and should meet our needs. Thanks!

matthewmcnew:
Reopen.

pviraj59:
solves https://github.com/pivotal/kpack/issues/964

PR #934, opened 2022-03-09
Create kpack.io/services annotation when converting to v1alpha1 to prevent loss of data

tomkennedy513:


tylerphelan (approved):

matthewmcnew:
Could you explain how you came across this with kp?

tomkennedy513:
> Could you explain how you came across this with kp?

I basically applied a kubectl patch against the v1alpha2 resource and saw that services were no overwritten, but when applying that same patch against a v1alpha1 image resource, services were overwritten. When looking at our conversion logic I noticed that we throw away service bindings during conversions which caused any edits (even patches) to overwrite them

matthewmcnew (approved):

Issue #185, opened 2019-10-20
Include builder info in logs


drnic:
Currently the logs/init containers do not mention which builder name/tag nor sha256 were used for the build. Might help with CNB debugging to know/confirm which builder was used.

Currently:

```
[creds-init] {"level":"info","ts":1571606031.5970826,"logger":"fallback-logger","caller":"creds-init/main.go:40","msg":"Credentials initialized.","commit":"002a41a"}
[source-init] source-init:git.go:37: Successfully cloned "https://github.com/drnic/eirinix-pancake.git" @ "4df412cb9e8645e3dc9fd0fecadd1274637f6164" in path "/workspace"
```

dumez-k:
Proposed implementation:
```
Build reason(s): COMMIT
COMMIT:
	- 016fab8b41e0d8d520e23e1ff9d7e07d4b9afb9b
	+ c2ceeca17c6900ed585914fd00531b3cc031df35
Loading secret for "gcr.io" from secret "gcr" at location "/var/build-secrets/gcr"
Cloning "https://github.com/pivotal/kpack.git" @ "c2ceeca17c6900ed585914fd00531b3cc031df35"...
Successfully cloned "https://github.com/pivotal/kpack.git" @ "c2ceeca17c6900ed585914fd00531b3cc031df35" in path "/workspace"
building with:
    builder:
        image: gcr.io/cf-build-service-dev-219913/kenny/tbs111/base@sha256:c8a68c77611d46e45c7c4e04d88dfe0c4a12a833696777921a00f2dc8ca2ac77
        name: base
        kind: ClusterBuilder
===> DETECT
```

PR #944, opened 2022-03-30
Builder info in prepare log

jjbustamante:
This is a draft PR taking the work done by @dumez-k  and rebasing that code with the latest from the main branch. 

Fix: #185 

Solution:

This is how the message in shown in the log

Screen Shot 2022-04-07 at 5 59 15 PM


pkg/apis/build/v1alpha1/builder_resource.go
```diff
@@ -8,4 +8,5 @@ type BuilderResource interface {
 	Ready() bool
 	BuildpackMetadata() corev1alpha1.BuildpackMetadataList
 	RunImage() string
+	GetKind() string
```
tylerphelan, line 11: this should probably only need to be in the `v1alpha2` package

pkg/apis/core/v1alpha1/build_types.go
```diff
@@ -19,6 +19,7 @@ type BuildBuilderSpec struct {
 	// +patchStrategy=merge
 	// +listType
 	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty" patchStrategy:"merge" patchMergeKey:"name" protobuf:"bytes,15,rep,name=imagePullSecrets"`
+	Kind             string                        `"json:kind,omitempty"`
```
tylerphelan, line 22: This might not be needed for the build resource since the `name` `image` and `kind` are passed in to builds via flags?

tylerphelan: maybe `Name` needs to be here too

test/execute_build_test.go
```diff
@@ -531,6 +531,9 @@ func validateImageCreate(t *testing.T, clients *clients, image *buildapi.Image,
 	_, _, err = registryClient.Fetch(authn.DefaultKeychain, image.Spec.Tag)
 	require.NoError(t, err)
 
+	//TODO: Delete me:
```
tylerphelan, line 534: This could be a great place to integration test this feature

cmd/build-init/main.go
```diff
@@ -138,6 +142,8 @@ func main() {
 		logger.Fatalf("error while processing the project descriptor: %s", err)
 	}
 
+	logger.Printf("building with:\n Image:%s, Name:%s, Kind:%s ", *builderImage, *builderName, *builderKind)
```
tylerphelan, line 145: maybe this should explicitly include the term `Builder` instead of `building with` to be extra clear

pkg/apis/core/v1alpha1/build_types.go
```diff
@@ -19,6 +19,8 @@ type BuildBuilderSpec struct {
 	// +patchStrategy=merge
 	// +listType
 	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty" patchStrategy:"merge" patchMergeKey:"name" protobuf:"bytes,15,rep,name=imagePullSecrets"`
+	Kind             string                        `json:"kind,omitempty"`
```
matthewmcnew, line 22: I think this may unnecessarily confuse the Build Spec by implying that these fields have behavior. We already print info [logging from an an annotation](https://github.com/pivotal/kpack/blob/cecb3373638e6cc440bf7ae5d74c459987cfacad/pkg/apis/build/v1alpha2/build_pod.go#L382) can we use a similar strategy here?

pkg/apis/build/v1alpha1/builder_resource.go
matthewmcnew: Where is the `GetKind()` method called? Is it necessary?

jjbustamante:
Hi @matthewmcnew, @tylerphelan

Coming back to this issue, could you take a look at the latest approach I mentioned, I just want to get feedback on the approach and after that fix some problems with the unit tests.

cmd/build-init/main.go
```diff
@@ -147,6 +151,9 @@ func main() {
 		logger.Fatalf("error while processing the project descriptor: %s", err)
 	}
 
+	logger.Printf("Builder:\n Image: %s \n Name: %s \n Kind: %s ", *builderImage,
```
matthewmcnew, line 154: The build may be executed without this info. In that case it shouldn't log this message.

cmd/build-init/main.go
```diff
@@ -147,6 +151,11 @@ func main() {
 		logger.Fatalf("error while processing the project descriptor: %s", err)
 	}
 
+	if *builderImage != "" && *builderName != "" && *builderKind != "" {
```
matthewmcnew, line 154: Nice!

matthewmcnew (approved):

tylerphelan (approved):

ncarlson (commented):
Looks good!

Issue #945, opened 2022-04-01
Nil pointer error with lifecycle provider


tylerphelan:
https://kubernetes.slack.com/archives/C02UKAM5F2P/p1648765229200129

PR #946, opened 2022-04-04
Prevent possible Nil Pointer Reference when lifecycle error is overwritten

matthewmcnew:
Resolves https://github.com/pivotal/kpack/issues/945


matthewmcnew:
@jjbustamante

tylerphelan (approved):

ncarlson (approved):

PR #947, opened 2022-04-04
Invalid secret name

pviraj59:
Check invalid secret name  
Added tests


matthewmcnew:
This is to resolve: https://github.com/pivotal/kpack/issues/759?

go.mod
```diff
@@ -1,62 +1,330 @@
 module github.com/pivotal/kpack
 
-go 1.16
+go 1.17
```
matthewmcnew, line 3: Can we update the go version in a separate pr?

go.mod
```diff
@@ -1,62 +1,330 @@
...
 
 require (
-	cloud.google.com/go/kms v1.2.0 // indirect
 	github.com/BurntSushi/toml v1.0.0
 	github.com/Masterminds/semver/v3 v3.1.1
-	github.com/Microsoft/go-winio v0.5.2 // indirect
-	github.com/aryann/difflib v0.0.0-20170710044230-e206f873d14a
+	github.com/aryann/difflib v0.0.0-20210328193216-ff5ff6dc229b
```
matthewmcnew, line 8: Do we need to update deps for this PR?

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -849,8 +853,12 @@ func (b *Build) setupSecretVolumesAndArgs(secrets []corev1.Secret, filter func(s
 		}
 
 		volumeName := fmt.Sprintf(SecretTemplateName, secret.Name)
+		msgs := validation.IsDNS1123Label(volumeName)
```
matthewmcnew, line 856: Currently, the invalid volume name fails on pod creation leading to a reconcile error. This update will catch the error earlier but, will not lead to a successfully reconcile. kpack should successfully schedule a build with the provided secrets so, instead of returning an error we should generate a valid functioning volume name.

matthewmcnew (changes requested):
Can a valid volume name be constructed instead of returning an error?

pkg/apis/build/v1alpha2/build_pod.go
pviraj59: should we reconcile/ convert the invalid secret volume name to valid?
1) Would there be a concern that end users supplied volume name is converted to a valid name/ would users interact with volume name directly? 
2) If we change it to a valid name, should we output the new volume name?

matthewmcnew: The volume name here is just used an in implementation detail in the build pod. So, we can name the volume anything we like without interfering with the user's experience.

matthewmcnew: It might even make sense to use a generated volume name for each.

tomkennedy513: I think it would make sense to use a generated volume name to prevent any possible collision if the valid name we generate is the same as one that might be generated by a different secret name

tomkennedy513:
> Can a valid volume name be constructed instead of returning an error?

Could this potentially get us into a case where another secret is created that would result in a volume name that is now being used by a different secret

matthewmcnew:
> Could this potentially get us into a case where another secret is created that would result in a volume name that is now being used by a different secret

I think the solution is unique generated volume names?

tomkennedy513:
> > Could this potentially get us into a case where another secret is created that would result in a volume name that is now being used by a different secret
> 
> 
> 
> I think the solution is unique generated volume names? 

Like always add some unique string to volume? that would work

matthewmcnew:
> Like always add some unique string to volume? that would work

I think every volume could just have a unique name? Perhaps not even derived from the secret?

tylerphelan:
closing in favor of https://github.com/pivotal/kpack/pull/1023

PR #952, opened 2022-04-18
Improved error handling so that binding to a ProvisionedService object no longer swallows errors

ncarlson:
Fixed error handling on readProvisionedServiceDuckType. Added additional test coverage.

### DIFF
--- a/pkg/buildpod/generator_test.go
+++ b/pkg/buildpod/generator_test.go
@@ -3,6 +3,8 @@ package buildpod_test
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"knative.dev/pkg/apis/duck"
 	"testing"
 
 	"github.com/buildpacks/lifecycle/platform"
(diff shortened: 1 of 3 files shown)


pkg/buildpod/generator_test.go
```diff
@@ -316,17 +308,21 @@ func testGenerator(t *testing.T, when spec.G, it spec.S) {
...
+					SecretRef: &corev1.LocalObjectReference{Name: bindingSecret.Name},
+				},
+				&corev1alpha1.ServiceBinding{
+					Name:      ps.Name,
 					SecretRef: &corev1.LocalObjectReference{Name: psBindingSecret.Name},
 				},
 			}
 
-			assert.Len(t, build.buildPodCalls[0].BuildContext.Bindings, 1)
+			assert.Len(t, build.buildPodCalls[0].BuildContext.Bindings, 2)
```
matthewmcnew, line 325: I like the catch of the missing test!

pkg/buildpod/generator_test.go
```diff
@@ -301,13 +281,25 @@ func testGenerator(t *testing.T, when spec.G, it spec.S) {
 		})
 
 		it("passes in k8s service bindings if present", func() {
+			// Create provisioned service here b/c client wouldn't initialize with it
```
matthewmcnew, line 284: Could you explain, What happened here? Did it not work to initialize with the fakeDynamicClient?

ncarlson: That's right. For some reason, Tyler and I were unable to use the fakeDynamicClient to create the provisioned service under test.  At the moment, it's not clear to me why the fakeDynamicClient failed to meet our needs. More investigation will be done today.

pkg/buildpod/generator_test.go
ncarlson: Happy to help.

pkg/buildpod/generator_test.go
tylerphelan: maybe we should go over this

tylerphelan (approved):

matthewmcnew (approved):

PR #958, opened 2022-05-12
Adding activeDeadlineSeconds to build and pod spec

### LINKED ISSUE #866: Set a timeout on the builds
Currently it's possible for a malicious actor or an improper dependency downloaded by a buildpack to create a cluster DoS attack by hoarding resources and resulting in a never ending build. kpack should expose active deadline seconds as a way to kill builds that take longer than a certain amount.

daraghlowe:
**What are we doing?**
- Add a timeout on builds so they won't run infinitely
- Fixes #866


samj1912:
I believe we need both active deadline seconds and ttlSecondsAfterFinished. The former stops builds that are long running and the latter cleans up pods in terminal state.

We would also need to expose these fields in the image spec. We would also need to document what happens with the build controller if we introduce active deadline seconds and the pod is terminated before completion.

Lastly we also need to expose these fields in the image spec.

docs/image.md
```diff
@@ -12,6 +12,7 @@ The following defines the relevant fields of the `image` resource spec in more d
 - `builder`: Configuration of the `builder` resource the image builds will use. See more info [Builder Configuration](builders.md).
 - `serviceAccountName`: The Service Account name that will be used for credential lookup.
 - `source`: The source code that will be monitored/built into images. See the [Source Configuration](#source-config) section below.
+- `activeDeadlineSeconds`: Optional configurable max active time that the pod building the image can run for
```
samj1912, line 15: We should probably call this buildTimeout or similar. I think it would be easier to teach and use for end users rather than active deadline seconds in the image. It also doesn't make sense at the image spec level since the image resource itself doesn't have a deadline, rather the builds it creates. Thoughts?

daraghlowe: That makes sense to me regarding the changing the name to `buildTimeout` 👍🏻 

Do you have any ideas on a better place to put it than image spec?

samj1912: Ah, the latter sentence was about the active deadline seconds field itself. If we were to rename the image spec field to build timeout, I think that would be great and solve my concerns :)

daraghlowe: Updated now 👍🏻

docs/build.md
```diff
@@ -29,6 +29,7 @@ spec:
     git:
       url: https://github.com/buildpack/sample-java-app.git
       revision: main
+  activeDeadlineSeconds: 1800
```
ncarlson, line 32: Would a default value be useful? If so, what might the default be?

docs/build.md
```diff
@@ -62,6 +63,7 @@ spec:
 - `builder.image`: This is the tag to the [Cloud Native Buildpacks builder image](https://buildpacks.io/docs/using-pack/working-with-builders/) to use in the build. Unlike on the Image resource, this is an image not a reference to a Builder resource.    
 - `builder.imagePullSecrets`: An optional list of pull secrets if the builder is in a private registry. [To create this secret please reference this link](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#registry-secret-existing-credentials)
 - `source`: The source location that will be the input to the build. See the [Source Configuration](#source-config) section below.
+- `activeDeadlineSeconds`: Optional configurable max active time that the pod can run for
```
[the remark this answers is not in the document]
daraghlowe: We can definitely add one if you think it's a good idea? We picked 30 minutes for our deadline, as this makes sense for our own use of buildpacks. I don't know what a good number would be for all use cases.

docs/image.md
matthewmcnew: Could `activeDeadlineSeconds` go on the `spec.build` to indicate it is a build specific property?

docs/build.md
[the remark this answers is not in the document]
matthewmcnew: I like the idea of a default but, I imagine we would want to make the default configureable which is not something we currently support for any defaults. I would say that should be outside the scope of this PR.

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -329,6 +329,7 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
 		Spec: corev1.PodSpec{
 			// If the build fails, don't restart it.
 			RestartPolicy:     corev1.RestartPolicyNever,
```
tylerphelan, line 331: `gofmt` ?

tylerphelan (approved):

ncarlson (approved):

docs/image.md
daraghlowe: Updated @matthewmcnew 👍🏻

matthewmcnew (approved):

tomkennedy513 (approved):

matthewmcnew:
Thanks @daraghlowe !

PR #959, opened 2022-05-17
Use ggcr k8s keychain when authn with registry

tylerphelan:
- This enables use of credentials helpers in builds
- Move build metadata processing from controller to completion container
- Write build metadata to container termination path to be read by
controller


cmd/build-init/main.go
```diff
@@ -80,16 +82,16 @@ func main() {
...
+	creds, err := dockercreds.ParseBasicAuthSecrets(buildSecretsDir, basicDockerCredentials)
 	if err != nil {
 		logger.Fatal(err)
 	}
 
 	for _, c := range append(dockerCfgCredentials, dockerConfigCredentials...) {
 		credPath := filepath.Join(buildSecretsDir, c)
 
-		dockerCfgCreds, err := dockercreds.ParseDockerPullSecrets(credPath)
+		dockerCfgCreds, err := dockercreds.ParseDockerConfigSecret(credPath)
```
matthewmcnew, line 94: Good rename!

cmd/build-init/main.go
```diff
@@ -104,15 +106,22 @@ func main() {
...
 
-	err = dockercreds.VerifyWriteAccess(creds, *imageTag)
+	logger.Println("Loading cluster credential helpers")
+	k8sNodeKeychain, err := k8schain.NewNoClient(context.Background())
+	if err != nil {
+		logger.Fatal(err)
+	}
+	keychain := authn.NewMultiKeychain(creds, k8sNodeKeychain)
+
+	err = dockercreds.VerifyWriteAccess(keychain, *imageTag)
```
matthewmcnew, line 116: Super minor but, would it make sense to inline keychain to call out the limited scope of this keychain?

cmd/completion/main.go
```diff
@@ -28,25 +36,23 @@ const (
...
 var (
+	cacheTag                string
 	notaryV1URL             string
 	dockerCredentials       flaghelpers.CredentialsFlags
 	dockerCfgCredentials    flaghelpers.CredentialsFlags
 	dockerConfigCredentials flaghelpers.CredentialsFlags
 	cosignAnnotations       flaghelpers.CredentialsFlags
 	cosignRepositories      flaghelpers.CredentialsFlags
 	cosignDockerMediaTypes  flaghelpers.CredentialsFlags
-	basicGitCredentials     flaghelpers.CredentialsFlags
```
matthewmcnew, line 38: I believe the same flags from build-init are used on completion container for simplicity and these flags were added to support those flags without erroring.

cmd/completion/main.go
```diff
@@ -57,56 +63,112 @@ func init() {
...
+	builtImageRef := fmt.Sprintf("%s@%s", report.Image.Tags[0], report.Image.Digest)
+
+	k8sNodeKeychain, err := k8schain.NewNoClient(context.Background())
 	if err != nil {
-		return errors.Wrap(err, "toml decode")
+		log.Fatal(err)
 	}
 
-	creds, err := dockercreds.ParseMountedAnnotatedSecrets(registrySecretsDir, dockerCredentials)
+	creds, err := dockercreds.ParseBasicAuthSecrets(registrySecretsDir, dockerCredentials)
```
matthewmcnew, line 83: It is unfortunate we still need to do this vs just using the same <home>/.docker/config.json that the lifecycle uses.

cmd/completion/main.go
```diff
@@ -57,56 +63,112 @@ func init() {
...
+		ImageFetcher: &registry.Client{},
+	}
 
-		homeDir, err := os.UserHomeDir()
+	cacheImageRef := ""
+	if cacheTag != "" {
+		cacheImageRef, err = metadataRetriever.GetCacheImage(cacheTag)
 		if err != nil {
-			return errors.Wrapf(err, "error obtaining home directory")
+			log.Fatal(err)
```
matthewmcnew, line 123: While we are here. Could we make this non-fatal? To protect from this:https://github.com/buildpacks/lifecycle/issues/830?

cmd/completion/main.go
```diff
@@ -3,20 +3,28 @@ package main
 import (
 	"context"
 	"flag"
+	"fmt"
```
matthewmcnew, line 6: There is alot in there these days. Might make sense to explore moving it out of main so it can be better tested.

cmd/completion/main.go
matthewmcnew: Actually, if we made this nonfatal we could have it always check for a cache image and perhaps simplify the build pod generation.

cmd/completion/main.go
```diff
@@ -57,56 +63,112 @@ func init() {
...
+	data, err := compressor.Compress(buildStatusMetadata)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if err := ioutil.WriteFile(buildapi.CompletionTerminationMessagePath, []byte(data), 0666); err != nil {
+		log.Fatal(err)
+	}
+
+	if hasCosign() || notaryV1URL != "" {
```
matthewmcnew, line 150: Because cosign uses the docker home keychain it can't use k8s crendential helpers. Correct?

pkg/cnb/cnb_metadata.go
```diff
@@ -1,74 +1,37 @@
...
 	"github.com/google/go-containerregistry/pkg/name"
 	ggcrv1 "github.com/google/go-containerregistry/pkg/v1"
-	"github.com/pkg/errors"
-	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	buildapi "github.com/pivotal/kpack/pkg/apis/build/v1alpha2"
-	"github.com/pivotal/kpack/pkg/registry"
 	"github.com/pivotal/kpack/pkg/registry/imagehelpers"
+	"github.com/pkg/errors"
```
matthewmcnew, line 12: Move this dep up?

pkg/cosign/image_signer.go
```diff
@@ -92,7 +92,7 @@ func (s *ImageSigner) sign(ctx context.Context, refImage, secretLocation, cosign
 	if err := s.signFunc(
 		ctx,
 		ko,
-		options.RegistryOptions{},
+		options.RegistryOptions{KubernetesKeychain: true},
```
matthewmcnew, line 95: Oh nice.

cmd/completion/main.go
matthewmcnew: Nevermind. https://github.com/pivotal/kpack/pull/959/files#r875069531

pkg/reconciler/build/build.go
```diff
@@ -61,21 +59,21 @@ func NewController(opt reconciler.Options, k8sClient k8sclient.Interface, inform
...
+	Client               versioned.Interface
+	Lister               buildlisters.BuildLister
+	K8sClient            k8sclient.Interface
+	PodLister            v1Listers.PodLister
+	PodGenerator         PodGenerator
+	metadataDecompressor metadataDecompressor
 }
 
-func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
+func (r *Reconciler) Reconcile(ctx context.Context, key string) error {
```
matthewmcnew, line 70: For consistency all reconcilers have a `c` receiver. Might make sense to switch them to `r` but, can we do that on its own?

pkg/reconciler/build/build_test.go
```diff
@@ -24,14 +22,9 @@ import (
 	corev1alpha1 "github.com/pivotal/kpack/pkg/apis/core/v1alpha1"
 	"github.com/pivotal/kpack/pkg/buildpod"
 	"github.com/pivotal/kpack/pkg/client/clientset/versioned/fake"
-	"github.com/pivotal/kpack/pkg/cnb"
-	"github.com/pivotal/kpack/pkg/reconciler/build"
-	"github.com/pivotal/kpack/pkg/reconciler/build/buildfakes"
 	"github.com/pivotal/kpack/pkg/reconciler/testhelpers"
 )
 
-//go:generate counterfeiter . MetadataRetriever
```
matthewmcnew, line 33: Nice!

cmd/build-init/main.go
ncarlson: What do you mean by "call out"?

pkg/apis/build/v1alpha2/image_builds.go
```diff
@@ -108,7 +108,6 @@ func lastBuild(latestBuild *Build) *LastBuild {
 
 	return &LastBuild{
 		Image:   latestBuild.BuiltImage(),
-		Cache:   BuildCache{Image: latestBuild.CacheImage()},
```
matthewmcnew, line 111: If this is not passed through. Can we remove the whole reading of the cache image completely?

cmd/build-init/main.go
matthewmcnew: Make it limited scope clear. This keychain is only used in this location but, that is unclear.

cmd/completion/main.go
```diff
@@ -28,25 +36,23 @@ const (
...
-	sshGitCredentials       flaghelpers.CredentialsFlags
 	logger                  *log.Logger
 )
 
 func init() {
+	flag.StringVar(&cacheTag, "cache-tag", os.Getenv(buildapi.CacheTagEnvVar), "Tag of image cache")
 	flag.StringVar(&notaryV1URL, "notary-v1-url", "", "Notary V1 server url")
 	flag.Var(&dockerCredentials, "basic-docker", "Basic authentication for docker of the form 'secretname=git.domain.com'")
 	flag.Var(&dockerCfgCredentials, "dockercfg", "Docker Cfg credentials in the form of the path to the credential")
 	flag.Var(&dockerConfigCredentials, "dockerconfig", "Docker Config JSON credentials in the form of the path to the credential")
```
ncarlson, line 55: What is the difference between these two flags? Is a `Docker Cfg` file different from a `Docker Config Json` file?

matthewmcnew: Yes. Two different types of docker secrets: https://kubernetes.io/docs/concepts/configuration/secret/#secret-types

cmd/completion/main.go
```diff
@@ -57,56 +63,112 @@ func init() {
...
 
-	if hasCosign() || notaryV1URL != "" {
-		err := signImage()
-		if err != nil {
-			log.Fatal(err)
-		}
+	var report platform.ExportReport
+	_, err := toml.DecodeFile(reportFilePath, &report)
+	if err != nil {
+		log.Fatal(err, "toml decode")
```
ncarlson, line 69: How about `error decoding report toml file`?

pkg/cosign/image_signer.go
tylerphelan: right?

hack/tools.go
```diff
@@ -1,4 +1,4 @@
-// +build tools
+//go:build tools
```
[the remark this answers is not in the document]
tylerphelan: `goimports` did this so I went with it

pkg/reconciler/build/metadata.go
```diff
@@ -0,0 +1,63 @@
...
+
+	corev1alpha1 "github.com/pivotal/kpack/pkg/apis/core/v1alpha1"
+)
+
+type BuildStatusMetadata struct {
+	BuildpackMetadata corev1alpha1.BuildpackMetadataList `json:"buildpackMetadata"`
+	LatestImage       string                             `json:"latestImage"`
+	LatestCacheImage  string                             `json:"latestCacheImage"`
+	StackRunImage     string                             `json:"stackRunImage"`
+	StackID           string                             `json:"stackID"`
```
ncarlson, line 18: Would there be value in sorting these fields alphabetically?

cmd/completion/main.go
tylerphelan: so confusing right!

ncarlson (approved):

api/openapi-spec/swagger.json
```diff
@@ -4550,12 +4550,15 @@
           "type": "string"
         },
         "metadata": {
+          "default": {},
```
matthewmcnew, line 4553: Is this a change in the version that created this file?

cmd/completion/main.go
tylerphelan: I'm a little confused by this concern, could you expand on the 
>support those flags without erroring

matthewmcnew:
I know this is not the main feature here but, very excited to finally use report.toml and kill a rare but serious race condition

cmd/completion/main.go
tylerphelan: a lot of formatting?

cmd/completion/main.go
matthewmcnew: All the build-init flags are passed through to the completion. So, the completion container would error on the unknown `git` flags.

cmd/completion/main.go
matthewmcnew: Oh. I just picked the fist line. I meant the whole file itself.

cmd/completion/main.go
tylerphelan: oh I see

pkg/apis/build/v1alpha2/image_builds.go
tylerphelan: This was an oversight, I guess there are not tests

tylerphelan: I added it back

cmd/completion/main.go
tylerphelan: attempted to improve this

cmd/completion/main.go
tylerphelan: I made a pass at improving this but maybe we can do better

cmd/completion/main.go
```diff
@@ -57,56 +64,77 @@ func init() {
...
+		log.Fatal(err)
+	}
+
+	if hasCosign() || notaryV1URL != "" {
+		if err := signImage(report, keychain); err != nil {
+			log.Fatal(err)
 		}
 	}
 
+	logger.Println("Build successful")
```
matthewmcnew, line 134: The one line of this file that stays the same.

pkg/cnb/build_metadata.go
```diff
@@ -0,0 +1,189 @@
...
+		bpMetadata = append(bpMetadata, corev1alpha1.BuildpackMetadata{
+			Id:       metadata.ID,
+			Version:  metadata.Version,
+			Homepage: metadata.Homepage,
+		})
+	}
+	return bpMetadata
+}
+
+func CompressBuildMetadata(metadata *BuildMetadata) ([]byte, error) {
```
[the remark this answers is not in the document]
matthewmcnew: If it is truncated, I assume the `DecompressBuildMetadata` will error throwing a reconcile loop?

pkg/cnb/build_metadata.go
```diff
@@ -0,0 +1,189 @@
...
+	"io/ioutil"
+
+	lifecyclebuildpack "github.com/buildpacks/lifecycle/buildpack"
+	"github.com/buildpacks/lifecycle/platform"
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	ggcrv1 "github.com/google/go-containerregistry/pkg/v1"
+	corev1alpha1 "github.com/pivotal/kpack/pkg/apis/core/v1alpha1"
+	"github.com/pivotal/kpack/pkg/registry/imagehelpers"
+	"github.com/pkg/errors"
```
matthewmcnew, line 18: Seperate pkg/errors as it is not a kpack import?

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -799,13 +802,28 @@ func (b *Build) rebasePod(buildContext BuildContext, images BuildPodImages) (*co
 	}, nil
 }
 
+func (b *Build) completionContainer(images BuildPodImages, buildContext BuildContext, args []string, env []corev1.EnvVar, volumeMounts []corev1.VolumeMount) corev1.Container {
```
matthewmcnew, line 805: Can we inline this method now that it doesn't have any special logic?

pkg/cnb/build_metadata.go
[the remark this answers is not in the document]
tylerphelan: good call!

pkg/apis/build/v1alpha2/build_pod.go
tylerphelan: I changed it a bit but I don't think it's worth putting back as it is in 2 places and has a little extraction value

matthewmcnew (approved):

pkg/apis/build/v1alpha2/build_pod.go
matthewmcnew: Fair. I think there is a bit of stylistic discussion to have here. All the other container specs are defined inline here to make it "easy" to see the defined pod all at once as opposed to navigating into methods. We have gone back and force on the completion container a few times here.

PR #960, opened 2022-05-18
Document how to run unit tests and how to install the libgit2 v1.3.0 dependency

ncarlson:


docs/local.md
```diff
@@ -9,6 +9,10 @@ kubectl cluster-info # ensure you have access to a cluster
 
 ### Running Tests
 
+```bash
+go test ./...
```
tylerphelan, line 13: we should split the unit tests and integration tests. so something like:
```
Running Unit Tests

go test ./pkg/...

Running the end-to-end tests

go test ./test/...
```

docs/local.md
```diff
@@ -21,4 +25,44 @@ export KPACK_TEST_NAMESPACE_LABELS="istio-injection=disabled,purpose=test"
...
\ No newline at end of file
+```
+
+### libgit2 v1.3.0 dependency and installation
+
+Several unit tests depend upon libgit2 v1.3.0.
+
+__macOS Installation Instructions (Intel):__
+
+1. Install `cmake` and `pkg-config`
```
tylerphelan, line 36: I think on mac libgit2 can be installed with `brew install libgit2@1.3.0`

tylerphelan: nevermind! it looks like it's been removed.

PR #961, opened 2022-05-24
Improve logging with CR kind being reconciled

tylerphelan:
- Add logging-config to configure logging

### DIFF
--- a/config/config-logging.yaml
+++ b/config/config-logging.yaml
@@ -1,0 +1,31 @@
+# Copyright 2020 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-logging
+  namespace: kpack
+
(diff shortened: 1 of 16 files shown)


config/config-logging.yaml
```diff
@@ -0,0 +1,31 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-logging
+  namespace: kpack
+
+data:
+  zap-logger-config: |
```
matthewmcnew, line 11: Nice!

config/config-logging.yaml
```diff
@@ -0,0 +1,31 @@
...
+    {
+      "level": "info",
```
ncarlson, line 13: Is this log-level overridable? If not, could be a future enhancement.

ncarlson (approved):

PR #963, opened 2022-05-27
Some changes to the kp tutorial

yaelharel:

### DIFF
--- a/docs/tutorial_kp.md
+++ b/docs/tutorial_kp.md
@@ -20,7 +20,7 @@ build an OCI image from source and allow kpack rebuild the OCI image with update
 
     ```bash
     docker login <REGISTRY-HOSTNAME> -u <REGISTRY-USER>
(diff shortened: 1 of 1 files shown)


docs/tutorial_kp.md
```diff
@@ -20,17 +20,15 @@ build an OCI image from source and allow kpack rebuild the OCI image with update
 
     ```bash
     docker login <REGISTRY-HOSTNAME> -u <REGISTRY-USER>
-    kp secret save kp-default-registry-creds \
-       --registry <REGISTRY-HOSTNAME> \
-       --registry-user <REGISTRY-USER> \
-       -n kpack
+    kp secret create kp-default-registry-creds \
```
tylerphelan, line 23: by default this should probably still be `--registry` instead of `--gcr` because we shouldn't assume everyone is using gcr.

tylerphelan (approved):
nice!

PR #967, opened 2022-06-08
Fix windows build metadata functionality

tylerphelan:
- Termination message path is not supported for docker windows runtime
- For windows builds, metadata must be pulled from the registry


cmd/completion/main.go
```diff
@@ -121,6 +127,10 @@ func main() {
 		log.Fatal(err)
 	}
 
+	if err := os.MkdirAll(filepath.Dir(buildapi.CompletionTerminationMessagePath), 0777); err != nil {
```
matthewmcnew, line 130: Was this the result of an issue?

matthewmcnew (approved):

cmd/completion/main.go
matthewmcnew: Also, does the `cnb` user have permission to write to `/dev/termination-log`

ncarlson (approved):

Issue #957, opened 2022-05-12
Build from Zip Blob with Unknown Parent Folder Name


diarmuidie:
We're using the build resource directly to build from a zipped source:
```
source:
  blob:
    url: "https://example.com/source.zip"
  subPath: ""
```

When unzipped that `source.zip` file has a parent directory and within that is the source of the application. Because the zip is coming from a third party (Github, or direct from the our end users in our case) the name of the parent folder is not consistent enough that we can use the `subPath` to set it.

For example if you download the source of a Github repo using the [Github zipball api endpoint](https://docs.github.com/en/rest/repos/contents#download-a-repository-archive-zip) you get a different parent folder name depending on if it's a public (`{org_name}-{repo_name}-{short_sha}`) or private (`{org_name}-{repo_name}-{long_sha}`) repo.

One way to work around this if there was an option to set something like the tar `--strip-components` so we could tell the [archive package](https://github.com/pivotal/kpack/blob/main/pkg/archive/archive.go) to ignore the first N paths when extracting (this is what [Paketos Libpak extraction package does](https://github.com/paketo-buildpacks/libpak/blob/8979ae3422306d7bf5a10c136bd187807b4e744d/crush/crush.go#L240))

Do you have any other thoughts or ideas on how to solve this?

Thanks!

matthewmcnew:
That is an interesting problem! Providing `strip-components` would be an interesting addition to the `source.blob` kpack api. In the meantime would programmatically calculating the subpath from the blob url be possible? Perhaps from an earlier automation job?

diarmuidie:
Hey @matthewmcnew, ya for now we're looking at having a component re-format the zip into a standard format before it gets to kpack, but it would be nice not to have to do that! Thanks

PR #976, opened 2022-06-15
Add Ability to Strip Components from Blob Files

diarmuidie:
### What:

This change adds a new field to the blob source that lets you specify the number of base directory components you want to strip from the path when extracting the files. This is similar in functionality to the the tar `--strip-components` argument.

The field will default to `0` so that if it is not set the existing build functionality is unchanged.

### Why:

You might be receiving an archive blob from a third party in your build, where you have no prior knowledge of the parent directory name so you can't use the `subPath` option to specify the root. For example when using [Githubs archive API](https://docs.github.com/en/rest/repos/contents#download-a-repository-archive-tar) the parent directory's name depends on if the repo is public or private, using Github Enterprise server or not etc. With this change you don't need to know the details but can tell kpack to strip 1 component when extracting and that will put the working dir for the build in the root of the repo archive.

Fixes: https://github.com/pivotal/kpack/issues/957


tylerphelan (approved):
nice! thank you!

ncarlson (approved):
Looks good! Thank you for the contribution.

jjbustamante (approved):
Looks good! 
I also ran the test locally and everything is fine. I just did it because the job is pending on github.

Great job @diarmuidie !

diarmuidie:
> Looks good! I also ran the test locally and everything is fine. I just did it because the job is pending on github.
> 
> Great job @diarmuidie !

Thanks @jjbustamante . Looks like the tests aren't running because I'm a first time contributor:
> 1 workflow awaiting approval
> First-time contributors need a maintainer to approve running workflows

ncarlson (approved):

tomkennedy513 (approved):

matthewmcnew (approved):

matthewmcnew:
Thanks so much @diarmuidie !

PR #975, opened 2022-06-15
chore: bump cosign dependencies

hectorj2f:
Signed-off-by: hectorj2f <hectorf@vmware.com>

Bumped cosign dependencies to latest version.

PR #977, opened 2022-06-15
go.mod updates to allow using in other modules

kieron-dev:
The k8s.io/client-go version in go.mod was accidentally bumped to 1.5.2 which is marked as retracted. The correct versions are current in the 0.24.x range. For some reason, the replace on client-go stops go issuing a warning about this.

This change allows kpack to be included in other go projects without causing problems with the client-go dependency

cc @georgethebeatle

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	golang.org/x/sync v0.0.0-20220601150217-0de741cfad7f
 	k8s.io/api v0.24.1
 	k8s.io/apimachinery v0.24.1
-	k8s.io/client-go v1.5.2
+	k8s.io/client-go v0.24.1
 	k8s.io/code-generator v0.23.6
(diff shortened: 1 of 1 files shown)


kieron-dev:
We've just seen that #975 beat us to this by a couple of hours! However, please still consider this PR after merging that one, since the other changes are still important for us to be able to bump kpack in korifi.

tylerphelan:
can this be rebased? we might have included all the necessary dependencies now.

kieron-dev:
@tylerphelan - yes this did benefit from a rebase.

The only problem that wasn't now fixed in main was the client-go version being set to the retracted v1.5.2 version. So I've reworded this PR and pruned it down to simply fixing that client-go version.

tylerphelan:
the k8s client libraries are currently intentionally pinned to 0.20.x. We will bump these all of them asap!

davewalter:
I think this PR can be closed now that #1007 has been merged.

PR #988, opened 2022-07-11
Adding run-image reference to the BuildSpec to be use during pod rebase

### LINKED ISSUE #641: Churn bug when clusterstack run image digest is out of date with tagged run image
When clusterstack run image is not fully qualified and is instead a soft pointer like `paketo-buildpacks/run` and the last reconcile of the clusterbuilder is out of date with the current tagged image on dockerhub, then an infinite number of builds are scheduled as kpack is trying to get a build to produce an image with the old run image. 

To reproduce:
Create a stack with spec.image.run-image pointing to a run image with tag
Create a builder and build with this stack
Update image ref by tag
Trigger a new build

Here are also some other [reproduction steps](https://github.com/wind57/kpack) reported in issue #979

jjbustamante:
### Solution

We implemented the suggested option 2

- Specify the run image with fully qualified digest that should be used for builds and rebases

### Summary

This Pull Request is adding the `run-image` reference into the *BuildSpec*,  currently the reference is extracted from the [BuilderResource](https://github.com/pivotal/kpack/blob/main/pkg/apis/build/v1alpha2/builder_resource.go) because it contains the digest of the image. 

Fixes #641


pkg/apis/build/v1alpha2/build_types.go
```diff
@@ -56,6 +56,7 @@ type BuildSpec struct {
 	ServiceAccountName string                        `json:"serviceAccountName,omitempty"`
 	Source             corev1alpha1.SourceConfig     `json:"source"`
 	Cache              *BuildCacheConfig             `json:"cache,omitempty"`
+	RunImage           string                        `json:"runImage,omitempty"`
```
tomkennedy513, line 59: To line up with our other crds, maybe this can be a struct with a nested `image` field like we have in the [clusterstack spec](https://github.com/pivotal/kpack/blob/33d72b29eed4afa7ce4830783e4b6ac24deddbbb/pkg/apis/build/v1alpha2/cluster_stack_types.go#L32-L33) and the [clusterstore spec](https://github.com/pivotal/kpack/blob/33d72b29eed4afa7ce4830783e4b6ac24deddbbb/pkg/apis/build/v1alpha2/cluster_store_types.go#L32)

pkg/apis/build/v1alpha2/image_builds.go
```diff
@@ -58,6 +58,7 @@ func (im *Image) Build(sourceResolver *SourceResolver, builder BuilderResource,
 		Spec: BuildSpec{
 			Tags:                  im.generateTags(buildNumber),
 			Builder:               builder.BuildBuilderSpec(),
+			RunImage:              builder.RunImage(),
```
jjbustamante, line 61: Based on [comment](https://github.com/pivotal/kpack/pull/988/files#r918365344):

```golang
RunImage: BuildSpecImage{
    Image: builder.RunImage(),
},
```

pkg/apis/build/v1alpha2/build_types.go
matthewmcnew: Yes! All images in kpack spec fields should have the key `image`. I think @tylerphelan suggested maybe putting the run image it in the builder spec BuildBuilderSpec?

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -818,7 +822,7 @@ func (b *Build) rebasePod(buildContext BuildContext, images BuildPodImages) (*co
 					Resources: b.Spec.Resources,
 					Args: args(a(
 						"--run-image",
-						buildContext.BuildPodBuilderConfig.RunImage,
+						runImage,
```
matthewmcnew, line 825: I think we also want to pass the run image until the non-rebase build.

jjbustamante: Yeah, I didn't do it just to get early feedback on it, but I will add it

pkg/apis/build/v1alpha2/image_builds.go
tomkennedy513: Ya this is exactly what I had in mind. Maybe we could reuse the spec from the cluster stack and move it to core api. No strong preference though.

pkg/apis/build/v1alpha2/build_types.go
jjbustamante: If I remember from the meeting last Friday, the current BuildBuilderSpec has the following structure:

```golang
type BuildBuilderSpec struct {
	Image string `json:"image,omitempty"`
	// +patchMergeKey=name
	// +patchStrategy=merge
	// +listType
	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty" patchStrategy:"merge" patchMergeKey:"name" protobuf:"bytes,15,rep,name=imagePullSecrets"`
}
```

We already have an `Image` field there, adding it there requires to refactor a little bit that structure, right? and I think that's a change that can make some troubles to current users.

tylerphelan: I like:
`Build.spec.runImage.image`

pkg/apis/build/v1alpha2/build_pod.go
jjbustamante: Working on this, but got the following error on my first attempt:

Screen Shot 2022-07-12 at 12 28 04 PM

tylerphelan: this is suggesting that the value used for the `-run-image` flag argument is unset

tylerphelan: it appears `-run-image` was moved from exporter to analyzer in platform api version 0.7: https://buildpacks.io/docs/reference/spec/migration/platform-api-0.6-0.7/

Maybe we don't pass the run image to the non-rebase builds for now until we have implemented this

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -245,6 +250,14 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
 		Resources: b.Spec.Resources,
 		Args: args(
 			[]string{"-layers=/layers", "-analyzed=/layers/analyzed.toml"},
+			func() []string {
+				if !platformAPILessThan07 {
+					if b.Spec.RunImage.Image != "" {
```
matthewmcnew, line 255: Maybe we should drop platform api support

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -566,20 +579,38 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
...
+							[]corev1.EnvVar{
+								homeEnv,
+								{
+									Name:  platformAPIEnvVar,
+									Value: platformAPI.Original(),
+								},
 							},
-						},
+							func() corev1.EnvVar {
+								if platformAPILessThan07 {
```
matthewmcnew, line 606: Is there any harm in always having this env var?

matthewmcnew (approved):

pkg/apis/build/v1alpha2/build_pod.go
tylerphelan: below 0.7?

tylerphelan, line 255: I think the variable can be used here and elsewhere (`runImage`).

Regardless of where our source of truth `runImage` came from we can still pass it through in all these places (even though we didn't before)

pkg/apis/build/v1alpha2/image_builds.go
```diff
@@ -58,6 +58,9 @@ func (im *Image) Build(sourceResolver *SourceResolver, builder BuilderResource,
 		Spec: BuildSpec{
 			Tags:                  im.generateTags(buildNumber),
 			Builder:               builder.BuildBuilderSpec(),
+			RunImage:              BuildSpecImage{
```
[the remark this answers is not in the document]
jjbustamante: Yes, I just run `go fmt` and fixed them

pkg/apis/build/v1alpha2/build_pod.go
jjbustamante: @tylerphelan 
you mean to change it to:
```golang
if !platformAPILessThan07 {
   return []string{"-run-image=" + runImage}
}
```
rigth?

pkg/apis/build/v1alpha2/build_pod.go
jjbustamante: I don't think so.

pkg/apis/build/v1alpha2/build_pod.go
tylerphelan: right

tylerphelan: and `if b.Spec.RunImage.Image != ""`

jjbustamante: Fixed on [commit](https://github.com/pivotal/kpack/pull/988/commits/c190e3f3e8744b25ba10210409e1ce3ca58013c0)

pkg/apis/build/v1alpha2/build_pod.go
jjbustamante: Fixed on [commit](https://github.com/pivotal/kpack/pull/988/commits/c190e3f3e8744b25ba10210409e1ce3ca58013c0)

tomkennedy513 (approved):

matthewmcnew:
@jjbustamante Could you update this branch with main?

jjbustamante:
> @jjbustamante Could you update this branch with main?

Done!

PR #990, opened 2022-07-12
Set logrus log level to fatal

tylerphelan:
- This is done in all binaries that might interact with a registry
- The goal is to reduce the noisiness of the github.com/awslabs/amazon-ecr-credential-helper package

The ecr helper has not yet released a configurable logger: https://github.com/awslabs/amazon-ecr-credential-helper/releases/tag/v0.6.0

The implementation idea is to set the global logrus log level to `Fatal` so only fatal or panic errors are logged from the ecr cred helper package

The risk is that this would configure all unconfigured logrus loggers to also be set to the Fatal log level. I think this is acceptable because it is not used in our code. It is only notably used in the notary package which does have configurable logger which we set.

### DIFF
--- a/cmd/build-init/main.go
+++ b/cmd/build-init/main.go
@@ -17,6 +17,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/authn/k8schain"
 	"github.com/pkg/errors"
 
+	_ "github.com/pivotal/kpack/internal/logrus/fatal"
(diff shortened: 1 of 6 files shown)


matthewmcnew (approved):

matthewmcnew:
So good!

tomkennedy513 (approved):

tomkennedy513:
Nice!

Issue #554, opened 2020-11-19
Use value from secrets in environment variables


cirocosta:
Hi!

While going through the docs and the types for the Build CRD, I got the impression that given that the `corev1.EnvVar` type is used, I could leverage `valueFrom: {secretKeyRef: <>}` to fill that variable with a value from a secret, but as those references are never resolved before the marshalling (https://github.com/pivotal/kpack/blob/3b1a44a1b15c8e0036e054718a9edd1b0695ac00/pkg/apis/build/v1alpha1/build_pod.go#L98-L101), we end up with just the pure json representation of it:

```
      - name: PLATFORM_ENV_VARS
        value: '[{"name":"FOO","value":"bar},{"name":"CAZ","valueFrom":{"secretKeyRef":{"name":"zaz","key":"zaz"}}]'
```

_(^ from the pod spec)_

It makes sense to not have the values in plain text there in the json, but I wonder if it'd be interesting to have those refs dereferenced when writing them down to `platform/env/<env_key>`.

I don't _really_ have a use-case for it that can't be covered through bindings, but, it seemed a bit confusing to me to not have that ability given the type used - perhaps the action here would be an update in the docs?

thank you!

matthewmcnew:
It seems reasonable to expect the value from format to be accepted.

matthewmcnew:
Closed in https://github.com/pivotal/kpack/issues/554

PR #997, opened 2022-07-27
Accept all types of corev1.EnvVar as build env vars

kieron-dev:
This allows consumers of the kpack Image resource to specify environment variables to pass to the build process in all formats supported by kubernetes. Previously, only literal string values were accepted.

To do this, we set the supplied env vars on the prepare init container, using a prefix to identify them. Kubernetes transparently fetches values from secrets or however else the env vars are defined, and provides the literal values in the container's environment.

Finally, the prepare init container code fetches those env vars with the given prefix, and serializes them to the env directory as before.

It fixes this issue: https://github.com/pivotal/kpack/issues/554

### DIFF
--- a/cmd/build-init/main.go
+++ b/cmd/build-init/main.go
@@ -29,9 +29,8 @@ import (
 )
 
 var (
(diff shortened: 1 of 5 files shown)


matthewmcnew (approved):

matthewmcnew:
Thank you this looks great!

tomkennedy513 (approved):

PR #999, opened 2022-07-28
wip: istio support rfc

tomkennedy513:


rfcs/0000-execute-builds-in-standard-containers.md
```diff
@@ -0,0 +1,31 @@
...
+**Complexity:**
+
+**Prior Art:**
+Tekton: `https://github.com/tektoncd/pipeline/tree/main/cmd/entrypoint#waiting-for-sidecars`
+kpack spike: `https://github.com/pivotal/kpack/tree/execute-builds-in-regular-containers`
+
+**Alternatives:**
+
+* We can build the Tekton entrypoint code as an image to use in kpack
+* We could execute all build steps in one standard container similar to pack build with the `--trust-builder` flag
```
matthewmcnew, line 27: How would this work with the prepare step?

rfcs/0000-execute-builds-in-standard-containers.md
```diff
@@ -0,0 +1,31 @@
...
+**Prior Art:**
+Tekton: `https://github.com/tektoncd/pipeline/tree/main/cmd/entrypoint#waiting-for-sidecars`
+kpack spike: `https://github.com/pivotal/kpack/tree/execute-builds-in-regular-containers`
+
+**Alternatives:**
+
+* We can build the Tekton entrypoint code as an image to use in kpack
+* We could execute all build steps in one standard container similar to pack build with the `--trust-builder` flag 
+
+**Risks:**
```
matthewmcnew, line 29: Might be worth calling out the impact on logging.

rfcs/0000-execute-builds-in-standard-containers.md
tomkennedy513: My thinking was that we would create our own binary that runs the prepare step and then execs the lifecycle steps inside one container. I don't intend for this to be a really viable solution but I was just trying to call out that it is the other end of the spectrum of solutions

samj1912:
@tomkennedy513 / @matthewmcnew this will have quite a bit of an adverse effect on the resource utilization of the final build pod. Currently the resource utilization is calculated by max(max(init containers), sum(containers)). This change will effectively lead to each build pod consuming 5x the resources due to each container's requested cpu and memory.

tomkennedy513:
> @tomkennedy513 / @matthewmcnew this will have quite a bit of an adverse effect on the resource utilization of the final build pod. Currently the resource utilization is calculated by max(max(init containers), sum(containers)). This change will effectively lead to each build pod consuming 5x the resources due to each container's requested cpu and memory.

I will add this as a risk, but I think we could mitigate this by setting a lower request/limit for each container other than build. Additionally, The resources required during a containers wait period should be pretty minimal, but it is worth calling out that it will definitely be more than the current init container flow

rfcs/0000-execute-builds-in-standard-containers.md
```diff
@@ -0,0 +1,132 @@
...
+2. We can very easily define the order of the steps occurring as well as waiting for the previous step to complete
+   before continuing on. Additionally, failures in one init container cause the entire build to fail without additional
+   logic on our part.
+
+As kpack has grown, we have seen a need to support [service meshes][istio-service-mesh] such as [Istio][istio], which do
+not work when you have init containers that need to reach out to the network. This is
+a [well documented][istio-init-container-compatability] technical limitation related to
+Istio's [automatic sidecar injection][istio-sidecar-injection]. The limitation is due to Istio adding a sidecar
+container to the kpack build pod which runs [envoy proxy][envoy-proxy] and sets up traffic from the build containers to
+be routed through the proxy. Due to the fact that init containers run before any normal container can start, kpack
```
stormqueen1990, line 19: Q: it makes sense to me that the limitation is linked to the fact that sidecars are regular containers, but wouldn't this limitation be on Kubernetes (i.e. the way that `initContainers` relate to regular containers)?

rfcs/0000-execute-builds-in-standard-containers.md
```diff
@@ -0,0 +1,132 @@
...
+   method for stopping the proxy once a build exits
+    1. To do this we can update the running pod's `spec.activeDeadlineSeconds` to 1 which will cause the pod to go into
+       a failed state and kill all containers after 1 second. This will happen in the case that during a build
+       reconcile, kpack detects that a pod is running, but completion container is not. (The completion container will
+       exit in the event that another build container fails)
+    2. This will kill the sidecar, but it will also leave our pod in a failed state with the message `DeadlineExceeded`,
+       leading the reconciler to believe
+       the build has failed. To circumvent this we will add a check that if a pod has failed with the
+       reason `DeadlineExceeded` AND the completion container has successfully completed, then the build is actually
+       successful.
```
stormqueen1990, line 80: If I'm not mistaken, Tekton uses a mechanism that involves changing the image in a sidecar container to a no-op image: https://github.com/tektoncd/pipeline/blob/119f7bdf2dd73bbd2a6b74506c75e000c7535f49/pkg/pod/entrypoint.go#L238
Would this be an avenue worth exploring?

rfcs/0000-execute-builds-in-standard-containers.md
```diff
@@ -0,0 +1,132 @@
...
+
+[kpack-spike-build-waiter]: https://github.com/pivotal/kpack/blob/execute-builds-in-regular-containers/cmd/build-waiter/main.go
+
+[step-modifier]: https://github.com/pivotal/kpack/blob/44257a70c2c9e9ac5703eadc871c7c9be22cfadc/pkg/apis/build/v1alpha2/build_pod.go#L181
+
+## Complexity
+
+* Our current build logging relies heavily on the fact that our builds exist in init containers so that logic will
+  need to be reworked to support this change. This logging change will also need to be aware of any feature gates
+  to enable this feature
```
stormqueen1990, line 90: Could using a logging sidecar be a potential solution to this problem?

rfcs/0000-execute-builds-in-standard-containers.md
tomkennedy513: Ya it is a kubernetes limitation, but the main reason that Istio is causing issues for us is because their init container runs before ours and sets the iptables rules for all containers to use the istio-proxy, but when our init container based builds start, the proxy is not started since it is a regular container. This results in our init containers not being able to access the outside internet

stormqueen1990: Ooohh got it!

rfcs/0000-execute-builds-in-standard-containers.md
tomkennedy513: Yes, this is proposed in the first alternative towards the bottom. The main reason we liked activeDeadlineSeconds is that it is a pod level spec change, so you don't have to iterate over ever sidecar. Additionally the no-op image still results in a failed pod because you can not modify the pre-hook lifecycle that istio is running so the no-op exits with command not found. Tekton gets around this during the reconcile by ignoring that failure similar to how I am proposing. The main difference is that it is very easy to catch the activeDeadlineSeconds pod failure int he reconcile of a build

rfcs/0000-execute-builds-in-standard-containers.md
tomkennedy513: we could, but i think tailing the logs from a normal container build will be even easier than init containers

rfcs/0000-execute-builds-in-standard-containers.md
tomkennedy513: Here is some other context from the tekton slack https://tektoncd.slack.com/archives/CLCCEBUMU/p1660923060292999

rfcs/0000-execute-builds-in-standard-containers.md
```diff
@@ -0,0 +1,132 @@
...
+
+https://github.com/tektoncd/pipeline/tree/main/cmd/entrypoint#waiting-for-sidecars
+
+### 2. kpack spike
+
+https://github.com/pivotal/kpack/pull/1019
+
+## Alternatives
+
+* We can create tekton [`TaskRuns`][tekton-task-run] to execute each step of the build since tekton has come up with
```
tylerphelan, line 105: I think due to the maintenance benefits of this alternative, I would vote for it

rfcs/0000-execute-builds-in-standard-containers.md
```diff
@@ -0,0 +1,132 @@
+## Problem
+
+Currently, kpack executes each stage of the [CNB Lifecycle][cnb-lifecycle] inside of
+individual [init containers][init-containers]. This has two major benefits:
+
+1. Different stages of the lifecycle are isolated from one another allowing the credentials required to push to the
```
matthewmcnew, line 6: Perhaps, just as important, this prevents application code from from accessing the creds as well.

rfcs/0000-execute-builds-in-standard-containers.md
```diff
@@ -0,0 +1,132 @@
...
+
+[istio-service-mesh]: https://istio.io/latest/about/service-mesh/#what-is-a-service-mesh
+
+[istio-init-container-compatability]: https://istio.io/latest/docs/setup/additional-setup/cni/#compatibility-with-application-init-containers
+
+## Outcome
+
+Kpack builds can execute successfully when an Istio sidecar when injected into a build pod.
+
+## Actions to take
```
matthewmcnew, line 47: Step 3, calls out the feature flag but, is in conjunction with an implementation detail. Can we call that the user experience of the feature flag explicitly?

rfcs/0000-execute-builds-in-standard-containers.md
tomkennedy513: I think the maintenance burden will be similar in either case because the build waiter image will likely be fairly static so the bulk of the maintenance will be in the build pod generation and reconciliation logic which will exist for both the main proposed solution and the alternative

scothis:
> @tomkennedy513 / @matthewmcnew this will have quite a bit of an adverse effect on the resource utilization of the final build pod. Currently the resource utilization is calculated by max(max(init containers), sum(containers)). This change will effectively lead to each build pod consuming 5x the resources due to each container's requested cpu and memory.

I'm not sure if this is a good idea, but can [ephemeral containers](https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/) be used?

tomkennedy513:
> > @tomkennedy513 / @matthewmcnew this will have quite a bit of an adverse effect on the resource utilization of the final build pod. Currently the resource utilization is calculated by max(max(init containers), sum(containers)). This change will effectively lead to each build pod consuming 5x the resources due to each container's requested cpu and memory.
> 
> I'm not sure if this is a good idea, but can [ephemeral containers](https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/) be used?

One issue with ephemeral containers is that you cannot specify resources on them at all. Also, I think this would cause an issue with our logging, particularly the use case where you want to look at a previous builds logs since the container wont exist anymore.

rfcs/0000-execute-builds-in-standard-containers.md
```diff
@@ -0,0 +1,132 @@
...
+       will exit with the same error code, forcing the pod to enter a failed state. There is also a configurable
+       timeout period that will specify how long the container is going to wait for th eprevious steps to complete
+       before exiting with an error. Note: For `prepare` we can provide an empty `wait step` so that it begins
+       executing immediately.
+
+   Example of possible `build-waiter` implementation is in the [spike][kpack-spike-build-waiter].
+
+2. Convert the existing init containers in `build_pod.go` to be regular containers and add an init container to copy
+   the `build-waiter` to a shared mount
+3. Using a similar [step modifier][step-modifier] approach , we can put this new behavior behind a feature gate and
```
matthewmcnew, line 66: This is getting extremely implementation specific which I believe is not important for the underlying RFC. Would it be possible to call out just the expected outcome?

tylerphelan (approved):

rfcs/0000-support-injected-sidecars.md
```diff
@@ -0,0 +1,113 @@
...
+
+[istio-sidecar-injection]: https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#injection
+
+[istio-service-mesh]: https://istio.io/latest/about/service-mesh/#what-is-a-service-mesh
+
+[istio-init-container-compatability]: https://istio.io/latest/docs/setup/additional-setup/cni/#compatibility-with-application-init-containers
+
+## Outcome
+
+Kpack builds can execute successfully when an Istio sidecar when injected into a build pod.
```
matthewmcnew, line 45: Can you add a bit of context about the feature flags here?

Like when kpack controller is started with an feature flag env variable builds can execute successfully when an Istio sidecar when injected into a build pod.

rfcs/0000-support-injected-sidecars.md
```diff
@@ -0,0 +1,117 @@
...
+   running concurrently. There will also need to be a mechanism for propagating errors in one build step so that a build
+   can exit on error.
+
+   Example of possible implementation is in the [spike][kpack-spike-build-waiter].
+2. Convert the existing init containers in to be regular containers and add modify the entrypoint to be the new binary.
+3. We will put this new behavior behind a feature gate (Environment variable added to kpack controller deployment
+   called `SUPPORT_INJECTED_SIDECARS`) and only execute in regular containers if enabled by the user. In
+   the event that the feature flag is not set or set to false, kpack will continue to execute builds in init containers.
+4. Due to the fact that sidecars such as istio-proxy, do not exit when all other containers exit, we will need to add a
+   method for stopping the proxy once a build exits
```
chenbh, line 61: Do we also want a mechanism to determine if side cars are ready? It it possible for a lifecycle step to execute before the envoy-proxy fully spins up?

tomkennedy513: I actually did that in the proposed implementation, but I wasn't sure if it was too much of an implementation detail to mention in the rfc. If you think it's worth calling out then I can add it

chenbh: Personally, I feel like if there's a call out for the side car stopping mechanism, there should be a call out for the lifecycle starting mechanism

chenbh (approved):

matthewmcnew (approved):

rfcs/0000-support-injected-sidecars.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Alternatives
+
+* We can create tekton [`TaskRuns`][tekton-task-run] to execute each step of the build since tekton has come up with
+  their own way to solve this problem which
+  involves [replacing the sidecar container images with a "noop" image][tekton-stop-sidecar].
+    * Note: Tekton currently will only stop sidecars (including the istio proxy) if at least one sidecar is defined in
+      the `TaskSpec`.
+* We can build the Tekton entrypoint code as an image to use in kpack.
+* We could execute all build steps in one standard container, similar to [Pack's][pack-build-docs] `pack build` command
```
yaelharel, line 103: Don't you think it's worth doing that?

rfcs/0000-support-injected-sidecars.md
```diff
@@ -0,0 +1,124 @@
...
+3. We will put this new behavior behind a feature gate (Environment variable added to kpack controller deployment
+   called `SUPPORT_INJECTED_SIDECARS`) and only execute in regular containers if enabled by the user. In
+   the event that the feature flag is not set or set to false, kpack will continue to execute builds in init containers.
+4. Due to the fact that sidecars such as istio-proxy, do not exit when all other containers exit, we will need to add a
+   method for stopping the proxy once a build exits
+    1. To do this we can update the running pod's `spec.activeDeadlineSeconds` to 1 which will cause the pod to go into
+       a failed state and kill all containers after 1 second.
+    2. This will kill the sidecar, but it will also leave our pod in a failed state with the message `DeadlineExceeded`,
+       leading the reconciler to believe the build has failed. To circumvent this we will need allow pods to
+       fail with the reason `DeadlineExceeded` as long the completion container has successfully completed.
```
yaelharel, line 66: Is there a different way to solve this?
I'm not sure it's a good idea to allow pods to fail with the reason `DeadlineExceeded`, since we may find in the future, a different flow that shouldn't be ignored with the same error.

rfcs/0000-support-injected-sidecars.md
```diff
@@ -0,0 +1,124 @@
...
+
+[pack-build-docs]: https://buildpacks.io/docs/tools/pack/cli/pack_build/
+
+[tekton-task-run]: https://tekton.dev/docs/pipelines/taskruns/#overview
+
+[tekton-stop-sidecar]: https://github.com/tektoncd/pipeline/blob/210dbe8965ab5fa9c0ca53e164b86d441899b763/pkg/reconciler/taskrun/taskrun.go#L240-L283
+
+## Risks
+
+* Because all the containers will be running at once, the pod will request more resources than it does today as due to
```
yaelharel, line 117: Can you please explain what do you mean by "all the containers will be running at once"?
Do you mean the containers that run the lifecycle stages? Because I'm not sure it's possible.

rfcs/0000-support-injected-sidecars.md
tomkennedy513: The alternative is to replace the image the sidecar is using with a no-op image that just exits (this is what tekton does). The problem with that is that the istio sidecar adds a lifecycle prehook (which we can't remove) that attempts to execute a particular command that isn't present in the no-op image so it still ends up exiting with an error. Another alternative would be to sent a POST request to the istio sidecar to tell it to shutdown, but I decided not to go down this path because it is very istio specific whereas the current proposal would work for any sidecar. I'm not super concerned about the DeadlineExceeded error because we are checking that the build still completed successfully before marking the build as successful. If the build failed and causes a deadline exceeded it will still fail the build.

rfcs/0000-support-injected-sidecars.md
tomkennedy513: The containers will all be started (as opposed to init containers which don't start until it is their turn), but the build waiter logic will prevent the actual lifecycle command from executing until the previous step completes. The container will just sit and wait.

rfcs/0000-support-injected-sidecars.md
tomkennedy513: Which one are you referring to, the github comment diff is showing multiple lines for me so I can't tell which one

rfcs/0000-support-injected-sidecars.md
yaelharel: ohh, and this should solve the network issue that exists with the init containers, right?

rfcs/0000-support-injected-sidecars.md
yaelharel: oops, I thought I marked different lines. Sorry about that.
I refer to:

> We could execute all build steps in one standard container, similar to [Pack's][pack-build-docs] `pack build` command

yaelharel (approved):

samj1912 (approved):

jjbustamante (approved):

PR #1000, opened 2022-08-02
Retry logic for builders on network failures

### LINKED ISSUE #550: Builders do not recover from temporary issues.
1. I created a `CustomBuilder` object and during the creation time the registry specified was down.
1. As expected the `Ready` condition is set to `false`.
1. The registry became available shortly after.

Expected behavior: Eventually the `Ready` condition will transition to `true` (if there are no further errors of course)
Observed behavior: `Ready` stays `false` forever.

I think this is caused by unconditionally wrapping any error with `NewPermanentError`:
https://github.com/pivotal/kpack/blob/0836eeb5791ce1961dbd68bcb6f2802d1034035c/pkg/reconciler/builder/builder.go#L88-L98

According to [knative documentation](https://pkg.go.dev/knative.dev/pkg/controller#NewPermanentError):
>  Users can wrap an error as permanentError with this in reconcile when they do not expect the key to get re-queued.

IMHO this should not be used for external dependencies to be more resilient against short unavailabilities.

jjbustamante:
### Solution

A new *NetworkErrorReconciler* struct was created to wrapped the existing Reconciler for:
 - BuilderReconciler
 - ClusterBuilderReconciler
 - ClusterStackReconciler
 - ClusterStoreReconciler

The client library that is accessing the Registries (currently using GGCR library) wraps any network error with a new error type.
The *NetworkErrorReconciler* verifies if the error thrown by its wrapping Reconcilers is a network error, in such as case, it will re-queue the task otherwise it will behave as it is right now

Fixes #550 

Signed-off-by: Juan Bustamante <jbustamante@vmware.com>

### DIFF
--- a/pkg/reconciler/builder/builder.go
+++ b/pkg/reconciler/builder/builder.go
@@ -55,7 +55,9 @@ func NewController(opt reconciler.Options,
 		zap.String(logkey.Kind, buildapi.BuilderCRName),
 	)
 
-	impl := controller.NewImpl(c, logger, ReconcilerName)
+	impl := controller.NewImpl(&reconciler.NetworkErrorReconciler{
+		Reconciler: c,
+	}, logger, ReconcilerName)
 	builderInformer.Informer().AddEventHandler(reconciler.Handler(impl.Enqueue))
 
 	c.Tracker = tracker.New(impl.EnqueueKey, opt.TrackerResyncPeriod())
@@ -100,8 +102,7 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 		if err != nil {
 			return err
 		}
-
-		return controller.NewPermanentError(creationError)
+		return creationError
 	}
 
 	builder.Status.BuilderRecord(builderRecord)
--- a/pkg/reconciler/network_error_reconciler.go
+++ b/pkg/reconciler/network_error_reconciler.go
@@ -1,0 +1,31 @@
+package reconciler
+
+import (
+	"context"
+	"github.com/pkg/errors"
+	"knative.dev/pkg/controller"
+)
+
+type NetworkErrorReconciler struct {
+	Reconciler controller.Reconciler
+}
+
+func (r *NetworkErrorReconciler) Reconcile(ctx context.Context, key string) error {
+	if err := r.Reconciler.Reconcile(ctx, key); err != nil {
+		var networkError *NetworkError
+		if errors.As(err, &networkError) {
+			// Re-queue the key if it's a network error.
+			return err
+		}
+		return controller.NewPermanentError(err)
+	}
+	return nil
+}
+
+type NetworkError struct {
+	Err error
+}
+
+func (e *NetworkError) Error() string {
+	return e.Err.Error()
+}
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -2,12 +2,15 @@ package registry
 
 import (
 	"fmt"
+	"net/http"
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
+	"github.com/pivotal/kpack/pkg/reconciler"
 	"github.com/pkg/errors"
 )
 
@@ -22,7 +25,7 @@ func (t *Client) Fetch(keychain authn.Keychain, repoName string) (v1.Image, stri
 
 	image, err := remote.Image(reference, remote.WithAuthFromKeychain(keychain))
 	if err != nil {
-		return nil, "", err
+		return nil, "", handleError(err)
 	}
 
 	identifier, err := getIdentifier(image, reference)
@@ -51,7 +54,7 @@ func (t *Client) Save(keychain authn.Keychain, tag string, image v1.Image) (stri
 	}
 	err = remote.Write(ref, image, remote.WithAuthFromKeychain(keychain))
 	if err != nil {
-		return "", err
+		return "", handleError(err)
 	}
 
 	return identifier, remote.Tag(ref.Context().Tag(timestampTag()), image, remote.WithAuthFromKeychain(keychain))
@@ -82,4 +85,16 @@ func previousDigest(keychain authn.Keychain, ref name.Reference) string {
 	}
 
 	return hash.String()
+}
+
+func handleError(err error) error {
+	if transportErr, ok := err.(*transport.Error); ok {
+		if transportErr.StatusCode != http.StatusUnauthorized &&
+			transportErr.StatusCode != http.StatusForbidden {
+			return &reconciler.NetworkError{
+				Err: err,
+			}
+		}
+	}
+	return err
 }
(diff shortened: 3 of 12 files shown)


jjbustamante:
I tested this PR locally very quickly just to validate if I can reproduce/fix the problem it worked but the way I try to reproduce the network failure was very rudimentary (I disconnect my wireless card 😄). I will try to use some tool to introduce the failure but I just wanted to share the idea of the implementation and get early feedback

pkg/reconciler/handlers.go
```diff
@@ -12,3 +14,33 @@ func Handler(h func(interface{})) cache.ResourceEventHandler {
...
+}
+
+type retryErrorHandler struct {
+	retryDuration time.Duration
+}
+
+func (r *retryErrorHandler) OnReconcileError(err error) error {
+	var networkError *NetworkError
+	if errors.As(err, &networkError) {
+		return controller.NewRequeueAfter(r.retryDuration)
```
matthewmcnew, line 35: My understanding is that that underlying reconciler will automatically requeue errors with an exponential backoff. Do we need to explicitly provide the NewRequeueAfter with a duration?

matthewmcnew: Won't an explicit duration also prevent the backoff?

pkg/reconciler/handlers.go
```diff
@@ -12,3 +14,33 @@ func Handler(h func(interface{})) cache.ResourceEventHandler {
...
+	}
+}
+
+type retryErrorHandler struct {
+	retryDuration time.Duration
+}
+
+func (r *retryErrorHandler) OnReconcileError(err error) error {
+	var networkError *NetworkError
+	if errors.As(err, &networkError) {
```
matthewmcnew, line 34: Nice!

pkg/reconciler/handlers.go
```diff
@@ -12,3 +14,33 @@ func Handler(h func(interface{})) cache.ResourceEventHandler {
 		DeleteFunc: h,
 	}
 }
+
+type ErrorHandler interface {
```
matthewmcnew, line 18: Could we invert the dependency here? A retrying NetworkErrorReconciler that takes wraps a reconciler interface and is constructed with a reconciler?

pkg/reconciler/handlers.go
jjbustamante: I think you are right, I was checking the following [code](https://github.com/knative/pkg/blob/main/controller/controller.go#L560-L576) in the knative package. I misunderstood creating a *requeueKeyError* was mandatory. 
If just re-throw the transient error it actually be re-queued

pkg/reconciler/handlers.go
jjbustamante: I like it, let me give it a shot 😄

pkg/reconciler/network_error_reconciler.go
```diff
@@ -0,0 +1,31 @@
...
+
+type NetworkErrorReconciler struct {
+	Reconciler controller.Reconciler
+}
+
+func (r *NetworkErrorReconciler) Reconcile(ctx context.Context, key string) error {
+	if err := r.Reconciler.Reconcile(ctx, key); err != nil {
+		var networkError *NetworkError
+		if errors.As(err, &networkError) {
+			// Re-queue the key if it's a network error.
```
matthewmcnew, line 17: Nice!

matthewmcnew (approved):

pkg/registry/client.go
```diff
@@ -22,7 +23,9 @@ func (t *Client) Fetch(keychain authn.Keychain, repoName string) (v1.Image, stri
 
 	image, err := remote.Image(reference, remote.WithAuthFromKeychain(keychain))
 	if err != nil {
-		return nil, "", err
+		return nil, "", &reconciler.NetworkError{
```
tylerphelan, line 26: should this be returned for writes too?

pkg/reconciler/network_error_reconciler.go
```diff
@@ -0,0 +1,31 @@
...
+			return err
```
tylerphelan, line 18: this automatically results in exponential backoff, right?

pkg/registry/client.go
matthewmcnew, line 26: Should we attempt to not return a network error on unauthorized?

jjbustamante: Well, actually the `RegistryClient.Save()` method is been [called](https://github.com/pivotal/kpack/blob/44257a70c2c9e9ac5703eadc871c7c9be22cfadc/pkg/cnb/create_builder.go#L74) in the same logic that is been used by the reconciler. So, my answer would be yes!

jjbustamante: I haven't check the code, but, is not the access verification checked before? is it related to this [PR](https://github.com/pivotal/kpack/pull/1008) ?

jjbustamante: @tylerphelan I just added the same error handling method to the Save operation

jjbustamante: @matthewmcnew I added some validation to the error throw on `Fetch` and `Save` to verify in the error is not an authorized. Please take a look and let me know your thoughts

pkg/registry/client.go
```diff
@@ -83,3 +86,15 @@ func previousDigest(keychain authn.Keychain, ref name.Reference) string {
 
 	return hash.String()
 }
+
+func handleError(err error) error {
+	if transportErr, ok := err.(*transport.Error); ok {
+		if transportErr.StatusCode != http.StatusUnauthorized &&
+			transportErr.StatusCode != http.StatusForbidden {
```
matthewmcnew, line 93: Should we include 401s as well?

pkg/reconciler/builder/builder.go
```diff
@@ -23,6 +12,14 @@ import (
 	"github.com/pivotal/kpack/pkg/reconciler"
 	"github.com/pivotal/kpack/pkg/registry"
 	"github.com/pivotal/kpack/pkg/tracker"
```
tylerphelan, line 14: can we put the kpack packages at the bottom of the import clause similar to before?

tylerphelan (approved):
just clean up imports 👍

pkg/reconciler/builder/builder.go
jjbustamante: Done

jjbustamante:
> just clean up imports 👍

Done!

tomkennedy513 (approved):
Hey @jjbustamante are you able to squash some of those commits (or rewrite them) so that we don't have a wip commit on main

jjbustamante:
> Hey @jjbustamante are you able to squash some of those commits (or rewrite them) so that we don't have a wip commit on main

Done @tomkennedy513

matthewmcnew:
Marking this for the 0.7.0 milestone.

danail-branekov:
@matthewmcnew Is it possible to have this PR released with the next release, looking at the milestones, there seems to be 0.6.2 planned.

Reconcilers returning `PermanentError` effectively stopping the reconcile loop is really annoying. We (Korifi) often get a clusterbuilder that never gets ready as the referenced store is not found, while in fact it has been created just after the builder. We have also minor workarounds in the code to ensure that all kpack build prerequisites are in place before creating the kpack build as it would otherwise never run (because at the time the build is created the referenced secret is not there). 

Stopping the reconcile loop for not a really-really very-very-very good reason is definitely not the K8S way of doing things. We are all looking forward having this fix.

cc @cloudfoundry/cf-k8s

matthewmcnew:
We are being intentionally conservative with our patch releases. However, the 0.7.0 release should be within the next week or  two.

ncarlson (approved):

PR #1007, opened 2022-08-11
Bump k8s.io libs to 0.23.9

tylerphelan:
- Update knative/pkg to match


tomkennedy513 (approved):

cmd/controller/main.go
```diff
@@ -101,6 +101,7 @@ func main() {
 	}
 
 	options := reconciler.Options{
+		Context:                 ctx,
```
matthewmcnew, line 104: Storing contexts in structs is an antipattern to avoid if possible: https://go.dev/blog/context-and-structs. 

Could instead the context be passed into each controller's constructor?

go.mod
```diff
@@ -291,25 +291,18 @@ require (
...
 
 replace (
-	github.com/containerd/containerd => github.com/containerd/containerd v1.5.10
 	github.com/prometheus/common => github.com/prometheus/common v0.26.0
-	k8s.io/api => k8s.io/api v0.20.11
-	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.20.11
-	k8s.io/apimachinery => k8s.io/apimachinery v0.20.11
-	k8s.io/client-go => k8s.io/client-go v0.20.11
-	k8s.io/code-generator => k8s.io/code-generator v0.20.11
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20210305001622-591a79e4bda7
```
matthewmcnew, line 307: Should this be bumped as well to to at least the 1.22 branch? https://github.com/kubernetes/kube-openapi/tree/release-1.22

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/pivotal/kpack
 
-go 1.17
+go 1.18
```
matthewmcnew, line 3: Generics here we come.

cmd/controller/main.go
```diff
@@ -230,9 +231,9 @@ func main() {
 	}
 }
 
-func run(ctrl *controller.Impl, threadiness int) doneFunc {
+func run(ctx context.Context, ctrl *controller.Impl, threadiness int) doneFunc {
```
tylerphelan, line 234: are these `run` changes correct? @tomkennedy513 @matthewmcnew

cmd/controller/main.go
tylerphelan: yea that's def better

go.mod
tylerphelan: release-1.22 downgrades the other k8s.io deps. updated it to match https://github.com/tektoncd/pipeline

matthewmcnew:
Could you bump the golang version in the github action dockerfile?

cmd/controller/main.go
matthewmcnew: I think perhaps this could simplify it? https://github.com/pivotal/kpack/pull/1011

tomkennedy513 (approved):

ncarlson (approved):

matthewmcnew (approved):

PR #1008, opened 2022-08-11
Use the ggcr CheckPushPermission function in VerifyWriteAccess

### LINKED ISSUE #943: Prepare step will fail for https registries that use implicitly "local" urls
The `prepare` step will fail if using an https registry that has an implicitly local/insecure URL determined by [ggcr](https://github.com/google/go-containerregistry/blob/a0c4bd256482b8522065d5f6cf966281ef270680/pkg/name/registry.go#L65)

yaelharel:
Fixes #943

### DIFF
--- a/pkg/dockercreds/access_checker.go
+++ b/pkg/dockercreds/access_checker.go
@@ -18,7 +18,11 @@ func VerifyWriteAccess(keychain authn.Keychain, tag string) error {
 		return errors.Wrapf(err, "Error parsing reference %q", tag)
 	}
 
-	return remote.CheckPushPermission(ref, keychain, http.DefaultTransport)
+	if err = remote.CheckPushPermission(ref, keychain, http.DefaultTransport); err != nil {
+		return diagnoseIfTransportError(err)
+	}
+
+	return nil
 }
 
 func VerifyReadAccess(keychain authn.Keychain, tag string) error {
--- a/pkg/dockercreds/access_checker.go
+++ b/pkg/dockercreds/access_checker.go
@@ -3,7 +3,6 @@ package dockercreds
 import (
 	"fmt"
 	"net/http"
-	"net/url"
 	"strings"
 
 	"github.com/google/go-containerregistry/pkg/authn"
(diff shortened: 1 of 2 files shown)


tomkennedy513 (approved):
So much cleaner!!

pkg/dockercreds/access_checker.go
```diff
@@ -14,42 +13,12 @@ import (
...
-
-	if err = transport.CheckError(resp, http.StatusCreated, http.StatusAccepted); err != nil {
-		return err
-	}
-
-	return nil
+	return remote.CheckPushPermission(ref, keychain, http.DefaultTransport)
 }
 
 func VerifyReadAccess(keychain authn.Keychain, tag string) error {
```
tylerphelan, line 24: I'm not sure if we need the `func diagnoseIfTransportError(err error) error` still. This appears to be handling an edge case for artifactory. I think we should remove this or add the artifactory test back. What's the context for this? @matthewmcnew

yaelharel: I saw that we're using it in [VerifyReadAccess](https://github.com/pivotal/kpack/blob/44257a70c2c9e9ac5703eadc871c7c9be22cfadc/pkg/dockercreds/access_checker.go#L62).

yaelharel: By the way, I don't know if it matters or not but [imgutil is using the ggcr Head function](https://github.com/buildpacks/imgutil/blob/56137f75e24d5b567db1dbb47ab05e958e257ec2/remote/remote.go#L476) in order to check read access while [kpack is using the ggcr Get function](https://github.com/pivotal/kpack/blob/44257a70c2c9e9ac5703eadc871c7c9be22cfadc/pkg/dockercreds/access_checker.go#L61).

tylerphelan: `imgutil` is probably doing it correctly

matthewmcnew (approved):

pkg/dockercreds/access_checker.go
tylerphelan: using get is probably fine

tylerphelan: let's just add `diagnoseIfTransportError` to the write check. using get should still be fine.

yaelharel: Please see a67ce1afb9eda1ec9c5517379b6600b70381372b (I'll merge all of the commits before merging to `main`).

tylerphelan (approved):

PR #1012, opened 2022-08-19
Add workaround to fix regression

tylerphelan:
- Bring back support for AZURE_CONTAINER_REGISTRY_CONFIG auth with acr

POC for supporting `AZURE_CONTAINER_REGISTRY_CONFIG`

### DIFF
--- a/pkg/dockercreds/k8sdockercreds/azurecredentialhelperfix/keychain.go
+++ b/pkg/dockercreds/k8sdockercreds/azurecredentialhelperfix/keychain.go
@@ -1,0 +1,68 @@
+package azurecredentialhelperfix
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	credentialprovider "github.com/vdemeester/k8s-pkg-credentialprovider"
+)
+
+var (
+	once    sync.Once
+	keyring credentialprovider.DockerKeyring
+)
+
+func AzureFileKeychain() authn.Keychain {
+	once.Do(func() {
+		keyring = credentialprovider.NewDockerKeyring()
+	})
+
+	return &keychain{keyring: keyring}
+}
+
+// Copied from https://github.com/google/go-containerregistry/blob/d9bfbcb99e526b2a9417160e209b816e1b1fb6bd/pkg/authn/k8schain/k8schain.go#L141
+type lazyProvider struct {
+	kc    *keychain
+	image string
+}
+
+// Authorization implements Authenticator.
+func (lp lazyProvider) Authorization() (*authn.AuthConfig, error) {
+	creds, found := lp.kc.keyring.Lookup(lp.image)
+	if !found || len(creds) < 1 {
+		return nil, fmt.Errorf("keychain returned no credentials for %q", lp.image)
+	}
+	authConfig := creds[0]
+	return &authn.AuthConfig{
+		Username:      authConfig.Username,
+		Password:      authConfig.Password,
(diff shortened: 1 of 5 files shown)


pkg/dockercreds/k8sdockercreds/keychain.go
```diff
@@ -0,0 +1,73 @@
...
+
+// Options holds configuration data for guiding credential resolution.
+type Options = kauth.Options
+
+// New returns a new authn.Keychain suitable for resolving image references as
+// scoped by the provided Options.  It speaks to Kubernetes through the provided
+// client interface.
+func New(ctx context.Context, client kubernetes.Interface, opt Options) (authn.Keychain, error) {
+	if os.Getenv("AZURE_CONTAINER_REGISTRY_CONFIG") != "" {
+		azureKeychain = azurecredentialhelperfix.AzureFileKeychain()
```
matthewmcnew, line 47: Would it work to place both keychains in the chain?

pkg/dockercreds/k8sdockercreds/azurecredentialhelperfix/keychain.go
```diff
@@ -0,0 +1,68 @@
...
+	"fmt"
+	"sync"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	credentialprovider "github.com/vdemeester/k8s-pkg-credentialprovider"
+)
+
+func AzureFileKeychain() authn.Keychain {
+	var (
```
matthewmcnew, line 13: I think you would want these to be package level vars.

pkg/dockercreds/k8sdockercreds/keychain.go
tylerphelan: added it as a multi keychain

matthewmcnew: I guess, I am wondering if we should always include both the old and the new azure keychain at the same time.

tylerphelan: I think it's a one-off scenario

matthewmcnew: But, is it it necessary? Wouldn't it prevent using the both at the same time perhaps complicating the migration?

tylerphelan: I'm not sure how both at the same time works, won't it always use only 1?

matthewmcnew: My understanding is that the [multikey chain logic](https://github.com/google/go-containerregistry/blob/main/pkg/authn/multikeychain.go#L31-L40) will iterate through all the keychains until it finds one that does not return [anonymous auth](https://github.com/google/go-containerregistry/blob/main/pkg/authn/keychain.go#L168).

tylerphelan: I can't picture how the migration path would be impacted

tylerphelan: maybe if the order of these 2 is switched it would mitigate the risk: 
```
azureKeychain = authn.NewMultiKeychain(azurecredentialhelperfix.AzureFileKeychain(), azureKeychain)
```

matthewmcnew: I missed the addition of the authn.MultiKeyChain. I think that solves the migration problem. 

But, I still think the conditional on the existence of the env var is unnecessary. Why not just put both in there always?

tylerphelan:
closing in favor of https://github.com/pivotal/kpack/pull/1014

PR #1019, opened 2022-08-24
Add support for injected sidecars in build pods

### LINKED ISSUE #986: Support istio sidecar injection in build namespaces
Istio/envoy proxy does not work well with kpack, especially in build namespaces, because kpack uses init containers for the CNB process.

Can we improve this to support istio networking on kpack clusters?

tomkennedy513:
implementation of proposed #999 

resolves #986

### DIFF
--- a/cmd/build-waiter/main.go
+++ b/cmd/build-waiter/main.go
@@ -1,0 +1,147 @@
+package main
+
+import (
+	"flag"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+var (
+	mode     = flag.String("mode", "wait", "One of: wait or copy")
+	to       = flag.String("to", "", "Where to copy this binary")
+	waitFile = flag.String("wait-file", "", "file to wait on")
+	doneFile = flag.String("done-file", "", "file to write on completion")
+	execute  = flag.String("execute", "", "What to run after waiting")
+	timeout  = flag.String("timeout", "60m", "How long to wait")
+	errFile  = flag.String("error-file", "", "Location of shared error file")
+)
+
+func main() {
+	flag.Parse()
+
+	switch *mode {
+	case "wait":
+		if *doneFile == "" {
+			log.Fatal("need -done-file with -mode=wait")
+		}
+		if *execute == "" {
+			log.Fatal("need -execute with -mode=wait")
+		}
+
+		exists, err := exists(*waitFile)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		if exists {
+			log.Fatal("file specified for writing already exists")
+		}
+
+		if *waitFile != "" {
+			log.Printf("waiting on %s\n", *waitFile)
+			err = waitForStep(*waitFile)
+			if err != nil {
+				log.Fatal(err)
+			}
+		}
+
+		split := strings.Split(*execute, " ")
+
+		cmd := exec.Command(split[0], split[1:]...)
+
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+
+		if err := cmd.Run(); err != nil {
+			if *errFile != "" {
+				os.Create(*errFile)
+			}
+			log.Fatalf("error running command %s", err.Error())
+		}
+
+		err = ioutil.WriteFile(*doneFile, []byte("done"), 0666)
+		if err != nil {
+			log.Fatalf("error writing file %s", err.Error())
+		}
+	case "copy":
+		if *to == "" {
+			log.Fatal("-to must be specified with -mode=copy")
+		}
+		binaryPath, err := os.Executable()
+		if err != nil {
+			log.Fatal(err)
+		}
+		from, err := os.Open(binaryPath)
+		if err != nil {
+			log.Fatal(err)
+		}
+		defer from.Close()
+
+		needsDelete, err := exists(*to)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		if needsDelete {
+			err := os.RemoveAll(*to)
+			if err != nil {
+				log.Fatal(err)
+			}
+		}
+
+		to, err := os.OpenFile(*to, os.O_RDWR|os.O_CREATE, 0777)
+		if err != nil {
+			log.Fatal(err)
+		}
+		defer to.Close()
+
+		_, err = io.Copy(to, from)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+}
+
+func exists(fileName string) (bool, error) {
+	_, err := os.Stat(fileName)
+	if os.IsNotExist(err) {
+		return false, nil
+	}
+	return err != nil, err
+}
+
+func waitForStep(filepath string) error {
+	duration, err := time.ParseDuration(*timeout)
+	if err != nil {
+		return err
+	}
+	timeoutChan := time.After(duration)
+	tickerChan := time.NewTicker(time.Second)
+	defer tickerChan.Stop()
+
+	for {
+		select {
+		case <-timeoutChan:
+			return errors.New("timeout waiting for previous step to complete")
+		case <-tickerChan.C:
+			if *errFile != "" {
+				if _, err := os.Stat(*errFile); err == nil {
+					log.Fatalf("another step has errored")
+				}
+			}
+			if fi, err := os.Stat(filepath); err == nil {
+				//downward api places a file even if the annotation is empty, so we need to wait for non-empty file
+				if fi.Size() > 0 {
+					return nil
+				}
+			}
+		}
+	}
+}
--- a/cmd/build-waiter/main.go
+++ b/cmd/build-waiter/main.go
@@ -1,26 +1,25 @@
 package main
 
 import (
+	"context"
+	"errors"
 	"flag"
+	"fmt"
 	"io"
-	"io/ioutil"
 	"log"
 	"os"
 	"os/exec"
 	"strings"
 	"time"
-
-	"github.com/pkg/errors"
 )
 
 var (
-	mode     = flag.String("mode", "wait", "One of: wait or copy")
-	to       = flag.String("to", "", "Where to copy this binary")
+	mode     = flag.String("mode", "wait", "one of: wait or copy")
+	to       = flag.String("to", "", "where to copy this binary")
 	waitFile = flag.String("wait-file", "", "file to wait on")
 	doneFile = flag.String("done-file", "", "file to write on completion")
 	execute  = flag.String("execute", "", "What to run after waiting")
-	timeout  = flag.String("timeout", "60m", "How long to wait")
-	errFile  = flag.String("error-file", "", "Location of shared error file")
+	errFile  = flag.String("error-file", "", "shared error file")
 )
 
 func main() {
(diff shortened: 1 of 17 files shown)


pkg/apis/build/v1alpha2/build.go
```diff
@@ -137,8 +137,7 @@ func (b *Build) PodName() string {
 }
 
 func (b *Build) MetadataReady(pod *corev1.Pod) bool {
-	return !b.Status.GetCondition(corev1alpha1.ConditionSucceeded).IsTrue() &&
-		pod.Status.Phase == "Succeeded"
+	return pod.Status.Phase == "Succeeded" || (pod.Status.Phase == "Failed" && pod.Status.Reason == "DeadlineExceeded")
```
stormqueen1990, line 140: If you're interested in using constants for phase values, they're defined on k8s.io/api/core/v1: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L2393-L2414

pkg/apis/build/v1alpha2/build.go
```diff
@@ -137,8 +137,7 @@ func (b *Build) PodName() string {
 }
 
 func (b *Build) MetadataReady(pod *corev1.Pod) bool {
-	return !b.Status.GetCondition(corev1alpha1.ConditionSucceeded).IsTrue() &&
-		pod.Status.Phase == "Succeeded"
+	return pod.Status.Phase == corev1.PodSucceeded || (pod.Status.Phase == corev1.PodFailed && pod.Status.Reason == "DeadlineExceeded")
```
tylerphelan, line 140: is it possible for deadlineexceeded to be an actual error?

tomkennedy513: Yes, but since we validate that the completion step had finished, the build should be considered successful. If the build exits with deadline exceeded and the completion step hadnt finished then it will still fail

pkg/logs/build_logs.go
```diff
@@ -122,24 +122,20 @@ func (c *BuildLogsClient) watchReadyContainers(ctx context.Context, readyContain
...
-						readyContainers <- readyContainer{
-							podName:       pod.Name,
-							containerName: c.Name,
-							namespace:     pod.Namespace,
-						}
+				for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
+					if readyContainer, found := containers[container.Name]; found {
+						readyContainers <- readyContainer
 					}
 				}
```
chenbh, line 140: Is there an edge case where the logs are not streamed in order if a container becomes ready faster than the others?

E.g.
1. A new build pod is created, all the container statuses have `c.State.Waiting != nil`
1. A container manages to start up faster than the rest (e.g. `pod.Status.ContainerStatuses[2].State.Waiting == nil` while the rest are still non-nil) and a `watch.Modified` event gets sent out
1. `containers[pod.Spec.Containers[2].Name]` is set and sent to the `readyContainers` chan
1. the client starts streaming for container[2], but that container is blocked until container[0] and [1] completes
1. the final printed logs will be from containers 2 -> 0 -> 1 -> 3 -> 4 -> ...

pkg/reconciler/build/build.go
```diff
@@ -152,11 +170,45 @@ func (c *Reconciler) reconcile(ctx context.Context, build *buildapi.Build) error
...
+		if !container.Ready {
+			return pod, nil
+		}
+	}
+
+	patch, err := json.Marshal([]jsonpatch.JsonPatchOperation{{
+		Operation: "replace",
+		Path:      "/metadata/annotations/" + strings.Replace(buildapi.BuildReadyAnnotation, "/", "~1", 1),
+		Value:     "true",
+	}})
```
chenbh, line 193: Does the replace operation create the label if it doesn't exist? I noticed that the annotation isn't set in the pod spec, how does that affect the json patch and downwards api?

tomkennedy513: I just checked and its actually not supposed to, but for some reason it's working. We can do a check first though

cmd/build-waiter/main.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+
+		exists, err := exists(*waitFile)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		if exists {
+			log.Fatal("file specified for writing already exists")
+		}
```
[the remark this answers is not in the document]
tomkennedy513: ya i can change that. none of our code paths are writing to a downward api file, but it is weird nonetheless

cmd/build-waiter/main.go
```diff
@@ -0,0 +1,152 @@
...
+	"flag"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
```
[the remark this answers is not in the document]
tomkennedy513: ya we use that library heavily which is why I used it, but we should move off of it

cmd/build-waiter/main.go
```diff
@@ -0,0 +1,145 @@
...
+					log.Fatalf("another step has errored")
+				}
+			}
+			if fi, err := os.Stat(filepath); err == nil {
+				//downward api places a file even if the annotation is empty, so we need to wait for non-empty file
+				if fi.Size() > 0 {
+					return nil
+				}
+			}
+		}
```
chenbh, line 143: I don't think we need it, but you can also use something like https://pkg.go.dev/github.com/fsnotify/fsnotify to wait on file modified events instead of polling every x interval.

cmd/build-waiter/main.go
```diff
@@ -0,0 +1,135 @@
...
+
+func exists(fileName string) (bool, error) {
+	_, err := os.Stat(fileName)
+	if os.IsNotExist(err) {
+		return false, nil
+	}
+	return err != nil, err
+}
+
+func waitForStep(filepath string) error {
```
matthewmcnew, line 116: Can we pass in `errFile`?

config/controller.yaml
```diff
@@ -19,6 +19,14 @@ data:
 ---
 apiVersion: v1
 kind: ConfigMap
+metadata:
+  name:  build-waiter-image
+  namespace: kpack
+data:
+  image: #@ data.values.build_waiter_image
+---
```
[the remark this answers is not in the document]
matthewmcnew: The images are in single configmaps so they can all have the key `image:` that allows tools like `kbld` to identify them as images.

hack/common.sh
```diff
@@ -57,6 +60,7 @@ function compile() {
     -v controller_image=${controller_image} \
     -v webhook_image=${webhook_image} \
     -v build_init_image=${build_init_image} \
+    -v build_waiter_image=${build_waiter_image} \
```
matthewmcnew, line 63: I wonder if we should just have one image for completion,build-init,waiting. (Steal a page from our lifecycle friends)

That obviously could happen in a different PR

cmd/build-waiter/main.go
```diff
@@ -0,0 +1,178 @@
...
+			exitWithError(err.Error(), *errFile)
+		}
+	}
+}
+
+func waitForStep(waitPath string) error {
+	tickerChan := time.NewTicker(time.Second)
+	defer tickerChan.Stop()
+
+	for range tickerChan.C {
```
tylerphelan, line 54: can this possibly wait forever?

cmd/build-waiter/main.go
```diff
@@ -0,0 +1,178 @@
...
+
+	case "copy":
+		err := copy(*to)
+		if err != nil {
+			exitWithError(err.Error(), *errFile)
+		}
+	}
+}
+
+func waitForStep(waitPath string) error {
```
tylerphelan, line 50: this only returns nil

cmd/build-waiter/main.go
```diff
@@ -0,0 +1,178 @@
...
+		log.Printf("waiting on %s\n", waitFile)
+		err = waitForStep(waitFile)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+
+	split := strings.Split(execute, " ")
+
+	cmd := exec.Command(split[0], split[1:]...)
```
tylerphelan, line 87: can this panic on index out of bounds?

cmd/build-waiter/main.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	destFile, err := os.OpenFile(to, os.O_RDWR|os.O_CREATE, 0777)
+	if err != nil {
+		return err
+	}
+	defer destFile.Close()
+
+	_, err = io.Copy(destFile, srcFile)
+	if err != nil {
```
tylerphelan, line 137: can save 3 lines and return err

pkg/logs/build_logs.go
```diff
@@ -122,24 +122,20 @@ func (c *BuildLogsClient) watchReadyContainers(ctx context.Context, readyContain
 
 			switch r.Type {
 			case watch.Added, watch.Modified:
-
-				for _, c := range pod.Status.InitContainerStatuses {
-					if c.State.Waiting == nil {
-						readyContainers <- readyContainer{
+				containers := map[string]readyContainer{}
+				for _, c := range append(pod.Status.InitContainerStatuses, pod.Status.ContainerStatuses...) {
```
tylerphelan, line 126: what is the behavior of logs when they are run in regular containers? is the behavior of this package identical?

also, I'm surprised there were no test changes

pkg/reconciler/build/build.go
```diff
@@ -152,11 +170,49 @@ func (c *Reconciler) reconcile(ctx context.Context, build *buildapi.Build) error
...
+	return c.K8sClient.CoreV1().Pods(pod.Namespace).Patch(ctx, pod.Name, types.JSONPatchType, patch, metav1.PatchOptions{})
+}
+
+func (c *Reconciler) cleanupIfNeeded(ctx context.Context, pod *corev1.Pod) (*corev1.Pod, error) {
+	// check if pod is running && completion is terminated
+	if pod.Status.Phase == corev1.PodRunning && contains(stepsCompleted(pod), "completion") {
+		activeDeadlineSeconds := int64(1)
+		podCopy := pod.DeepCopy()
+		podCopy.Spec.ActiveDeadlineSeconds = &activeDeadlineSeconds
+		return c.K8sClient.CoreV1().Pods(pod.Namespace).Update(ctx, podCopy, metav1.UpdateOptions{})
```
tylerphelan, line 211: why update and not patch while the other is patch?

tylerphelan (commented):
what a pr! added some comments

cmd/build-waiter/main.go
tomkennedy513: ya I originally had a timeout but removed it. I figured that currently a build can execute forever and the solution is to provide an active deadline seconds when configuring the image which would still work in this case.

cmd/build-waiter/main.go
tomkennedy513: ya it can. I should probably add a a check even though we are the only ones using the build waiter

config/controllerrole.yaml
```diff
@@ -62,6 +62,12 @@ rules:
   - update
   - delete
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - patch
```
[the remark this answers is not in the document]
tomkennedy513: I had them together but opted to separate them just so i didnt give us a permission we don't need but either way I think it's low risk

pkg/reconciler/build/build.go
tomkennedy513: honestly no clue, I think i just did each function at different times and wasn't paying attention

pkg/logs/build_logs.go
tomkennedy513: the only change is that we make sure the containers are pushed to the channel jn the order they are defined in the spec. The behavior shouldn't change for users, but I will see if i can add a test

cmd/build-waiter/main.go
tomkennedy513: I did some tests and the when you do a `:` in the slice notation it just returns an empty array if the index doesn't exist. Since we check in the beginning of the function that `execute != ""` we shouldn't hit any panics for index out of bounds here.

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -179,9 +185,19 @@ var (
 		Name:  serviceBindingRootEnvVar,
 		Value: filepath.Join(platformMount.MountPath, "bindings"),
 	}
+	buildWaitMount = corev1.VolumeMount{
+		Name:      buildWaitVolumeName,
+		MountPath: "/buildWait",
+		ReadOnly:  false,
+	}
+	downwardMount = corev1.VolumeMount{
```
matthewmcnew, line 193: Nice.

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -735,6 +758,85 @@ func addNetworkWaitLauncherVolume() stepModifier {
...
+		fmt.Sprintf("-execute=%s %s", startCommand[0], strings.Join(containerArgs, " ")),
+	}
+	if waitFile != "" {
+		container.Args = append(container.Args, fmt.Sprintf("-wait-file=%s", waitFile))
+	}
+	return container
+
+}
+
+func (b *Build) useStandardContainers(buildWaiterImage string) podModifier {
```
matthewmcnew, line 779: Is there a reason to wrap this as function? It appears to always be immediately invoked.

pkg/logs/build_logs.go
```diff
@@ -122,24 +122,20 @@ func (c *BuildLogsClient) watchReadyContainers(ctx context.Context, readyContain
 
 			switch r.Type {
 			case watch.Added, watch.Modified:
-
```
matthewmcnew, line 125: It is a bit unclear to me what happens without these changes. 
Does it tail logs out of order? Tail only the isitio logs?

test/execute_build_test.go
```diff
@@ -131,6 +116,11 @@ func testCreateImage(t *testing.T, when spec.G, it spec.S) {
 					Name: dockerSecret,
 				},
 			},
+			ImagePullSecrets: []corev1.LocalObjectReference{
```
matthewmcnew, line 119: Was this driven by testing this in an environment that needed pull secrets?

pkg/logs/build_logs.go
tomkennedy513: Both. It will tail logs out of order if any other container is ready before prepare. It will also tag every container, even sidecars

test/execute_build_test.go
tomkennedy513: ya, I wanted it to be easy for anyone to run the tests on their own deployment

test/execute_build_test.go
```diff
@@ -82,10 +81,14 @@ func testCreateImage(t *testing.T, when spec.G, it spec.S) {
 
 		deleteNamespace(t, ctx, clients, testNamespace)
 
+		labels := readNamespaceLabelsFromEnv()
```
matthewmcnew, line 84: Is there a reason this should be a specific config option?

pkg/apis/build/v1alpha2/build_pod.go
tomkennedy513: No thats an artifact of the original spike that I just left around. I will remove it

test/execute_build_test.go
tomkennedy513: as opposed to just reading the env when we need it? Not really and I can see why it feels out of place compared to the other config variables

pkg/logs/build_logs.go
tomkennedy513: ya this might be possible, I haven't been able to produce it myself, but in theory this could happen. Maybe there is another check I can do

config/controller.yaml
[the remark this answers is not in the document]
chenbh: Wouldn't it still work with nested fields?
```
data:
  build-waiter:
    image: ...
  rebase:
    image: ...
```

cmd/build-waiter/main.go
```diff
@@ -0,0 +1,169 @@
...
+		return err
+	}
+
+	if needsDelete {
+		if err := os.RemoveAll(to); err != nil {
+			return err
+		}
+	}
+
+	destFile, err := os.OpenFile(to, os.O_RDWR|os.O_CREATE, 0777)
```
tylerphelan, line 114: kinda surprised this one is 0777 perm, should it be 0666?

tomkennedy513: ya thats my bad

docs/injected_sidecars.md
```diff
@@ -0,0 +1,25 @@
+# Support for Injected Sidecars
```
tylerphelan, line 1: nice

tylerphelan (approved):
unbelievable! good concurrency

PR #1023, opened 2022-08-31
Remove secret name from volume naming

tylerphelan:
- Secrets can contain dots while volumes can't
- Volume names also can't be more than 63 characters

Volume names must be valid RFC 1123 Label Names: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

Fixes: https://github.com/pivotal/kpack/issues/759


pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -1194,3 +1221,12 @@ func envs(envs []corev1.EnvVar, envVar corev1.EnvVar) []corev1.EnvVar {
 	}
 	return envs
 }
+
+// Volume names must be valid RFC 1123 Label Names: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names
+func secretNameToVolumeName(secretName string) (string, error) {
+	volumeName := fmt.Sprintf(volumeSecretNameTemplate, secretName)
+	if len(volumeName) > 63 {
+		return "", errors.Errorf("secret name %s is too long, it must contain fewer than 50 characters", secretName)
```
ncarlson, line 1229: Should this "50" be "64"?

tylerphelan: `volumeSecretNameTemplate` is 14 characters

ncarlson (approved):

pkg/apis/build/v1alpha2/build_pod.go
matthewmcnew, line 1229: This effectively prevents long secrets from being used because we are enforcing the volume name to contain the secret name. That feels unnecessary? Why not generate a random name or sequence of volume names volume-1,volume-2,...?

tylerphelan: oh right volume names don't matter across builds, that's what I was concerned about

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -1194,3 +1221,12 @@ func envs(envs []corev1.EnvVar, envVar corev1.EnvVar) []corev1.EnvVar {
...
+	}
+	return strings.ReplaceAll(volumeName, ".", "-"), nil
```
matthewmcnew, line 1231: [This is reaaaally theoretical but, channeling my inner](https://github.com/pivotal/kpack/pull/947#discussion_r846638162) @tomkennedy513 is there a risk for collision here? 

`my.secret` and `my-secret` will have the same volume name?

pkg/apis/build/v1alpha2/build_pod.go
tomkennedy513: I'm in favor of generating completely random volume names

pkg/apis/build/v1alpha2/build_pod.go
tylerphelan: added uniqueness!

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -75,6 +75,10 @@ type BuildPodImages struct {
 	CompletionWindowsImage string
 }
 
+type volumeCounter struct {
```
matthewmcnew, line 78: Can we use the index of the secrets to avoid needing to count the volumes?

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -1194,3 +1200,15 @@ func envs(envs []corev1.EnvVar, envVar corev1.EnvVar) []corev1.EnvVar {
 	}
 	return envs
 }
+
+// Volume names must be valid RFC 1123 Label Names: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names
+func secretNameToVolumeName(secretName string, volCounter *volumeCounter) string {
+	volumeName := fmt.Sprintf(volumeSecretNameTemplate, secretName)
+	// leave space for - and 3 digits
+	if len(volumeName) > 63 {
```
matthewmcnew, line 1208: Is there a real reason to use the secret name in the volume name at all?

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -1194,3 +1200,15 @@ func envs(envs []corev1.EnvVar, envVar corev1.EnvVar) []corev1.EnvVar {
...
+		volumeName = volumeName[:59]
+	}
+	volumeName = fmt.Sprintf("%s-%v", volumeName, volCounter.count)
```
ncarlson, line 1211: Would there be any value in using leading zeros? Like `volume-name-001` instead of `volume-name-1`?

pkg/apis/build/v1alpha2/build_pod.go
tomkennedy513, line 78: can we remove this since it's no longer being used

tylerphelan: whoops

tomkennedy513 (approved):

tomkennedy513 (approved):
can we squash the commits since we aren't doing the name conversion anymore

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -939,10 +939,10 @@ func (b *Build) setupImagePullVolumes(secrets []corev1.LocalObjectReference) ([]
 		volumeMounts []corev1.VolumeMount
 		args         []string
 	)
-	for _, secret := range deduplicate(secrets, b.Spec.Builder.ImagePullSecrets) {
+	for i, secret := range deduplicate(secrets, b.Spec.Builder.ImagePullSecrets) {
 		args = append(args, fmt.Sprintf("-imagepull=%s", secret.Name))
-		volumeName := fmt.Sprintf(volumeSecretNameTemplate, secret.Name)
 
+		volumeName := fmt.Sprintf(pullSecretVolumeNameTemplate, i)
```
matthewmcnew, line 945: So simple!

matthewmcnew (approved):

pkg/apis/build/v1alpha2/build_pod.go
tylerphelan: can't think of anything

PR #1026, opened 2022-09-01
Add support for platform api 0.9 and bump lifecycle to 0.14.1

### LINKED ISSUE #998: Utilize Platform API 0.9 when available
### Background 
When builds are executed with [Platform API 0.9 or greater they will omit the legacy BOM from the output image](https://github.com/buildpacks/lifecycle/issues/808). This will prevent the image from having potentially problematic large label sizes. This should resolve [some issues](https://github.com/pivotal/kpack/issues/893) with reported images built by kpack.

### Additional Context
* This should be documented as a possible breaking change.
* This should be released alongside the bom support being removed from kp. https://github.com/vmware-tanzu/kpack-cli/pull/261

tomkennedy513:
resolves #998

### DIFF
--- a/pkg/apis/build/v1alpha2/build_pod.go
+++ b/pkg/apis/build/v1alpha2/build_pod.go
@@ -998,7 +998,7 @@ func (b *Build) setupCosignVolumes(secrets []corev1.Secret) ([]corev1.Volume, []
 var (
 	lowestSupportedPlatformVersion = semver.MustParse("0.3")
 
-	supportedPlatformAPIVersionsWithWindowsAndReportToml = []*semver.Version{semver.MustParse("0.8"), semver.MustParse("0.7"), semver.MustParse("0.6"), semver.MustParse("0.5"), semver.MustParse("0.4")}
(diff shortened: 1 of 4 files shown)


pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -998,7 +998,7 @@ func (b *Build) setupCosignVolumes(secrets []corev1.Secret) ([]corev1.Volume, []
 var (
 	lowestSupportedPlatformVersion = semver.MustParse("0.3")
 
-	supportedPlatformAPIVersionsWithWindowsAndReportToml = []*semver.Version{semver.MustParse("0.8"), semver.MustParse("0.7"), semver.MustParse("0.6"), semver.MustParse("0.5"), semver.MustParse("0.4")}
+	supportedPlatformAPIVersionsWithWindowsAndReportToml = []*semver.Version{semver.MustParse("0.9"), semver.MustParse("0.8"), semver.MustParse("0.7"), semver.MustParse("0.6"), semver.MustParse("0.5"), semver.MustParse("0.4")}
```
ncarlson, line 1001: Maybe at some point we can replace the "supported versions" list with a semver constraint. 
https://github.com/Masterminds/semver#checking-version-constraints

ncarlson (approved):

pkg/apis/build/v1alpha2/build_pod.go
tomkennedy513: we should also not rely on the order of the list like we do now

tylerphelan (approved):

PR #1027, opened 2022-09-01
Support correct specification of project descriptor

### LINKED ISSUE #1017: Support correct specification of project descriptor
See https://github.com/buildpacks/pack/pull/1479

jjbustamante:
## Problem

The CNB project descriptor v0.2 was [updated](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md#iobuildpacksbuildenv-optional) and the way the environment variables are set was changed. We need to add support for this.

## Solution

A new structured was added for parsing the new way of setting environment variables in the project descriptor, but we also kept the previous one for backward compatibility. 

Fixes #1017 

Signed-off-by: Juan Bustamante <jbustamante@vmware.com>

### DIFF
--- a/pkg/cnb/env_vars.go
+++ b/pkg/cnb/env_vars.go
@@ -1,6 +1,7 @@
 package cnb
 
 import (
+	"github.com/pkg/errors"
 	"io/ioutil"
 	"os"
 	"path"
@@ -26,4 +27,56 @@ func serializeEnvVars(envVars []envVariable, platformDir string) error {
 type envVariable struct {
 	Name  string `json:"name" toml:"name"`
 	Value string `json:"value" toml:"value"`
+}
+
+type buildEnvVariable struct {
+	Env []envVariable `toml:"env"`
+}
+
+type envBuildVariable struct {
+	Env []envVariable
+}
+
+func (a *envBuildVariable) UnmarshalTOML(f interface{}) error {
+	var (
+		env []envVariable
+		err error
+	)
+	switch v := f.(type) {
+	case map[string]interface{}:
+		if envs, ok := v["build"].([]map[string]interface{}); ok {
+			env, err = buildEnv(envs)
+			if err != nil {
+				return err
+			}
+		}
+	case []map[string]interface{}:
+		env, err = buildEnv(v)
+		if err != nil {
+			return err
+		}
+	default:
+		return errors.New("environment variables in project descriptor could not be parsed")
+	}
+	a.Env = env
+	return nil
+}
+
+func buildEnv(v []map[string]interface{}) ([]envVariable, error) {
+	var e []envVariable
+	for _, env := range v {
+		if name, nameOk := env["name"].(string); nameOk {
+			if value, valueOk := env["value"].(string); valueOk {
+				e = append(e, envVariable{
+					Name:  name,
+					Value: value,
+				})
+			} else {
+				return nil, errors.Errorf("environment variable '%s' is not a string value", name)
+			}
+		} else {
+			return nil, errors.New("environment variable 'name' is not a string")
+		}
+	}
+	return e, nil
 }
--- a/pkg/cnb/project_descriptor.go
+++ b/pkg/cnb/project_descriptor.go
@@ -41,7 +41,7 @@ func ProcessProjectDescriptor(appDir, descriptorPath, platformDir string, logger
 	if err := processFiles(appDir, d.IO.Buildpacks.build); err != nil {
 		return err
 	}
-	return serializeEnvVars(d.IO.Buildpacks.Env, platformDir)
+	return serializeEnvVars(d.env(), platformDir)
 }
 
 func parseProjectDescriptor(file string, logger *log.Logger) (descriptorV2, error) {
@@ -71,6 +71,11 @@ func v1ToV2(v1 descriptorV1) descriptorV2 {
 			Buildpacks: cnbTableV2{
 				build: v1.Build.build,
 				Group: v1.Build.Buildpacks,
+				buildEnvVariableV2: buildEnvVariableV2{
+					BuildEnv: buildEnvVariable{
+						Env: v1.Build.Env,
+					},
+				},
 			},
 		},
 	}
@@ -129,6 +134,17 @@ type descriptorV2 struct {
 	IO      ioTable `toml:"io"`
 }
 
+// Because CNB Project Descriptor v0.2 has two ways for defining environment variables.
+// see https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md#iobuildpacksbuildenv-optional
+// This function calculates the final environment variables
+func (d *descriptorV2) env() []envVariable {
+	env := d.IO.Buildpacks.BuildEnv.Env
+	if env == nil {
+		env = d.IO.Buildpacks.EnvBuild.Env
+	}
+	return env
+}
+
 type project struct {
 	SchemaVersion string `toml:"schema-version"`
 }
@@ -138,15 +154,21 @@ type ioTable struct {
 }
 
 type cnbTableV2 struct {
-	build `toml:",inline"`
-	Group []buildpack `toml:"group"`
+	build              `toml:",inline"`
+	buildEnvVariableV2 `toml:",inline"`
+	Group              []buildpack `toml:"group"`
 }
 
 type build struct {
-	Include []string      `toml:"include"`
-	Exclude []string      `toml:"exclude"`
-	Builder string        `toml:"builder"`
-	Env     []envVariable `toml:"env"`
+	Include []string `toml:"include"`
+	Exclude []string `toml:"exclude"`
+	Builder string   `toml:"builder"`
+}
+
+type buildEnvVariableV2 struct {
+	BuildEnv buildEnvVariable `toml:"build"`
+	// Deprecated: use `[[io.buildpacks.build.env]]` instead. see https://github.com/buildpacks/pack/pull/1479
+	EnvBuild envBuildVariable `toml:"env"`
 }
 
 type buildpack struct {
@@ -161,5 +183,6 @@ type descriptorV1 struct {
 
 type cnbTableV1 struct {
 	build      `toml:",inline"`
-	Buildpacks []buildpack `toml:"buildpacks"`
+	Buildpacks []buildpack   `toml:"buildpacks"`
+	Env        []envVariable `toml:"env"`
 }
--- a/pkg/cnb/project_descriptor_test.go
+++ b/pkg/cnb/project_descriptor_test.go
@@ -225,8 +225,73 @@ uri = "check-this-out.com"
 
 			when("using descriptor v2", func() {
 				when("the descriptor has build env vars", func() {
-					it.Before(func() {
-						ioutil.WriteFile(projectToml, []byte(`
+					when("vars where set with older v0.2 project.toml", func() {
+						when("io.buildpacks.env.build format is used", func() {
+							when("format is valid", func() {
+								it.Before(func() {
+									ioutil.WriteFile(projectToml, []byte(`
+[_]
+schema-version = "0.2"
+[[io.buildpacks.env.build]]
+name = "keyA"
+value = "valueA"
+[[io.buildpacks.env.build]]
+name = "keyB"
+value = "valueB"
+[[io.buildpacks.env.build]]
+name = "keyC"
+value = "valueC"
+# check that later keys override previous ones
+[[io.buildpacks.env.build]]
+name = "keyC"
+value = "valueAnotherC"
+				`), 0644)
+								})
+								it("writes all env var files to the platform dir", func() {
+									assert.Nil(t, cnb.ProcessProjectDescriptor(appDir, descriptorPath, platformDir, logger))
+									checkEnvVar(t, platformDir, "keyA", "valueA")
+									checkEnvVar(t, platformDir, "keyB", "valueB")
+									checkEnvVar(t, platformDir, "keyC", "valueAnotherC")
+								})
+							})
+							when("format is invalid", func() {
+								when("'value' is invalid", func() {
+									it.Before(func() {
+										ioutil.WriteFile(projectToml, []byte(`
+[_]
+schema-version = "0.2"
+[[io.buildpacks.env.build]]
+name = "KeyA"
+value = 1
+				`), 0644)
+									})
+									it("writes all env var files to the platform dir", func() {
+										err := cnb.ProcessProjectDescriptor(appDir, descriptorPath, platformDir, logger)
+										assert.EqualError(t, err, "environment variable 'KeyA' is not a string value")
+									})
+								})
+								when("'name' is invalid", func() {
+									it.Before(func() {
+										ioutil.WriteFile(projectToml, []byte(`
+[_]
+schema-version = "0.2"
+[[io.buildpacks.env.build]]
+name = 1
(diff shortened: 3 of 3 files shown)


tomkennedy513 (approved):

pkg/cnb/project_descriptor_test.go
```diff
@@ -243,12 +244,41 @@ value = "valueC"
...
+							checkEnvVar(t, platformDir, "keyC", "valueAnotherC")
+						})
 					})
-					it("writes all env var files to the platform dir", func() {
-						assert.Nil(t, cnb.ProcessProjectDescriptor(appDir, descriptorPath, platformDir, logger))
-						checkEnvVar(t, platformDir, "keyA", "valueA")
-						checkEnvVar(t, platformDir, "keyB", "valueB")
-						checkEnvVar(t, platformDir, "keyC", "valueAnotherC")
+
+					when("vars where set with new v0.2 project.toml", func() {
```
samj1912, line 256: @jjbustamante can we also add a test case where both are set? The new one should have higher precedence

jjbustamante: Interesting.

Yes, let me add that case, thanks @samj1912 !

pkg/cnb/project_descriptor.go
```diff
@@ -105,6 +107,35 @@ func processFiles(appDir string, d build) error {
...
+	} else {
+		if d.IO.Buildpacks.BuildEnv.Env != nil {
+			// both env variables definition were used
+			env = mergeEnvVariables(env, d.IO.Buildpacks.BuildEnv.Env)
+		}
+	}
+	return env
+}
+
+func mergeEnvVariables(old []envVariable, new []envVariable) []envVariable {
```
tylerphelan, line 129: pack appears to just use the newer field instead of merge: https://github.com/buildpacks/pack/pull/1479/files#diff-4216f3c5d147f7361545afe9fb76c1fe23790f314e3f39ae1eca8559d726bd09R52

maybe this can help simplify things

anyone have opinions? @tomkennedy513 @samj1912 @matthewmcnew @ncarlson @yaelharel @chenbh

tomkennedy513: Ya, i think if the newer field is populated just use that

jjbustamante: Do you mean to keep it how it was before?

jjbustamante:
I just added more unit testing to the merge logic. I agreed that pack has not merging logic. If we agree to avoid that one I can revert to my first commit.

@samj1912, @tylerphelan

pkg/cnb/project_descriptor.go
tylerphelan: I think it was backwards before:

pack:
```
env := versionedDescriptor.IO.Buildpacks.Build.Env
if env == nil {
    env = versionedDescriptor.IO.Buildpacks.Env.Build
}
```

before:

```
env := d.IO.Buildpacks.Env
if env == nil {
	env = d.IO.Buildpacks.BuildEnv.Env
}
```

jjbustamante: Yeap, I added the logic before following @samj1912 [comment](https://github.com/pivotal/kpack/pull/1027#discussion_r961150010)

samj1912:
@jjbustamante what I meant by https://github.com/pivotal/kpack/pull/1027#discussion_r961150010 was that if both the fields are set, the new field should have higher precedence and we should ignore the old one. Apologies if it came across as merging the two. We should also have a test case similar to pack where we check that the env vars are from the new field and not the old one.

jjbustamante:
> @jjbustamante what I meant by [#1027 (comment)](https://github.com/pivotal/kpack/pull/1027#discussion_r961150010) was that if both the fields are set, the new field should have higher precedence and we should ignore the old one. Apologies if it came across as merging the two. We should also have a test case similar to pack where we check that the env vars are from the new field and not the old one.

Thanks @samj1912 for the clarification, I removed the merging code and the test case added was also modified. If you defined environment variables in the old version and the on the new version, kpack is going to take ONLY the one defines in the new version.

pkg/cnb/project_descriptor.go
```diff
@@ -105,6 +107,22 @@ func processFiles(appDir string, d build) error {
...
 
+// Because CNB Project Descriptor v0.2 has two ways for defining environment variables.
+// see https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md#iobuildpacksbuildenv-optional
+// This function calculates the final environment variables
+func getEnvVariables(d descriptorV2) []envVariable {
+	env := d.IO.Buildpacks.Env
+	if env == nil {
+		env = d.IO.Buildpacks.BuildEnv.Env
+	} else {
+		if d.IO.Buildpacks.BuildEnv.Env != nil {
```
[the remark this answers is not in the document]
tomkennedy513: Actually my bad, i misread which was the newer key, but couldnt you set teh env to the newer key and if the env is nil then set it to the older key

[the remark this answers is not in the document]
tylerphelan: does this work?

```
func getEnvVariables(d descriptorV2) []envVariable {
  env := d.IO.Buildpacks.BuildEnv.Env
  if env == nil {
    env = d.IO.Buildpacks.Env
  }
  return env
}
```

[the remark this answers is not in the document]
jjbustamante: Yeah, I realized that after pushing hahaha

tomkennedy513 (approved):

ncarlson (approved):

pkg/cnb/project_descriptor_test.go
```diff
@@ -243,12 +244,61 @@ value = "valueC"
...
+							checkEnvVar(t, platformDir, "keyC", "valueAnotherC")
+						})
+					})
+
+					when("vars where set with both versions of v0.2 project.toml", func() {
+						it.Before(func() {
+							ioutil.WriteFile(projectToml, []byte(`
+[_]
+schema-version = "0.2"
+[[io.buildpacks.env]]
```
samj1912, line 289: @jjbustamante shouldn't this be io.buildpacks.env.build?

samj1912, line 289:
```suggestion
[[io.buildpacks.env.build]]
```

samj1912 (changes requested):
The current 'old' code in kpack seems a bit incorrect? See output section of https://github.com/buildpacks/pack/pull/1479

for the old key vs the new key

pkg/cnb/project_descriptor_test.go
jjbustamante: That one is defined in L293 see [here](https://github.com/pivotal/kpack/blob/0e3c15f9d5230ff454dbc1776362e7a3f6c7a966/pkg/cnb/project_descriptor_test.go#L289-L295)

In that case, I defined the same variable twice:

```toml
schema-version = "0.2"
[[io.buildpacks.env]]
name = "keyA"
value = "valueA"
# check that new declaration must have higher precedence
[[io.buildpacks.build.env]]
name = "keyA"
value = "newValueA"
```

And then in the assert, I verify the one taken was `newValueA`

samj1912: @jjbustamante I meant to comment on the key itself. Left some notes at https://github.com/pivotal/kpack/pull/1027#pullrequestreview-1095151602

jjbustamante:
> The current 'old' code in kpack seems a bit incorrect? See output section of [buildpacks/pack#1479](https://github.com/buildpacks/pack/pull/1479)
> 
> for the old key vs the new key
> 
> old pack key - io.buildpacks.env.build
> 
> new pack key - io.buildpacks.build.env
> 
> old kpack key - io.buildpacks.env? (this seems incorrect?)

Oh! That's interesting. @matthewmcnew @tylerphelan @tomkennedy513  any thoughts on that? I didn't touch that part

pkg/cnb/project_descriptor.go
```diff
@@ -105,6 +107,17 @@ func processFiles(appDir string, d build) error {
 	})
 }
 
+// Because CNB Project Descriptor v0.2 has two ways for defining environment variables.
+// see https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md#iobuildpacksbuildenv-optional
+// This function calculates the final environment variables
+func getEnvVariables(d descriptorV2) []envVariable {
+	env := d.IO.Buildpacks.BuildEnv.Env
+	if env == nil {
+		env = d.IO.Buildpacks.Env
```
tylerphelan, line 116: let's start the a thread here per: https://github.com/pivotal/kpack/pull/1027#issuecomment-1235771433

> The current 'old' code in kpack seems a bit incorrect? See output section of https://github.com/buildpacks/pack/pull/1479

It appears we incorrectly used ` io.buildpacks.env` while `pack` incorrectly used `io.buildpacks.env.build`.

We can either support all three or ignore support for `io.buildpacks.env.build`

@samj1912 @tomkennedy513 @matthewmcnew @jjbustamante 
for the old key vs the new key

old pack key - io.buildpacks.env.build

new pack key - io.buildpacks.build.env

old kpack key - io.buildpacks.env? (this seems incorrect?)

matthewmcnew: Kpack exists in the pack ecosystem where users would reasonably expect kpack to behave similarly to pack. So, I think we should provide backwards compatibility to both `io.buildpacks.env.build` and kpack incorrect key: `io.buildpacks.env`.

matthewmcnew: Good 👀 @samj1912 !

samj1912:
So IIUC

highest to lowest precedence -

- io.buildpacks.build.env
- io.buildpacks.env.build
- io.buildpacks.env
?

jjbustamante:
> * io.buildpacks.build.env
> * io.buildpacks.env.build
> * io.buildpacks.env

I agree with the major reason to do it like this, but, if I am not wrong, I am not able to use the same *structs* to deal with the 3 cases. The struct to handle `build.env` and `env.build` can be the same, with two fields but if I get an error trying to parse the toml file then I will have to create a different *struct* to try to parse the case `env`, right?

jjbustamante:
I am not totally happy, but I did a refactor to include the compatibility for all the 3 cases:

- io.buildpacks.build.env
- io.buildpacks.env.build
- io.buildpacks.env

If anyone have suggestions about the approach, let me know.

pkg/cnb/project_descriptor.go
```diff
@@ -41,13 +41,22 @@ func ProcessProjectDescriptor(appDir, descriptorPath, platformDir string, logger
...
+	env := getEnvVariables(d)
+	return serializeEnvVars(env, platformDir)
 }
 
 func parseProjectDescriptor(file string, logger *log.Logger) (descriptorV2, error) {
 	var d descriptorV2
 	if _, err := toml.DecodeFile(file, &d); err != nil {
-		return descriptorV2{}, err
+		var dV2NoSpec descriptorV2NoSpec
+		if _, err2 := toml.DecodeFile(file, &dV2NoSpec); err2 != nil {
```
matthewmcnew, line 53: Would it be easier to use the `encoding.TextUnmarshaler` interface to add logic to unmarshall the different types of `env` fields to avoid a whole different type?

pkg/cnb/project_descriptor.go
```diff
@@ -105,6 +119,17 @@ func processFiles(appDir string, d build) error {
 	})
 }
 
+// Because CNB Project Descriptor v0.2 has two ways for defining environment variables.
+// see https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md#iobuildpacksbuildenv-optional
+// This function calculates the final environment variables
+func getEnvVariables(d descriptorV2) []envVariable {
```
matthewmcnew, line 125: Can we add this instead as a method on `descriptorV2` so that consumers of  descriptorV2 don't need to care about the heirachy of env and can just ask ` descriptor.Env()`?

jjbustamante: Done

pkg/cnb/project_descriptor.go
jjbustamante: @matthewmcnew you mean to use [this](https://github.com/BurntSushi/toml#using-the-marshaler-and-encodingtextunmarshaler-interfaces)?

matthewmcnew: Yeah!

jjbustamante: Thanks for the help @matthewmcnew. Done

pkg/cnb/env_vars.go
```diff
@@ -27,3 +28,37 @@ type envVariable struct {
...
+
+type buildEnvVariable struct {
+	Env []envVariable `toml:"env"`
+}
+
+type envBuildVariable struct {
+	Env []envVariable
+}
+
+func (a *envBuildVariable) UnmarshalTOML(f interface{}) error {
```
tylerphelan, line 40: this will prioritize `io.buildpacks.build` over legacy pack's `io.buildpacks.env.build` but I think it's ok

jjbustamante: In fact, I think it is not giving any priority, it is just parsing both cases depending on what the user specifies because they go to different **case** in the switch.

tylerphelan: yea I think this is ok b/c `io.buildpacks.env` and `io.buildpacks.env.build` is not valid toml

tomkennedy513 (approved):

pkg/cnb/env_vars.go
```diff
@@ -27,3 +28,37 @@ type envVariable struct {
...
+	default:
+		return errors.New("environment variables in project descriptor could not be parsed")
+	}
+}
+
+func buildEnv(v []map[string]interface{}) []envVariable {
+	var e []envVariable
+	for _, env := range v {
+		e = append(e, envVariable{
+			Name:  env["name"].(string),
```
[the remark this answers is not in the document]
matthewmcnew: So, I think we need to add the valid cast check.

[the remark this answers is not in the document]
matthewmcnew: Sorry, I dropped that for brevity earlier.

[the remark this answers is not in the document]
jjbustamante: Well, if I add a test case with the following configuration

``` toml
schema-version = "0.2"
[[io.buildpacks.env.build]]
name = keyA
value = "valueA"
```
It actually fails with the following message: "toml: line 5 (last key \"io.buildpacks.env.build.name\"): expected value but found \"keyA\" instead" 

It is actually not reaching the **UnmarshalTOML** implementations, it fails [here](https://github.com/BurntSushi/toml/blob/891d2617ddbdfa265c4095b53103c010c98e6591/parse.go#L98)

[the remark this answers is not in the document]
tylerphelan: try with an int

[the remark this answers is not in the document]
jjbustamante: With an **int** it actually gets into the code we want to protect, I just added 4 more test cases and some validations

matthewmcnew (approved):

tylerphelan (approved):

tomkennedy513:
@jjbustamante can you rebase this pr since its out of date

jjbustamante:
> @jjbustamante can you rebase this pr since its out of date

Done @tomkennedy513
I also squashed the commits.

PR #1029, opened 2022-09-06
Add configurable storage class name for image resource

tylerphelan:
- If set, will be used as the storageClassName in the pvc

### DIFF
--- a/docs/image.md
+++ b/docs/image.md
@@ -14,6 +14,7 @@ The following defines the relevant fields of the `image` resource spec in more d
 - `source`: The source code that will be monitored/built into images. See the [Source Configuration](#source-config) section below.
 - `cache`: Caching configuration, two variants are available:
   - `volume.size`: Creates a Volume Claim of the given size
+  - `volume.storageClassName`: (Optional) Creates a Volume Claim of the given storageClassName. If unset, the default storage class is used. The field is immutable.
   - `registry.tag`: Creates an image with cached contents
 - `failedBuildHistoryLimit`: The maximum number of failed builds for an image that will be retained.
 - `successBuildHistoryLimit`: The maximum number of successful builds for an image that will be retained.
@@ -254,6 +255,7 @@ spec:
   cache:
     volume:
       size: "1.5Gi" # Optional, if not set then the caching feature is disabled
+      storageClassName: "my-storage-class" # Optional, if not set then the default storageclass is used
   failedBuildHistoryLimit: 5 # Optional, if not present defaults to 10
   successBuildHistoryLimit: 5 # Optional, if not present defaults to 10
   source:
@@ -299,6 +301,7 @@ spec:
   cache:
     volume:
       size: "1.5Gi" # Optional, if not set then the caching feature is disabled
+      storageClassName: "my-storage-class" # Optional, if not set then the default storageclass is used
   failedBuildHistoryLimit: 5 # Optional, if not present defaults to 10
   successBuildHistoryLimit: 5 # Optional, if not present defaults to 10
   source:
--- a/pkg/apis/build/v1alpha2/image_builds.go
+++ b/pkg/apis/build/v1alpha2/image_builds.go
@@ -205,6 +205,10 @@ func (im *Image) CacheName() string {
 }
 
 func (im *Image) BuildCache() *corev1.PersistentVolumeClaim {
+	var storageClassName *string
+	if im.Spec.Cache.Volume.StorageClassName != "" {
+		storageClassName = &im.Spec.Cache.Volume.StorageClassName
+	}
 	return &corev1.PersistentVolumeClaim{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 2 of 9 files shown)


jjbustamante (commented):
It looks good to me!

tomkennedy513:
I'm wondering if we actually need to validate that the storage class exists since that will require additional rbac for our controller. If we create a pod with an invalid storage class can't we just let pod fail?

pkg/reconciler/image/image.go
```diff
@@ -145,6 +151,15 @@ func (c *Reconciler) reconcileImage(ctx context.Context, image *buildapi.Image)
 		return image, nil
 	}
 
+	if image.Spec.NeedVolumeCache() && image.Spec.Cache.Volume.StorageClassName != "" {
+		_, err := c.StorageClassLister.Get(image.Spec.Cache.Volume.StorageClassName)
+		if err != nil {
```
matthewmcnew, line 156: What is the reason to add this validation?

tylerphelan: @matthewmcnew @tomkennedy513 

I think this is an error that would only show up in the kpack-controller logs and result in a pending Image which is an error we should bubble up more

tomkennedy513: could we add it to the status of the image

tylerphelan: > could we add it to the status of the image

That was my intention. I think it should be ready: false with a message

tomkennedy513: ya I think that works. we should just make sure we document the required rbac change to the kpack controller

tomkennedy513 (approved):

pkg/reconciler/image/image.go
matthewmcnew: > I think this is an error that would only show up in the kpack-controller logs and result in a pending Image which is an error we should bubble up more

I am not sure I understand. Won't this be written to the status and prevent future reonciles?

tomkennedy513: I don't think the pvc will enter a failed state if the storage class does not exist, it just stays pending

matthewmcnew: After pondering this for a few hours, I don't think it makes sense for the image reconciler to validate what is and is not a valid storageClass or persistent volume configuration. It seems that should be the responsibility of the PersistentVolumeClaim resource.

tylerphelan: removed!

chenbh (approved):

matthewmcnew:
Does this need to be supported in v1alpha1 apis? https://github.com/pivotal/kpack/pull/934#issuecomment-1063374631

tylerphelan:
@matthewmcnew 
>Does this need to be supported in v1alpha1 apis? https://github.com/pivotal/kpack/pull/934#issuecomment-1063374631

converting v1alpha1 -> v1alpha2:
`storageClassName` is always "", v1alpha2 clients will use ""

converting v1alpha2 -> v1alpha1:
v1alpha1 client (old version of `kp`) patches existing v1alpha2 image, causing `storageClassName` to be set to "" which will violate the immutability constraint causing blocking error. This is not great.

My 2c is that if an old v1alpha1 client runs into an immutability error it will not lose data. If the storageClassName is unset in v1alpha2 behavior will not be affected by v1alpha1 clients. Users should upgrade kp.

matthewmcnew:
> v1alpha1 client (old version of kp) patches existing v1alpha2 image, causing storageClassName to be set to "" which will violate the immutability constraint causing blocking error. This is not great.

This seems rare but, not great. It essentially means we are not supporting v1alpha1.

tylerphelan:
>It essentially means we are not supporting v1alpha1.

risk seems very high

>If our strategy for v1alpha1 headaches is to say not use v1alpha1 we should probably just begin removing v1alpha1.

If we don't do this will we need to migrate all new fields to prevent data loss? ex `DefaultProcess`, `ProjectDescriptorPath`, `CreationTime` ?

matthewmcnew:
> If we don't do this will we need to migrate all new fields to prevent data loss? ex DefaultProcess, ProjectDescriptorPath, CreationTime

We probably should provide support for these. Hitting an expected immutability constraint seems slightly worse to me.

tylerphelan:
>Hitting an expected immutability constraint seems slightly worse to me.

I think data loss is equal if not worse

tylerphelan:
I think we should probably add migrations

matthewmcnew:
> I think we should probably add migrations

Do you think we should do that as this PR or create an issue to support it for all susceptible fields?

tylerphelan:
I've added the conversion for this one here and have added https://github.com/pivotal/kpack/issues/1036

tomkennedy513 (approved):

tylerphelan:
will merge after 0.7.1 is released

matthewmcnew (approved):

Issue #972, opened 2022-06-13
Support new Kubernetes PodSecurityStandards


danail-branekov:
Hi kpack,

[Korifi](https://github.com/cloudfoundry/korifi) here. As of Kubernetes 1.21 pod security policies are deprecated and are going to be removed in 1.25. They are going to be replaced by the [Pod security admission controller](https://kubernetes.io/docs/concepts/security/pod-security-admission/) and [pod security standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/). This means that pod security is going to be configured per namespace rather than per pod.

In a scenario where a namespace has been configured to enforce the `restricted` pod security standard, kpack build pod that runs in that namespace would fail because its security context does not comply with the `restricted` standard. We had an [explore spike](https://github.com/cloudfoundry/korifi/issues/1202) to figure out how to address that and came up with a [change](https://github.com/cloudfoundry/korifi/issues/1202#issuecomment-1151318859) that makes kpack build pods work in such a namespace.

With this issue we would like ot start a discussion on supporting the pod security standard. Some preliminary input from our side:
* Maybe always enforcing the `restricted` pod security standard is not ideal because there might be buildpacks that do not work in such an environment
* Kpack could look at the namespace's `pod-security` labels and configure the build pods according to the namespace security standard
* Alternative, kpack could be configured to enforce security standards on the build pods via some helm configu value. However, as security standards are configured per namespace, such a global configuration could not always work.

tylerphelan:
I think we should enforce the `restricted` standard by default potentially allowing it to be overwritten in the build config.

As for namespace-wide config we have a desire to add some namespace or global config to kpack builds which could help with namespaced policies.

danail-branekov:
Enforcing the restricted standard generally makes sense but I am concerned with buildpacks out there that do insecure stuff, e.g. running as root. After enforcing the restricted security, those buildpacks would start failing. That is why I suggested some sort of an emergency setting that users could override, however they must realise that doing so carries security implications.

matthewmcnew:
All Cloud Native Buildpack Builds should work with the restricted pod security standard. I don't think there is a need for modifying the pods based on the namespace's pod security standards.

PR #1030, opened 2022-09-06
Conform to restricted pod security standards

tomkennedy513:
resolves: #972

### DIFF
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -65,12 +65,25 @@ spec:
         app: kpack-controller
         version: #@ data.values.version
     spec:
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: "RuntimeDefault"
       priorityClassName: kpack-control-plane
       serviceAccountName: controller
       nodeSelector:
         kubernetes.io/os: linux
       containers:
       - name: controller
+        securityContext:
+          runAsNonRoot: true
+          allowPrivilegeEscalation: false
+          privileged: false
+          seccompProfile:
+            type: "RuntimeDefault"
+          capabilities:
+            drop:
+              - ALL
         image: #@ data.values.controller_image
         env:
         - name: ENABLE_PRIORITY_CLASSES
--- a/pkg/apis/build/v1alpha2/build_pod.go
+++ b/pkg/apis/build/v1alpha2/build_pod.go
@@ -292,6 +292,7 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
 				return []string{b.Tag()}
 			}(),
 		),
+		SecurityContext: containerSecurityContext(buildContext.BuildPodBuilderConfig),
 		VolumeMounts: volumeMounts([]corev1.VolumeMount{
 			layersMount,
 			workspaceVolume,
@@ -334,6 +335,7 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
 		Env: []corev1.EnvVar{
 			platformApiVersionEnvVar,
 		},
+		SecurityContext: containerSecurityContext(buildContext.BuildPodBuilderConfig),
 	}
 	detectContainerMods := ifWindows(buildContext.os(), addNetworkWaitLauncherVolume(), useNetworkWaitLauncher(dnsProbeHost))
 	return &corev1.Pod{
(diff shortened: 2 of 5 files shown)


config/controller.yaml
```diff
@@ -65,12 +65,25 @@ spec:
         app: kpack-controller
         version: #@ data.values.version
     spec:
+      securityContext:
```
matthewmcnew, line 68: Do we expect to remove the overlays in the corresponding package for kpack? https://github.com/vmware-tanzu/package-for-kpack/blob/main/config/overlay.yaml#L8

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -657,15 +664,36 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
 	}, nil
 }
 
+func boolPointer(b bool) *bool {
```
matthewmcnew, line 667: This could have been our first generic 😆

pkg/apis/build/v1alpha2/build_pod_test.go
```diff
@@ -2703,6 +2718,224 @@ func testBuildPod(t *testing.T, when spec.G, it spec.S) {
...
+
+				validateProcMount(pod.Spec.InitContainers)
+				validateProcMount(pod.Spec.Containers)
+				validateProcMount(rebasePod.Spec.InitContainers)
+				validateProcMount(rebasePod.Spec.Containers)
+			})
+			it("uses allowed sysctl", func() {
+				validateProcMount := func(pod *corev1.Pod) {
+					for _, sysctl := range pod.Spec.SecurityContext.Sysctls {
+						or(assert.Equal(t, "kernel.shm_rmid_forced", sysctl.Name),
```
matthewmcnew, line 2853: What is the or doing here?

tomkennedy513: I was trying to assert that multiple values are valid

tomkennedy513: but now that i think of it, any call to assert that fails will result in the error being set in t so this may not work

matthewmcnew: At this point can we validate that restricted pod security standard is met by just checking that no Sysctls are set?

tomkennedy513: true, but i was hoping to validate the full spec so that if we make changes it will cause the test to fail

tomkennedy513: we can do that and in the future if we need to use sysctl we can expand the test to match the actual spec

pkg/apis/build/v1alpha2/build_pod_test.go
```diff
@@ -2757,3 +2990,17 @@ func fetchEnvVar(envVars []corev1.EnvVar, name string) (corev1.EnvVar, bool) {
...
+	for _, b := range bools {
+		if b {
+			return true
+		}
+	}
+
+	return false
+}
+
+func boolPointer(b bool) *bool {
```
ncarlson, line 3004: Is there a place for "helper" functions to live, to avoid [redeclaration](https://github.com/pivotal/kpack/blob/4be6c3d36dd19025ca245a9e91f338526f8bdbbb/pkg/apis/build/v1alpha2/build_pod.go#L667)?

ncarlson (approved):

pkg/apis/build/v1alpha2/build_pod_test.go
tomkennedy513: I just asserted that those fields are unset for now

chenbh:
Does it make sense to add the label(s) to the `100-namespace.yaml` so that it's actually enforced?

pkg/apis/build/v1alpha2/build_pod_test.go
tomkennedy513: I don't believe there is a specific kpack wide place for helpers. we have some helpers in various packages they are used in.

tomkennedy513:
> Does it make sense to add the label(s) to the `100-namespace.yaml` so that it's actually enforced?

ya maybe we should do that. I originally was thinking that users would do that if required, but maybe its good for us to just do it for our controllers. Users will have to enable for the build pods namespace still

chenbh (approved):

tylerphelan (approved):

tylerphelan:
> > Does it make sense to add the label(s) to the `100-namespace.yaml` so that it's actually enforced?
> 
> ya maybe we should do that. I originally was thinking that users would do that if required, but maybe its good for us to just do it for our controllers. Users will have to enable for the build pods namespace still

I think it's ok for now

PR #1032, opened 2022-09-13
Enable setting SOURCE_DATA_EPOCH

### LINKED ISSUE #1009: Enable configuring the creation time
Hi,
Gitlab container registry displays the creation time on the GUI, so I am always shown that all container creation times are 1980.
I'd like to use buildpack's `--creation-time` option to work around this problem, but it isn't implemented in kpack yet.
Do you have a plan to implement `--creation-time` option?

Related ticket:
https://github.com/buildpacks/pack/issues/931
https://github.com/buildpacks/pack/issues/1281

yaelharel:
This environment variable is the timestamp for created time in app image config

Fixes #1009


pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -1211,9 +1219,12 @@ func cosignSecretArgs(secret corev1.Secret) []string {
...
-		envs = append(envs, envVar)
+func envs(envs []corev1.EnvVar, envVar1 corev1.EnvVar, envVar2 corev1.EnvVar) []corev1.EnvVar {
+	if envVar1.Name != "" && envVar1.Value != "" {
+		envs = append(envs, envVar1)
+	}
+	if envVar2.Name != "" && envVar2.Value != "" {
+		envs = append(envs, envVar2)
 	}
 	return envs
 }
```
tomkennedy513, line 1230: Maybe something like this so we can support more than 2 env vars in the future

```suggestion
func envs(envs []corev1.EnvVar, envVars ...corev1.EnvVar) []corev1.EnvVar {
	for _, envVar := range envVars {
		if envVar.Name != "" && envVar.Value != "" {
			envs = append(envs, envVar)
		}
	}
	return envs
}
```

yaelharel: I meant to change it and I forgot.
Thanks!

tomkennedy513:
Do we want this feature to be configurable at the controller level or at least the image level

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -1211,9 +1219,12 @@ func cosignSecretArgs(secret corev1.Secret) []string {
 	return cosignArgs
 }
 
-func envs(envs []corev1.EnvVar, envVar corev1.EnvVar) []corev1.EnvVar {
-	if envVar.Name != "" && envVar.Value != "" {
-		envs = append(envs, envVar)
+func envs(envs []corev1.EnvVar, envVar1 corev1.EnvVar, envVar2 corev1.EnvVar) []corev1.EnvVar {
```
ncarlson: Would it be useful to have `envs` accept a variable number of arguments? Something like:
```golang
func envs(envs []corev1.EnvVar, envVars ...corev1.EnvVar) []corev1.EnvVar {
	for _, envVar := range envVars {
		if envVar.Name != "" && envVar.Value != "" {
			envs = append(envs, envVar)
		}
	}
}
```

ncarlson (approved):

samj1912:
Should we try and match the pack behavior for this? i.e. allowing users to provide a iso timestamp or the word `now`?

yaelharel:
Hi @samj1912, thanks for your comment.
We thought about adding this option in [kpack-cli](https://github.com/vmware-tanzu/kpack-cli/issues/267).
What do you think?

matthewmcnew:
I think kpack needs to support the same configuation options as pack and this should be a non-boolean field that accepts similar inputs to pack's `--creation-time` flag.

matthewmcnew:
> We thought about adding this option in https://github.com/vmware-tanzu/kpack-cli/issues/267.

@yaelharel  A `--creation-time` flag on the `kp` cli would require exposing that functionality on the image/build spec as well, right?

tylerphelan:
@samj1912 @matthewmcnew @yaelharel 

does using a timestamp make sense as an input to the `image` resource? or only the `build` resource? I supposed it doesn't hurt..

tylerphelan:
@yaelharel let's support both unix timestamp and `now` because it's kinda nice to just pass values to the lifecycle

pkg/apis/build/v1alpha2/image_validation.go
```diff
@@ -164,6 +165,14 @@ func (ib *ImageBuild) Validate(ctx context.Context) *apis.FieldError {
 		}
 	}
 
+	if ib.CreationTime != "" && ib.CreationTime != "now" {
+		// check that the timestamp in CreationTime is in valid format
+		_, err := strconv.ParseInt(ib.CreationTime, 10, 64)
```
matthewmcnew, line 170: Nice!

matthewmcnew (approved):

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -592,6 +600,12 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
 								homeEnv,
 								platformApiVersionEnvVar,
 							},
+							func() corev1.EnvVar {
+								if dateTime != nil {
+									return corev1.EnvVar{Name: "SOURCE_DATE_EPOCH", Value: strconv.Itoa(int(dateTime.Unix()))}
+								}
+								return corev1.EnvVar{Name:"", Value:""}
```
tylerphelan, line 607: does this set an empty env var in the pod?

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -1211,9 +1225,29 @@ func cosignSecretArgs(secret corev1.Secret) []string {
 	return cosignArgs
 }
 
-func envs(envs []corev1.EnvVar, envVar corev1.EnvVar) []corev1.EnvVar {
-	if envVar.Name != "" && envVar.Value != "" {
-		envs = append(envs, envVar)
+func envs(envs []corev1.EnvVar, envVars ...corev1.EnvVar) []corev1.EnvVar {
+	for _, envVar := range envVars {
+		if envVar.Name != "" && envVar.Value != "" {
```
tylerphelan, line 1230: I see, empty env vars are filtered out here

tomkennedy513 (approved):
This is great!

tylerphelan:
let's hold off on merging this until kpack 0.7.1 is released

PR #1057, opened 2022-10-31
Add example of unix timestamp to docs for creation time

tomkennedy513:


chenbh (commented):
Is it worth mentioning the empty string is treated as epoch 0?

matthewmcnew (approved):

tylerphelan (approved):

matthewmcnew (approved):

tomkennedy513:
> Is it worth mentioning the empty string is treated as epoch 0?

I don't think the timestamp is passed to the build if it is empty string

PR #1067, opened 2022-11-10
Git: parse short scp-like syntax into proper SSH URLs

### LINKED ISSUE #1066: Build kpack controller with minimal C library dependencies
## Context

To support git repos, the kpack-controller (and build-init) uses https://github.com/libgit2/git2go (which is a Go wrapping around the C library libgit2) to programatically do git operations.

Like most C libraries, libgit2 is able to compile with different options for certain features, one of them is using libssh2 for SSH stuff. Currently, when we compile the kpack controller, we use another proprietary/commercial operator to create a custom stack with the libssh2 ubuntu bionic package installed.  

## Problem

We want to upgrade our base stack to ubuntu jammy and don't want to keep using the custom stack operator, so we need to figure out how do provide the SSH functionality.

Luckily, if libssh2 isn't used (or available), git2go is able to use Go's SSH implementation as the driver. However, this will break any repos using the short scp-like syntax of `[user@]host:path` (e.g. `git@github.com:pivotal/kpack.git`) URL since it's not an [official URL syntax](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols#_the_ssh_protocol).

## Proposed solution

1. Use the Go SSH implementation and modify the kpack-controller to silently convert short scp-like syntax to proper SSH URLs
1. Build the package ourselves from source
1. Figure out a way to get the jammy package installed onto the build image (maybe see if the full stack is willing to add it as a package?)

## Why all this trouble instead of using go-git?

See #213, and https://fluxcd.io/flux/components/source/gitrepositories/#git-implementation

chenbh:
fixes #1066

Parse all git URLs before passing to git2go. The parser will convert any short scp-like syntax to proper SSH style ones and leave all other URLs the same.

### DIFF
--- a/pkg/git/fetch.go
+++ b/pkg/git/fetch.go
@@ -24,7 +24,7 @@ func (f Fetcher) Fetch(dir, gitURL, gitRevision, metadataDir string) error {
 	}
 	defer repository.Free()
 
-	remote, err := repository.Remotes.CreateWithOptions(gitURL, &git2go.RemoteCreateOptions{
+	remote, err := repository.Remotes.CreateWithOptions(parseURL(gitURL), &git2go.RemoteCreateOptions{
 		Name:  "origin",
 		Flags: git2go.RemoteCreateSkipInsteadof,
 	})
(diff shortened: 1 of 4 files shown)


tomkennedy513 (approved):
Nice!

tylerphelan (approved):

yaelharel (approved):

PR #1071, opened 2022-11-17
Set acceptance test resource request to 128m

tomkennedy513:
- This value is the same for istio and non istio tests so this should work in both cases

### DIFF
--- a/test/execute_build_test.go
+++ b/test/execute_build_test.go
@@ -303,7 +303,7 @@ func testCreateImage(t *testing.T, when spec.G, it spec.S) {
 				corev1.ResourceMemory: resource.MustParse("1G"),
 			},
 			Requests: corev1.ResourceList{
-				corev1.ResourceMemory: resource.MustParse("256M"),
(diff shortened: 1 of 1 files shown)


test/execute_build_test.go
```diff
@@ -407,7 +407,7 @@ func generateRebuild(ctx *context.Context, t *testing.T, cfg config, clients *cl
 			corev1.ResourceMemory: resource.MustParse("1G"),
 		},
 		Requests: corev1.ResourceList{
-			corev1.ResourceMemory: resource.MustParse("512M"),
+			corev1.ResourceMemory: resource.MustParse("128m"),
```
yaelharel, line 410:
```suggestion
			corev1.ResourceMemory: resource.MustParse("128M"),
```

tomkennedy513: good catch!

yaelharel (approved):

tylerphelan (approved):

PR #1080, opened 2022-11-30
Retry if image fetch fails due to network error

### LINKED ISSUE #1035: If the lifecycle image fetch fails due to network error, it is not retried
errors such as `i/o timeout` will not be retried for the lifecycle image provider. The builder reconciler will retry but it will not attempt to re-pull the lifecycle image.

yaelharel:
Fixes #1035


cmd/controller/main.go
```diff
@@ -129,6 +129,7 @@ func main() {
 	k8sInformerFactory := informers.NewSharedInformerFactory(k8sClient, options.ResyncPeriod)
 	pvcInformer := k8sInformerFactory.Core().V1().PersistentVolumeClaims()
 	podInformer := k8sInformerFactory.Core().V1().Pods()
+	configmapInformer := k8sInformerFactory.Core().V1().ConfigMaps()
```
matthewmcnew, line 132: Can we restrict this to only operate on the namespaces we care about?

pkg/config/lifecycle_provider.go
```diff
@@ -58,17 +66,17 @@ func (l *LifecycleProvider) LayerForOS(os string) (v1.Layer, cnb.LifecycleMetada
 	}
 }
 
-func (l *LifecycleProvider) UpdateImage(cm *corev1.ConfigMap) {
+func (l *LifecycleProvider) UpdateImage(cm *corev1.ConfigMap) error {
 	lifecycle, err := l.read(context.Background(), cm)
 	if err != nil {
-		l.lifecycleData.Store(configmapRead{err: err})
```
matthewmcnew, line 64: Do we not need to store the error case?

pkg/config/lifecycle_provider.go
```diff
@@ -40,6 +40,14 @@ func NewLifecycleProvider(client RegistryClient, keychainFactory registry.Keycha
 	}
 }
 
+func (l *LifecycleProvider) Digest() (string, error) {
```
matthewmcnew, line 43: What is this Digest method for?

yaelharel: We used it in a previous version and forgot to delete it.
Thanks!

pkg/config/lifecycle_provider.go
yaelharel: Yes we do. Thanks!

yaelharel: Please see 2ef676cd5739484f47a67c05e77a1341670b5dff.

cmd/controller/main.go
yaelharel: Sure. Please see cc165a571de750365bf98ec581bbf0ff9794ffd3.
Thanks!

matthewmcnew (approved):

chenbh (approved):

PR #1082, opened 2022-12-02
Track dependency objects even if they don't exist

### LINKED ISSUE #1028: Resources are not tracked if they are created out-order
Slack: https://kubernetes.slack.com/archives/C02UKAM5F2P/p1662134975214189

The Builders aren't currently able to track a ClusterStack if it is created after the Builder

gcapizzi:
This _should_ fix #1028.

In order to be able to track objects that don't exist yet, `tracker.Tracker` tracks objects by a `reconciler.Key` consisting of its [`schema.GroupKind`](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime/schema#GroupKind) and its [`types.NamespacedName`](https://pkg.go.dev/k8s.io/apimachinery/pkg/types#NamespacedName).

Given the `schema.GroupKind` is derived from the object `TypeMeta`, I made sure that every object instantiated in tests has it set. I believe this should also be the case in production.

In order to make sure the `schema.GroupKind` was properly taken into account, I modified the `testhelpers.FakeTracker` to use the full `reconciler.Key` as its tracking key.

I wasn't able to run the end-to-end tests locally, unfortunately. Unit tests are passing but it would be nice if someone on the team could check the end-to-ends are also happy.

Please take a look and let me know how it looks!

### DIFF
--- a/pkg/reconciler/builder/builder.go
+++ b/pkg/reconciler/builder/builder.go
@@ -5,6 +5,8 @@ import (
 
 	"go.uber.org/zap"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"knative.dev/pkg/logging/logkey"
 
 	"github.com/google/go-containerregistry/pkg/authn"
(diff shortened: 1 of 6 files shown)


matthewmcnew (approved):

matthewmcnew:
Thanks @gcapizzi

PR #1121, opened 2023-01-19
Use parsed ssh username if git2go does not provide one

tomkennedy513:
This is required because git2go is unable to parse the username when calling `remote.Fetch` with the go ssh transport


chenbh (changes requested):
Can we add either a code comment or commit message that explains the rationale behind this change? I fear the context is going to be lost especially since we don't have an issue explaining the problem.

chenbh (approved):

Issue #1124, opened 2023-01-23
StoreBuildpackRepository.keychain is empty inside RemoteBuilder


chenbh:
Cannot push builders that reference a store with a buildpack from a different registry.

Because we use an empty keychain when creating the BuildpackRepository, it'll get propagated to the lazyMountableLayer, which means that when we save an image with this layer, it'll fail to auth to the ClusterStore registry. https://github.com/pivotal/kpack/blob/c46ce3e63eaed3b0586b8c7cbfc239c9e4a06539/cmd/controller/main.go#L193

```yaml
apiVersion: v1
imagePullSecrets:
- name: registry-1
kind: ServiceAccount
metadata:
  name: pull-sa
secrets:
- name: registry-1
```

```yaml
apiVersion: v1
imagePullSecrets:
- name: registry-2
kind: ServiceAccount
metadata:
  name: push-sa
secrets:
- name: registry-2
```

```yaml
apiVersion: kpack.io/v1alpha2
kind: ClusterStore
metadata:
  name: test
spec:
  serviceAccountRef:
    name: pull-sa
  sources:
  - image: registry-1/my-buildpack
```

```yaml
apiVersion: kpack.io/v1alpha2
kind: ClusterBuilder
metadata:
  name: test
spec:
  serviceAccountRef:
    name: push-sa
  stack:
    kind: ClusterStack
    name: default
  store:
    kind: ClusterStore
    name: test
  tag:  registry-2/my-builder
  order:
  - group:
    - id: some-buildpack-from-clusterstore
```

In this case, the ClusterBuilder will fail to reconcile because it will fail to upload the blobs over to registry-2. Note that even if we add the `registry-1` secret to the `push-sa`, it'll still fail.

tomkennedy513:
May be resolved by buildpack resource work

PR #1128, opened 2023-02-01
Implement Buildpack resources

### LINKED ISSUE #1081: Implement RFC 9: Buildpacks resource
RFC [proposal](https://github.com/pivotal/kpack/blob/main/rfcs/0009-buildpack-resource.md) and [discussion].(https://github.com/pivotal/kpack/pull/931)

- [ ] introduce a v1alpha3 API with the ClusterStore resource deprecated
- [ ] implement [Cluster]Buildpack resource and reconciler
- [ ] have [Cluster]Builder make use of the new resource

chenbh:
closes #1081, #1124 

I highly recommend reviewing this by commit

Summary of changes:
- Add CRDs and reconcilers for [Cluster]Buildpacks, the implementation is near identical to the ClusterStore
- Split `RemoteStoreReader` into 2 components, one that is responsible for resolving a reference to a specific buildpack status, and the other one responsible for taking the buildpack status and constructing a remote image for it
- Update the [Cluster]Builder reconcilers to support referencing buildpacks

Todo:
- [x] Add integration tests
- [x] Add documentation

### DIFF
--- a/pkg/cnb/create_builder.go
+++ b/pkg/cnb/create_builder.go
@@ -8,7 +8,6 @@ import (
 	buildapi "github.com/pivotal/kpack/pkg/apis/build/v1alpha2"
 	corev1alpha1 "github.com/pivotal/kpack/pkg/apis/core/v1alpha1"
 	"github.com/pivotal/kpack/pkg/registry"
-	k8scorev1 "k8s.io/api/core/v1"
 )
 
 type RegistryClient interface {
@@ -32,22 +31,22 @@ func (r *RemoteBuilderCreator) CreateBuilder(
 	builderKeychain authn.Keychain,
 	fetcher RemoteBuildpackFetcher,
 	clusterStack *buildapi.ClusterStack, spec buildapi.BuilderSpec,
-) ([]k8scorev1.ObjectReference, buildapi.BuilderRecord, error) {
+) (buildapi.BuilderRecord, error) {
 	buildImage, _, err := r.RegistryClient.Fetch(builderKeychain, clusterStack.Status.BuildImage.LatestImage)
 	if err != nil {
-		return nil, buildapi.BuilderRecord{}, err
+		return buildapi.BuilderRecord{}, err
 	}
 
 	builderBldr := newBuilderBldr(r.KpackVersion)
 
 	err = builderBldr.AddStack(buildImage, clusterStack)
 	if err != nil {
-		return nil, buildapi.BuilderRecord{}, err
+		return buildapi.BuilderRecord{}, err
 	}
 
 	lifecycleLayer, lifecycleMetadata, err := r.LifecycleProvider.LayerForOS(builderBldr.os)
 	if err != nil {
-		return nil, buildapi.BuilderRecord{}, err
+		return buildapi.BuilderRecord{}, err
 	}
 
 	builderBldr.AddLifecycle(lifecycleLayer, lifecycleMetadata)
@@ -58,7 +57,7 @@ func (r *RemoteBuilderCreator) CreateBuilder(
 		for _, buildpack := range group.Group {
 			remoteBuildpack, err := fetcher.ResolveAndFetch(ctx, buildpack)
 			if err != nil {
-				return nil, buildapi.BuilderRecord{}, err
+				return buildapi.BuilderRecord{}, err
 			}
 
 			buildpacks = append(buildpacks, remoteBuildpack.Optional(buildpack.Optional))
@@ -68,17 +67,17 @@ func (r *RemoteBuilderCreator) CreateBuilder(
 
 	writeableImage, err := builderBldr.WriteableImage()
 	if err != nil {
-		return nil, buildapi.BuilderRecord{}, err
+		return buildapi.BuilderRecord{}, err
 	}
 
 	identifier, err := r.RegistryClient.Save(builderKeychain, spec.Tag, writeableImage)
 	if err != nil {
-		return nil, buildapi.BuilderRecord{}, err
+		return buildapi.BuilderRecord{}, err
 	}
 
 	config, err := writeableImage.ConfigFile()
 	if err != nil {
-		return nil, buildapi.BuilderRecord{}, err
+		return buildapi.BuilderRecord{}, err
 	}
 
 	builder := buildapi.BuilderRecord{
@@ -94,7 +93,7 @@ func (r *RemoteBuilderCreator) CreateBuilder(
 		OS:                      config.OS,
 	}
 
-	return fetcher.UsedObjects(), builder, nil
+	return builder, nil
 }
 
 func buildpackMetadata(buildpacks []DescriptiveBuildpackInfo) corev1alpha1.BuildpackMetadataList {
--- a/pkg/reconciler/builder/builder.go
+++ b/pkg/reconciler/builder/builder.go
@@ -4,7 +4,6 @@ import (
 	"context"
 
 	"go.uber.org/zap"
-	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -34,7 +33,7 @@ const (
 )
 
 type BuilderCreator interface {
-	CreateBuilder(ctx context.Context, keychain authn.Keychain, fetcher cnb.RemoteBuildpackFetcher, clusterStack *buildapi.ClusterStack, spec buildapi.BuilderSpec) ([]corev1.ObjectReference, buildapi.BuilderRecord, error)
+	CreateBuilder(ctx context.Context, keychain authn.Keychain, fetcher cnb.RemoteBuildpackFetcher, clusterStack *buildapi.ClusterStack, spec buildapi.BuilderSpec) (buildapi.BuilderRecord, error)
 }
 
 func NewController(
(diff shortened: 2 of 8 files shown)


chenbh:
Creating as draft to get some early feedback, there's currently performance problems with the `cnb.Fetcher` where it doesn't cache keychains and has to make expensive api calls to get k8s secrets for every layer

image

chenbh:
> there's currently performance problems with the cnb.Fetcher where it doesn't cache keychains and has to make expensive api calls to get k8s secrets for every layer

Fixed in 592517b4fede7e8be6a406a186aa40b49949e81a

chenbh:
Not quite sure how I feel about passing around `[]*v1alpha2.Buildpack` and `[]*v1alpha2.ClusterBuildpack` around everywhere. Created #1140 as an alternate implementation and looking for feedback here

pkg/apis/build/v1alpha2/buildpack_types.go
```diff
@@ -0,0 +1,60 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   BuildpackSpec   `json:"spec"`
+	Status BuildpackStatus `json:"status"`
+}
+
+// +k8s:openapi-gen=true
+type BuildpackSpec struct {
+	// +listType
+	Source             corev1alpha1.StoreImage `json:"source,omitempty"`
```
matthewmcnew: Does this lead to double nesting the image reference?
```
apiVersion: kpack.io/v1alpha3
kind: Buildpack
metadata:
  name: my-custom-app-buildpack-0.3.1
  namespace: my-kpack-using-namespace
spec:
  source:
      image: gcr.io/my-org/my-custom-app-buildpack:0.3.1
  serviceAccountName: my-service-account
```

pkg/cnb/create_builder.go
```diff
@@ -1,53 +1,59 @@
...
+	Save(keychain authn.Keychain, tag string, image ggcrv1.Image) (string, error)
 }
 
 type LifecycleProvider interface {
-	LayerForOS(os string) (v1.Layer, LifecycleMetadata, error)
+	LayerForOS(os string) (ggcrv1.Layer, LifecycleMetadata, error)
 }
 
-type NewBuildpackRepository func(clusterStore *buildapi.ClusterStore) BuildpackRepository
+type BuilderCreator interface {
```
matthewmcnew, line 23: Is this interface necessary?

docs/buildpacks.md
```diff
@@ -0,0 +1,94 @@
...
+  sources:
+  - image: gcr.io/cf-build-service-public/node-engine-buildpackage@sha256:95ff756f0ef0e026440a8523f4bab02fd8b45dc1a8a3a7ba063cefdba5cb9493
+  - image: gcr.io/cf-build-service-public/npm-buildpackage@sha256:5058ceb9a562ec647ea5a41008b0d11e32a56e13e8c9ec20c4db63d220373e33
+  - image: gcr.io/paketo-buildpacks/builder:base
+```
+
+* `serviceAccountRef`: An object reference to a service account in any namespace. The object reference must contain `name` and `namespace`.
+* `sources`:  List of buildpackage images to make available in the ClusterStore. Each image is an object with the key image.
+
+> Note: ClusterBuilders will also work with a prebuilt builder image if a builpack is not available in a buildpackage.
```
matthewmcnew, line 77: Is this true for Builder resources as well?

pkg/reconciler/builder/builder.go
```diff
@@ -176,7 +185,30 @@ func (c *Reconciler) reconcileBuilder(ctx context.Context, builder *buildapi.Bui
...
-	return c.BuilderCreator.CreateBuilder(keychain, clusterStore, clusterStack, builder.Spec.BuilderSpec)
+	fetcher := cnb.NewRemoteBuildpackFetcher(c.KeychainFactory, clusterStore, buildpacks, clusterBuildpacks)
+
+	buildpackSources, buildRecord, err := c.BuilderCreator.CreateBuilder(ctx, keychain, fetcher, clusterStack, builder.Spec.BuilderSpec)
+	if err != nil {
+		return buildapi.BuilderRecord{}, err
+	}
+
+	for _, source := range buildpackSources {
+		err = c.Tracker.Track(reconciler.Key{
```
matthewmcnew, line 196: Is skipping tracking unused resources a problem if this leads to them not trigger updates to builders if they would be used by the builders?

pkg/apis/build/v1alpha2/buildpack_types.go
chenbh: Yep, I tired to keep it similar to ClusterStore's
```yaml
spec:
  sources:
  - image: gcr.io/cf-build-service-public/node-engine-buildpackage@sha256:95ff756f0ef0e026440a8523f4bab02fd8b45dc1a8a3a7ba063cefdba5cb9493
```

And ClusterStack's
```yaml
spec:
  buildImage:
    image: "paketobuildpacks/build:base-cnb"
  runImage:
    image: "paketobuildpacks/run:base-cnb"
```

That said, I'm not attached to it in anyway

pkg/cnb/create_builder.go
chenbh: It's used in the builder reconcilers to inject a fake during tests
https://github.com/pivotal/kpack/blob/2bf41f99a971be7b171032c016b369a7de6ea27c/pkg/reconciler/builder/builder.go#L96-L99
https://github.com/pivotal/kpack/blob/2bf41f99a971be7b171032c016b369a7de6ea27c/pkg/reconciler/builder/builder_test.go#L44-L57

pkg/reconciler/builder/builder.go
chenbh: I'm not quite sure I'm following what you mean by "if they would be used by the builders". Are you saying if a new version of a meta-buildpack references a previously untracked buildpack?

chenbh: Something like:

1. We have 3 Buildpack CRs, `paketo-buildpacks/go`, `paketo-buildpacks/go-dist`, and `paketo-buildpacks/go-vendor`
1. `paketo-buildpacks/go:1.0.0` is a meta-buildpack and only uses `go-dist`
1. `paketo-buildpacks/go:2.0.0` is a meta-buildpack and uses both `go-dist` and `go-vendor`
1. We create a Builder using `paketo-buildpacks/go:1.0.0`
1. The reconciler starts tracking `go` and `go-dist`
1. We bump the builder to `paketo-buildpacks/go:2.0.0`
1. It'll reconcile the entire builder again and figure out it needs to track all 3 Buildpacks now

docs/buildpacks.md
chenbh: This was originally copied over from `store.md`, but it did lead me to a fascinating and terrifying conclusion:

I tried this out and it's both valid and works

```yaml
apiVersion: kpack.io/v1alpha2
kind: Buildpack
metadata:
  name: my-buildpack
spec:
  source:
    image: index.docker.io/paketobuildpacks/builder-jammy-tiny
```

chenbh: Updated blurb in 51a0f1752390b3351921fbd0507fee9cd7612a8b

pkg/reconciler/builder/builder.go
chenbh: Talked offline, we need to track all Builpack resource changes because if a Builder uses only the `id` field, it's possible for a new Buildpack CR to be created with a newer version

chenbh: fixed in b7e42f1efd4329ee7794a88cd310ba13f0184b0d

pkg/tracker/tracker.go
```diff
@@ -53,28 +60,38 @@ type Tracker struct {
...
+	l, ok := i.objects[kind.String()]
+	if !ok {
+		l = set{}
+	}
+	// Overwrite the key with a new expiration.
+	l[obj] = time.Now().Add(i.leaseDuration)
+
+	i.kinds[kind.String()] = l
+}
+
```
tomkennedy513, line 94: This is awesome

docs/local.md
```diff
@@ -1,12 +1,67 @@
...
+#### When using public registries
+
 Access to a Kubernetes cluster is needed in order to install the kpack controllers.
 
 ```bash
 kubectl cluster-info # ensure you have access to a cluster
 ./hack/apply.sh <IMAGE/NAME> # <IMAGE/NAME> is a writable and publicly accessible location 
 ```
 
+#### When using private registries
```
tomkennedy513, line 12: nice!

pkg/cnb/create_builder.go
```diff
@@ -90,9 +90,11 @@ func (r *RemoteBuilderCreator) CreateBuilder(keychain authn.Keychain, clusterSto
 		Buildpacks:              buildpackMetadata(builderBldr.buildpacks()),
 		Order:                   builderBldr.order,
 		ObservedStackGeneration: clusterStack.Status.ObservedGeneration,
-		ObservedStoreGeneration: clusterStore.Status.ObservedGeneration,
+		ObservedStoreGeneration: fetcher.ClusterStoreObservedGeneration(),
 		OS:                      config.OS,
-	}, nil
+	}
+
+	return fetcher.UsedObjects(), builder, nil
```
tomkennedy513, line 97: Do we need to return the `fetcher.UsedObjects()` if they are not used in the builder or clusterbuilder reconcilers anymore?

chenbh: Thought about it some more, and the only information we would need for exit-early is already present in the builder spec, so we can get rid of `UsedObjects()` entirely

tomkennedy513 (approved):
This is a really incredible PR @chenbh

ncarlson (approved):

ncarlson:
> This is a really incredible PR @chenbh

agreed!

PR #1144, opened 2023-02-14
Bump kpack controller memory requests and limits

tomkennedy513:
- This will allow kpack to handle larger amounts of images and other resources
- Also remove the cpu limit because it is a compressible resource

### DIFF
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -141,7 +141,6 @@ spec:
         resources:
           requests:
             cpu: 20m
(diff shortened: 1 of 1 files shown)


chenbh (approved):
I can't believe we're only requesting `20m` of CPU

tomkennedy513:
> I can't believe we're only requesting `20m` of CPU, doesn't k8s default to `100m` if unspecified?

we're very efficient

PR #1157, opened 2023-02-27
Generate docker config in temp dir during image signing 

tomkennedy513:


chenbh:
If i'm understanding this correctly, this is to deal with platforms *cough* openshift *cough* where security policies are in place that forces every pod to run under a random uid/gid. In such platforms, we can't `chmod` the home directory and we can't explicitly configure the running user, so any attempts to write the `.dockerconfig.json` will fail.

So the solution here is to generate a tempdir and set `$DOCKER_CONFIG` env var (which supported in [ggcr](https://github.com/google/go-containerregistry/blob/main/pkg/authn/keychain.go#L80))

chenbh (approved):

PR #1181, opened 2023-04-10
Make our build and release process open source

tomkennedy513:


tomkennedy513:
the `--report-output-dir` flag wasn't working for windows so I am just pulling the digest out from the logs for the windows images, but once that issue is resolved we can remove that entire windows specific build job and windows/linux can share a build action

PR #1197, opened 2023-04-25
Bump cosign to v2 and use new APIs

ThomasVitale:
* Bump Cosign to 2.0.2
* Update cosign usage to use new Sign API
* Update cosign tests to use new Verify API

Fixes gh-1182


ThomasVitale:
@tomkennedy513 I'm looking into the failing tests and will update the PR asap. Thanks!

ThomasVitale:
@tomkennedy513 tests updated

pkg/cosign/image_signer_test.go
```diff
@@ -636,12 +556,12 @@ func pushRandomImage(t *testing.T, imageRef string) func() {
...
 	assert.Nil(t, err)
 
-	remoteImage, err := remote.Get(ref, regClientOpts...)
+	_, err = remote.Get(ref, regClientOpts...)
 	assert.Nil(t, err)
 
 	cleanup := func() {
 		_ = remote.Delete(ref, regClientOpts...)
-		ref, _ := ociremote.SignatureTag(remoteImage.Ref, ociremote.WithRemoteOptions(regClientOpts...))
+		ref, _ := ociremote.SignatureTag(ref, ociremote.WithRemoteOptions(regClientOpts...))
```
ThomasVitale, line 564: The new version of `github.com/google/go-containerregistry` changed the result type of the `remote.Get()` function on line 559 so that the `Ref` field is not exposed anymore. Since it would be the same value as the one provided as input, I changed this line to use the `ref` variable directly. What do you think?

tomkennedy513: I think that is fine especially for the test

tomkennedy513 (approved):
Thank you for this change!

Issue #1091, opened 2022-12-09
Image push to ECR registry is failing with an error related to write access verification


pranavek:
We have a Kpack setup in which hundreds of Kpack Images would get built at a time. While many of the images are getting pushed properly to the registry, some of them are failing with following write access error. We believe that network partitioning causes this issue. Is there a way to increase the timeout or set up a retry mechanism to reduce occurrences of this issue. 
 
<img width="1542" alt="AWS_ECR_Error" src="https://user-images.githubusercontent.com/6043543/206689565-46b9cdb7-fda1-431e-b21d-46e29ef04296.png">
Relevant code segment: https://github.com/google/go-containerregistry/blob/v0.11.0/pkg/v1/remote/check.go#L27,#L29

mvalliath:
We think the best path forward here would be to set up a retry mechanism for an entire build.

PR #1208, opened 2023-05-12
Add support for signing Builders and ClusterBuilders

### LINKED ISSUE #942: Support signing for images resulting from Builders/ClusterBuilders
The [cosign integration RFC](https://github.com/pivotal/kpack/blob/main/rfcs/0007-cosign-integration.md) specifies generation of signatures for images generated by kpack's `Image`, `Build`, `Builder`, and `ClusterBuilder` resources. However, kpack does not currently sign images resulting from `Builder` and `ClusterBuilder` resources.

stormqueen1990:
## Changes
* Implement Builder and ClusterBuilder signing with cosign, following the specification defined in [RFC 0007 - Integrate cosign with `kpack`](https://github.com/pivotal/kpack/blob/26d32c1a4e3995559c177f1deb5890f0a80c97e8/rfcs/0007-cosign-integration.md).

Fixes #942

### DIFF
--- a/cmd/completion/main.go
+++ b/cmd/completion/main.go
@@ -9,6 +9,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/sigstore/cosign/v2/pkg/oci/remote"
+
 	"github.com/BurntSushi/toml"
 	"github.com/buildpacks/lifecycle/platform/files"
 	"github.com/google/go-containerregistry/pkg/authn"
@@ -154,7 +156,7 @@ func main() {
 
 func signImage(report files.Report, keychain authn.Keychain) error {
 	if hasCosign() {
-		cosignSigner := cosign.NewImageSigner(logger, sign.SignCmd)
+		cosignSigner := cosign.NewImageSigner(sign.SignCmd, remote.SignatureTag)
 
 		annotations, err := mapKeyValueArgs(cosignAnnotations)
 		if err != nil {
--- a/pkg/apis/build/v1alpha2/builder_types.go
+++ b/pkg/apis/build/v1alpha2/builder_types.go
@@ -56,6 +56,12 @@ type NamespacedBuilderSpec struct {
 }
 
 // +k8s:openapi-gen=true
+type CosignSignature struct {
+	SigningSecret string `json:"signingSecret"`
+	TargetDigest  string `json:"targetDigest"`
+}
+
+// +k8s:openapi-gen=true
 type BuilderStatus struct {
 	corev1alpha1.Status     `json:",inline"`
 	BuilderMetadata         corev1alpha1.BuildpackMetadataList `json:"builderMetadata,omitempty"`
@@ -65,6 +71,7 @@ type BuilderStatus struct {
 	ObservedStackGeneration int64                              `json:"observedStackGeneration,omitempty"`
 	ObservedStoreGeneration int64                              `json:"observedStoreGeneration,omitempty"`
 	OS                      string                             `json:"os,omitempty"`
+	SignaturePaths          []CosignSignature                  `json:"signaturePaths,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/cnb/create_builder_test.go
+++ b/pkg/cnb/create_builder_test.go
@@ -177,6 +177,12 @@ func testCreateBuilderOs(os string, t *testing.T, when spec.G, it spec.S) {
 			KpackVersion:      "v1.2.3 (git sha: abcdefg123456)",
 			KeychainFactory:   keychainFactory,
 			LifecycleProvider: lifecycleProvider,
+			ImageSigner: &fakeBuilderSigner{
+				signBuilder: func(ctx context.Context, s string, secrets []*corev1.Secret, keychain authn.Keychain) ([]buildapi.CosignSignature, error) {
+					// no-op
+					return nil, nil
+				},
+			},
 		}
 
 		addBuildpack = func(t *testing.T, id, version, homepage, api string, stacks []corev1alpha1.BuildpackStack) {
@@ -346,7 +352,7 @@ func testCreateBuilderOs(os string, t *testing.T, when spec.G, it spec.S) {
 		})
 
 		it("creates a custom builder", func() {
-			builderRecord, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec)
+			builderRecord, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec, []*corev1.Secret{})
 			require.NoError(t, err)
 
 			assert.Len(t, builderRecord.Buildpacks, 4)
@@ -636,11 +642,12 @@ func testCreateBuilderOs(os string, t *testing.T, when spec.G, it spec.S) {
 		})
 
 		it("creates images deterministically ", func() {
-			original, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec)
+			original, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec, []*corev1.Secret{})
 			require.NoError(t, err)
 
 			for i := 1; i <= 50; i++ {
-				other, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec)
+				other, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec, []*corev1.Secret{})
+
 				require.NoError(t, err)
 
 				require.Equal(t, original.Image, other.Image)
@@ -670,7 +677,7 @@ func testCreateBuilderOs(os string, t *testing.T, when spec.G, it spec.S) {
 					},
 				}
 
-				_, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec)
+				_, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec, []*corev1.Secret{})
 				require.EqualError(t, err, "validating buildpack io.buildpack.unsupported.stack@v4: stack io.buildpacks.stacks.some-stack is not supported")
 			})
 
@@ -694,7 +701,7 @@ func testCreateBuilderOs(os string, t *testing.T, when spec.G, it spec.S) {
 					}},
 				}}
 
-				_, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec)
+				_, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec, []*corev1.Secret{})
 				require.EqualError(t, err, "validating buildpack io.buildpack.unsupported.mixin@v4: stack missing mixin(s): something-missing-mixin, something-missing-mixin2")
 			})
 
@@ -739,7 +746,7 @@ func testCreateBuilderOs(os string, t *testing.T, when spec.G, it spec.S) {
 					}},
 				}}
 
-				_, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec)
+				_, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec, []*corev1.Secret{})
 				require.Nil(t, err)
 			})
 
@@ -764,7 +771,7 @@ func testCreateBuilderOs(os string, t *testing.T, when spec.G, it spec.S) {
 					}},
 				}}
 
-				_, err := subject.CreateBuilder(ctx, builderKeychain, nil, fetcher, stack, clusterBuilderSpec)
+				_, err := subject.CreateBuilder(ctx, builderKeychain, nil, fetcher, stack, clusterBuilderSpec, []*corev1.Secret{})
 				require.Error(t, err, "validating buildpack io.buildpack.relaxed.old.mixin@v4: stack missing mixin(s): build:common-mixin, run:common-mixin, another-common-mixin")
 			})
 
@@ -787,7 +794,7 @@ func testCreateBuilderOs(os string, t *testing.T, when spec.G, it spec.S) {
 					}},
 				}}
 
-				_, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec)
+				_, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec, []*corev1.Secret{})
 				require.EqualError(t, err, "validating buildpack io.buildpack.unsupported.buildpack.api@v4: unsupported buildpack api: 0.1, expecting: 0.2, 0.3")
 			})
 
@@ -830,7 +837,7 @@ func testCreateBuilderOs(os string, t *testing.T, when spec.G, it spec.S) {
 					}},
 				}}
 
-				_, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec)
+				_, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec, []*corev1.Secret{})
 				require.NoError(t, err)
 			})
 		})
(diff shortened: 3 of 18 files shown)


pkg/cosign/image_signer.go
```diff
@@ -111,6 +134,139 @@ func (s *ImageSigner) sign(ro *options.RootOptions, refImage, secretLocation, co
...
+			if err != nil {
+				return nil, fmt.Errorf("failed to cleanup environment variable %s: %s", cosignRepositoryEnv, err)
+			}
+		}
+	}
+
+	return signaturePaths, nil
+}
+
+func FindCosignSecrets(ctx context.Context, client k8sclient.Interface, namespace string, serviceAccountName string) ([]corev1.Secret, error) {
```
tomkennedy513, line 243: I think it might be a little confusing that we have 2 methods with the same name that are used in different places and do different things

pkg/cosign/image_signer.go
```diff
@@ -1,34 +1,57 @@
...
-) error
+type SignFunc func(*options.RootOptions, options.KeyOpts, options.SignOptions, []string) error
+
+type FetchSignatureFunc func(name.Reference, ...cosignremote.Option) (name.Tag, error)
+
+type BuilderSigner interface {
+	SignBuilder(context.Context, string, []corev1.Secret, authn.Keychain) ([]v1alpha2.SignaturePath, error)
+}
 
 type ImageSigner struct {
```
tomkennedy513, line 33: Mayeb we could inject the k8s client here so that the (cluster)builder controller does not need to pass the k8s client through to the image signer

tomkennedy513: To elaborate a little more, I am thinking that the FindCosignSecret method could be called by the RemoteBuilderCreater and the logic would all be contained there so that the builder reconcilers don't need to care about cosign. Just a thought and I don't really have a strong opinion on this

tomkennedy513: I guess in that case you would have to pass in the service account to the remote builder create which probably does not make sense since the RemoteBuilderCreator doesn't deal with k8s resources like the reconcilers do

pkg/cosign/image_signer.go
stormqueen1990: Hi there, @tomkennedy513! I've renamed it to `FetchBuilderSigningSecrets` instead.

pkg/cosign/image_signer.go
```diff
@@ -1,34 +1,57 @@
...
-	Logger   *log.Logger
-	signFunc SignFunc
+	Logger             *log.Logger
+	signFunc           SignFunc
+	fetchSignatureFunc FetchSignatureFunc
 }
```
chenbh, line 37: The `Logger` field doesn't seem to be used anywhere, can we get rid of it entirely?

pkg/apis/build/v1alpha2/builder_types.go
```diff
@@ -65,6 +71,8 @@ type BuilderStatus struct {
 	ObservedStackGeneration int64                              `json:"observedStackGeneration,omitempty"`
 	ObservedStoreGeneration int64                              `json:"observedStoreGeneration,omitempty"`
 	OS                      string                             `json:"os,omitempty"`
+	Signed                  bool                               `json:"signed"`
+	SignaturePaths          []SignaturePath                    `json:"signaturePaths,omitempty"`
```
chenbh, line 75: I'm curious to the point of this bool, wouldn't the existence/non-emptiness of the `signaturePaths` field indicated if it's signed?

pkg/apis/build/v1alpha2/builder_types.go
```diff
@@ -55,6 +55,12 @@ type NamespacedBuilderSpec struct {
 	BackwardsCompatibleServiceAccount string `json:"serviceAccount,omitempty"`
 }
 
+// +k8s:openapi-gen=true
+type SignaturePath struct {
+	KeyName string `json:"keyName"`
+	Path    string `json:"path"`
+}
```
chenbh, line 62: If I'm understanding this correctly, a builder can be signed by multiple (cosign) secrets/keys. The purpose of this struct is to expose which keys were used and their corresponding signature (container image) locations in the status of the builder.

If so, maybe a better name for this would be
```suggestion
type CosignSignature struct {
	SigningSecret string `json:"signingSecret"`
	TargetImage    string `json:"targetImage"`
}
```

pkg/cosign/image_signer.go
```diff
@@ -111,6 +134,139 @@ func (s *ImageSigner) sign(ro *options.RootOptions, refImage, secretLocation, co
...
 
+func (s *ImageSigner) SignBuilder(
+	ctx context.Context,
+	imageReference string,
+	cosignSecrets []corev1.Secret,
+	builderKeychain authn.Keychain,
+) ([]v1alpha2.SignaturePath, error) {
+	const Timeout time.Duration = time.Second * 30
+
+	signaturePaths := make([]v1alpha2.SignaturePath, 0, 10)
```
chenbh, line 145: Any particular reason to use an initial capacity of 10?

chenbh (commented):
What do we want to happen if we add a cosign secret _after_ the [Cluster]Builder has been created?

Right now nothing happens, but we could get the builder and clusterbuilder reconcilers to track secrets? It does feel a little heavy hand to continuously read all new secrets in a namespace, but would that provide a better user experience?

(I don't have an opinion either way, just want to pick your brains a bit and maybe explicitly document this)

pkg/cosign/image_signer.go
```diff
@@ -111,6 +134,139 @@ func (s *ImageSigner) sign(ro *options.RootOptions, refImage, secretLocation, co
...
+	}
+
+	cosignSecrets := make([]corev1.Secret, 0)
+
+	for _, secretName := range serviceAccount.Secrets {
+		secret, err := client.CoreV1().Secrets(namespace).Get(ctx, secretName.Name, metav1.GetOptions{})
+		if err != nil {
+			// failed to retrieve one secret, fails signing
+			return nil, err
+		}
```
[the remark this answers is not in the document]
stormqueen1990: Part of the RFC was to fail building if kpack failed to sign and when writing it, I assumed missing secrets would be a failure scenario (although I did not specifically call that out in the RFC). I don't think it's extremely necessary to fail on missing secrets but failing silently might be confusing.

pkg/apis/build/v1alpha2/builder_types.go
stormqueen1990: Originally I thought about having it to indicate whether it has been signed or not but I agree that the `signaturePaths` field is enough. Will remove the `signed` field.

pkg/apis/build/v1alpha2/builder_types.go
stormqueen1990: Yes, that's correct. I'll rename it.

pkg/cosign/image_signer.go
stormqueen1990: No specific reason, will remove this too.

pkg/cosign/test_util.go
```diff
@@ -0,0 +1,57 @@
+package cosign
```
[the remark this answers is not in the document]
stormqueen1990: Update: moving the file to the `pkg/cosign/testing` folder creates an import cycle. Therefore, I'm renaming the file to `helpers_test.go` as you suggested.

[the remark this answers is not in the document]
stormqueen1990: Update 2: I had to externalize the constants into a new file `pkg/cosign/util/constants.go` to prevent a test import cycle while also allowing the cosign end-to-end tests to leverage that. Let me know if this doesn't work.

pkg/apis/build/v1alpha2/builder_types.go
stormqueen1990: I've renamed the field `Path` to `TargetDigest` as it's going to be holding the signature digest and not the image digest itself (that information is in another field of the `BuilderStatus` struct).

cmd/completion/main.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"github.com/sigstore/cosign/v2/pkg/oci/remote"
```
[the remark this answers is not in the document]
stormqueen1990: `goimports`ed everything!

test/execute_build_test.go
```diff
@@ -469,6 +472,618 @@ func testCreateImage(t *testing.T, when spec.G, it spec.S) {
 	})
 }
 
+func testSignBuilder(t *testing.T, _ spec.G, it spec.S) {
```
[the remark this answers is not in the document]
stormqueen1990: I've moved it to `cosign_e2e_test.go` in the same package, let me know your thoughts!

pkg/cosign/image_signer.go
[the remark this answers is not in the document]
stormqueen1990: I've updated this function to use `Fetcher.SecretsForServiceAccount`.

pkg/cosign/image_signer.go
stormqueen1990: I think we can. It doesn't look like it was being used in the previously merged code either.

pkg/apis/build/v1alpha2/builder_types.go
stormqueen1990: Removed `Signed` as requested.

pkg/apis/build/v1alpha2/builder_types.go
stormqueen1990: Renamed.

pkg/cosign/image_signer.go
stormqueen1990: Removed.

pkg/cosign/image_signer.go
stormqueen1990: Perhaps it was added for code parity with the `ImageSigner` in notary?
https://github.com/pivotal/kpack/blob/b323164d7c3f92c15b67b92ca0b1549383765369/pkg/notary/image_signer.go#L37-L41

stormqueen1990:
> What do we want to happen if we add a cosign secret _after_ the [Cluster]Builder has been created?
> 
> Right now nothing happens, but we could get the builder and clusterbuilder reconcilers to track secrets? It does feel a little heavy hand to continuously read all new secrets in a namespace, but would that provide a better user experience?
> 
> (I don't have an opinion either way, just want to pick your brains a bit and maybe explicitly document this)

For UX purposes, it sounds to me like that would be the best way to go forward. I'll wait for other opinions before making changes though.

pkg/cosign/image_signer.go
stormqueen1990: Removed the logger.

pkg/cnb/create_builder_test.go
```diff
@@ -857,13 +864,71 @@ func testCreateBuilderOs(os string, t *testing.T, when spec.G, it spec.S) {
...
 				require.EqualError(t, err, "unsupported platform apis in kpack lifecycle: 0.1, 0.2, 0.999, expecting one of: 0.3, 0.4, 0.5, 0.6, 0.7, 0.8")
 			})
 		})
+
+		when("signing a builder image", func() {
+			it("does not return the signed flag set when no secrets were present", func() {
+				builderRecord, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec, []corev1.Secret{})
+				require.NoError(t, err)
+				require.NotNil(t, builderRecord)
+			})
```
chenbh, line 877: This should also assert on `builderRecord.SignaturePaths` being empty

pkg/cnb/create_builder_test.go
```diff
@@ -857,13 +864,71 @@ func testCreateBuilderOs(os string, t *testing.T, when spec.G, it spec.S) {
...
+								SigningSecret: fmt.Sprintf("k8s://%s/%s", fakeSecret.Namespace, fakeSecret.Name),
+								TargetDigest:  "registry.local/test-image:signature-tag",
+							},
+						}, nil
+					},
+				}
+
+				builderRecord, err := subject.CreateBuilder(ctx, builderKeychain, stackKeychain, fetcher, stack, clusterBuilderSpec, []corev1.Secret{fakeSecret})
+				require.NoError(t, err)
+				require.NotNil(t, builderRecord)
```
chenbh, line 918: This should also assert on `builderRecord.SignaturePaths` being non-empty

test/execute_build_test.go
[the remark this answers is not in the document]
chenbh: I like it, it's annoying that we have to redefine all the `it.Before` and `it.After` again, but maybe we can look into a way of refactoring that in a future PR

pkg/cnb/create_builder_test.go
stormqueen1990: Updated!

pkg/cnb/create_builder_test.go
stormqueen1990: Updated!

test/execute_build_test.go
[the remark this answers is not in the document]
stormqueen1990: On that note, perhaps when this refactoring happens tests could be split using Go build constraint tags?
I've seen that pattern before where e2e tests are tagged as such using `//go:build e2e` (or something to that effect). That would allow unit tests to be run using `go test -v ./...`, as well as running e2e tests with something like `go test -v -tags=e2e ./...`.

pkg/reconciler/builder/builder.go
```diff
@@ -56,6 +61,7 @@ func NewController(
 		BuildpackLister:        buildpackInformer.Lister(),
 		ClusterBuildpackLister: clusterBuildpackInformer.Lister(),
 		ClusterStackLister:     clusterStackInformer.Lister(),
+		K8sClient:              k8sClient,
```
chenbh, line 64: Can we re-org this so that there's a nice separation of concerns where the access level is incrementally stripped down as it passes through abstractions:

1. `cmd/controller/main.go` would construct the `secret.Fetcher{}` struct and pass it to the reconcilers.
    - Only `main` should have raw access to the clients, everything else should be passed in a stripped down interface
1. `pkg/reconciler/builder/builder.go` gets passed a pointer to the `Fetcher` which it then uses to find the secrets for a particular service account. It then passes the list of secrets to the cosign signer
    - Only the reconciler knows what CRs and objects it needs to do a reconcile
1. `pkg/cosign/image_signer.go` gets passed a list of secrets which it then uses to filter out all non-cosign related secrets
    - Only the cosign package knows what's the difference between a regular secret and a cosign secret 

It's the same reason why we pass in all of these `BuildpackLister`, `ClusterBuildpackLister`,  `ClusterStackLister`, etc, instead of just directly calling `c.Client.KpackV1Alpha2().ClusterStack().List()`

test/cosign_e2e_test.go
```diff
@@ -0,0 +1,626 @@
...
+				},
+				RunImage: buildapi.ClusterStackSpecImage{
+					Image: "gcr.io/paketo-buildpacks/run:base-cnb",
+				},
+			},
+		}, metav1.CreateOptions{})
+		require.NoError(t, err)
+	})
+
+	it("Signs a Builder image successfully when the key is not password-protected", func() {
```
chenbh, line 173: Can you add 2 tests for:

- the failure case, where an invalid key (or password) should mark the builder as failed
- multiple secrets case, where a couple different secrets should result in different signatures

chenbh:
> What do we want to happen if we add a cosign secret after the [Cluster]Builder has been created?

We're going to bring this up at the next kpack wg, but from early talks with @tomkennedy513  and @matthewmcnew, we have some ideas:

- Do nothing, the user will have to recreate the Builder to get it signed by every secret
- Rely on the natural resync period for Builders (once every 10~12 hrs) and diff the current secrets found vs secrets used in the last run to see if we need a rebuild/resign
- Watch all service accounts referenced by Builders and queue for reconcile if any service account changes
- Watch all secrets and queue for reconcile if any cosign secrets (i.e. those that have `cosign.blah` in the data fields)

pkg/cosign/image_signer.go
```diff
@@ -111,6 +125,134 @@ func (s *ImageSigner) sign(ro *options.RootOptions, refImage, secretLocation, co
 	return nil
 }
 
+func (s *ImageSigner) SignBuilder(
+	ctx context.Context,
+	imageReference string,
+	cosignSecrets []corev1.Secret,
+	builderKeychain authn.Keychain,
+) ([]v1alpha2.CosignSignature, error) {
+	const Timeout = time.Second * 30
```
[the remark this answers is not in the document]
stormqueen1990: This was based on a previous version of this code that used cosign v1 and was timeouting without it. I came up with the value in a very non-scientific manner too, but I guess I could just remove it. Not sure if it's still needed.

pkg/reconciler/builder/builder.go
stormqueen1990: Could you please elaborate on the pointer to fetcher point? I think I'm not following why it's important to have a pointer there.

pkg/cosign/image_signer.go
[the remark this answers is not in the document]
stormqueen1990: I stand corrected and a non-default timeout is indeed needed (I get a context deadline exceeded error when a timeout is not specified).
cosign e2e tests uses [3 minutes as a default timeout](https://github.com/sigstore/cosign/blob/6f4f477ee452e3335a2557f2952e51706bf9118f/test/e2e_test.go#L159) so perhaps we should use the same value here.

test/cosign_e2e_test.go
stormqueen1990: @chenbh about this, I wasn't able to find a specific failure condition that gets written in the Builder/ClusterBuilder. What's the preferred method to tell a definitive failure apart from a transitory state caused by the reconcile loop in the context of kpack?

pkg/reconciler/builder/builder.go
stormqueen1990: I've addressed this via 82a7474 and 529ce1f. Looking for some more specific feedback in the creation of the interface for mocking + fake for the `secret.Fetcher`.

test/cosign_e2e_test.go
stormqueen1990: I've decided to use the message for now, but perhaps an RFC for changing this behaviour would be warranted? It's hard to tell the failure modes apart right now.

pkg/cosign/image_signer.go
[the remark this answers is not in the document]
stormqueen1990: I changed this to use the [same default timeout as cosign](https://github.com/sigstore/cosign/blob/79d4184cef8326738166b430a0f08339840b2da6/cmd/cosign/cli/options/root.go#L39) for now.

test/cosign_e2e_test.go
chenbh: I think just the error message is fine for now. 

The transitive error vs definitive errors is definitely something we've run into in other parts of kpack especially in regards to network hiccups (#1091). But the whole discussion around how to detect and potentially recover from transient errors is a much larger one and should be a separate issue/pr/rfc

pkg/reconciler/builder/builder.go
chenbh: > Could you please elaborate on the pointer to fetcher point? I think I'm not following why it's important to have a pointer there.

The pointer thing was mainly so that the reconcilers can take an interface instead of a concrete type, which is exactly how you implemented it.

> Looking for some more specific feedback in the creation of the interface for mocking + fake for the `secret.Fetcher`.

I would avoid calling it `IFetcher` and move the interface declaration to the reconcilers. The golang convention is to either export either the interface or the struct, but not both. So if you export the struct, the interface would be defined in the file that uses it, and if you export the interface, the struct would be private.

The convention used in kpack is the former, where we have the Builder and ClusterBuilder reconcilers declare the same `BuilderCreator` interface, and `pkg/cnb/create_builder.go` exports the struct that has the signature.

https://github.com/buildpacks-community/kpack/blob/a8479f9508289d435d513f3681823a096958d357/pkg/reconciler/builder/builder.go#L35-L37
https://github.com/buildpacks-community/kpack/blob/a8479f9508289d435d513f3681823a096958d357/pkg/cnb/create_builder.go#L23-L30

chenbh:
> We're going to bring this up at the next kpack wg, but from early talks with @tomkennedy513 and @matthewmcnew, we have some ideas:

We ended up deciding that because there's no exit-early when reconciling builders, it's ok to rely on the natural (10-hour) resync period. So if the cosign secrets changed, the users will have to wait up to 10 hours before kpack pushes a new builder with the updated signature(s).

chenbh:
I think this PR is in a good place once https://github.com/buildpacks-community/kpack/pull/1208#discussion_r1265844678 is addressed, @stormqueen1990 do you mind cleaning/squashing up the commits a little? There are several of them titled "changes from code review" with wildly different changes in them that should really be separated into different commits or squashed into another commit.

Just to be clear, I don't care about the amount of commits as long as each individual commit is immediately obvious to what it's trying to do and is focused on one "unit of work"

stormqueen1990:
> I think this PR is in a good place once [#1208 (comment)](https://github.com/buildpacks-community/kpack/pull/1208#discussion_r1265844678) is addressed, @stormqueen1990 do you mind cleaning/squashing up the commits a little? There are several of them titled "changes from code review" with wildly different changes in them that should really be separated into different commits or squashed into another commit.
> 
> Just to be clear, I don't care about the amount of commits as long as each individual commit is immediately obvious to what it's trying to do and is focused on one "unit of work"

Hi there, @chenbh! 👋

Sure thing! I was planning to squash commits into more meaningful chunks once this is good to go. I am a bit busy this week but will try to make all changes (squashing included) by EOD Thursday.

stormqueen1990:
Hi there, @chenbh! Sorry for the delay, I started facing issues after rebasing the branch for the last time. I'll have to investigate it and will push the updated and squashed branch as soon as possible.

pkg/reconciler/builder/builder.go
stormqueen1990: I updated this bit, let me know if any further changes are needed!

chenbh (approved):
lgtm!

chenbh:
@tomkennedy513 @matthewmcnew @samj1912 do one of you mind taking a quick 👀 at this PR? It's a pretty beefy one and I would like at least one more review on it before merging

tomkennedy513 (approved):
Hey @stormqueen1990 would you be able to rebase this pr?

stormqueen1990:
> Hey @stormqueen1990 would you be able to rebase this pr?

Hi there, @tomkennedy513! Yes, just rebased it. Please let me know if there's anything else that's needed to move this PR forward 🙏

tomkennedy513:
Thank you so much for this work! @stormqueen1990

PR #1221, opened 2023-05-23
Improve kpack conditions

evankanderson:
Improve condition fields for kpack Images, particularly for use with [Cartographer health rules](https://cartographer.sh/docs/development/health-rules/#multi-match):

> When a matcher set is satisfied, the message field of the first condition will be replicated on the owner object.
(etc for single-condition matchers)

I also did a little cleanup, following the [Kubernetes API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties):

Conditions should always have 'reason' filled ("This field may not be empty").

Additionally, 'message' should be filled in for Unknown or Error conditions so that users know how to resolve the issue.  (Inspired by a question on how to represent "build in progress" status from the `tanzu apps` CLI.)

### DIFF
--- a/pkg/reconciler/image/reconcile_build.go
+++ b/pkg/reconciler/image/reconcile_build.go
@@ -60,7 +60,7 @@ func (c *Reconciler) reconcileBuild(ctx context.Context, image *buildapi.Image, 
 	case corev1.ConditionFalse:
 		return buildapi.ImageStatus{
 			Status: corev1alpha1.Status{
-				Conditions: noScheduledBuild(result, builder, latestBuild, sourceResolver),
+				Conditions: noScheduledBuild(result.ConditionStatus, builder, latestBuild, sourceResolver),
 			},
 			LatestBuildRef:             latestBuild.BuildRef(),
 			LatestBuildReason:          latestBuild.BuildReason(),
@@ -75,7 +75,7 @@ func (c *Reconciler) reconcileBuild(ctx context.Context, image *buildapi.Image, 
 	}
 }
 
-func noScheduledBuild(buildNeeded buildRequiredResult, builder buildapi.BuilderResource, build *buildapi.Build, sourceResolver *buildapi.SourceResolver) corev1alpha1.Conditions {
+func noScheduledBuild(buildNeeded corev1.ConditionStatus, builder buildapi.BuilderResource, build *buildapi.Build, sourceResolver *buildapi.SourceResolver) corev1alpha1.Conditions {
 	if !builder.Ready() {
 		return corev1alpha1.Conditions{
 			{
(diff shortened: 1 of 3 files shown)


tomkennedy513:
Thank you for the pr @evankanderson! We will take a look

tomkennedy513 (approved):

pkg/reconciler/image/reconcile_build.go
```diff
@@ -71,27 +76,49 @@ func (c *Reconciler) reconcileBuild(ctx context.Context, image *buildapi.Image,
 }
 
 func noScheduledBuild(buildNeeded corev1.ConditionStatus, builder buildapi.BuilderResource, build *buildapi.Build, sourceResolver *buildapi.SourceResolver) corev1alpha1.Conditions {
```
chenbh, line 78: This function reads a little bit awkward with all the ifs that immediately return. What do you think of converting it to a switch statement to make it more explicitly layout the possible causes:

```go
func noScheduledBuild(buildNeeded corev1.ConditionStatus, builder buildapi.BuilderResource, build *buildapi.Build, sourceResolver *buildapi.SourceResolver) corev1alpha1.Conditions {
	var (
		status  corev1.ConditionStatus
		reason  string
		message string
	)

	switch {
	case !builder.Ready():
		status = corev1.ConditionFalse
		reason = buildapi.BuilderNotReady
		message = builderError(builder)
	case buildNeeded == corev1.ConditionUnknown:
		status = corev1.ConditionUnknown
		reason = ResolverNotReadyReason
		if sourceResolver.Ready() {
			message = "Build status unknown"
		} else {
			message = fmt.Sprintf("SourceResolver %s is not ready", sourceResolver.GetName())
		}
	case !build.Status.GetCondition(corev1alpha1.ConditionSucceeded).IsTrue():
		status = unknownStatusIfNil(build.Status.GetCondition(corev1alpha1.ConditionSucceeded))
		reason = BuildFailedReason
		message = defaultMessageIfNil(build.Status.GetCondition(corev1alpha1.ConditionSucceeded), "Last build did not succeed")
	default:
		status = unknownStatusIfNil(build.Status.GetCondition(corev1alpha1.ConditionSucceeded))
		reason = UpToDateReason
		message = defaultMessageIfNil(build.Status.GetCondition(corev1alpha1.ConditionSucceeded), "Last build succeeded")
	}

	return corev1alpha1.Conditions{
		{
			Type:               corev1alpha1.ConditionReady,
			Status:             status,
			Reason:             reason,
			Message:            message,
			LastTransitionTime: corev1alpha1.VolatileTime{Inner: metav1.Now()},
		},
		builderCondition(builder),
	}
}

```

chenbh (approved):

PR #1223, opened 2023-05-24
Use keychain from clusterstack to pull build image in CreateBuild

tomkennedy513:
- fallback to builder keychain if there is no service account on the clusterstack to mimic existing behavior

### DIFF
--- a/pkg/reconciler/builder/builder.go
+++ b/pkg/reconciler/builder/builder.go
@@ -33,7 +33,7 @@ const (
 )
 
 type BuilderCreator interface {
-	CreateBuilder(ctx context.Context, keychain authn.Keychain, fetcher cnb.RemoteBuildpackFetcher, clusterStack *buildapi.ClusterStack, spec buildapi.BuilderSpec) (buildapi.BuilderRecord, error)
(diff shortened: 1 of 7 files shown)


pkg/reconciler/builder/builder.go
```diff
@@ -203,17 +203,28 @@ func (c *Reconciler) reconcileBuilder(ctx context.Context, builder *buildapi.Bui
...
-	keychain, err := c.KeychainFactory.KeychainForSecretRef(ctx, registry.SecretRef{
+	builderKeychain, err := c.KeychainFactory.KeychainForSecretRef(ctx, registry.SecretRef{
 		ServiceAccount: builder.Spec.ServiceAccount(),
 		Namespace:      builder.Namespace,
 	})
 	if err != nil {
 		return buildapi.BuilderRecord{}, err
 	}
 
+	stackKeychain := builderKeychain
```
chenbh, line 214: I'm a little hesitant on falling back to the stack keychain, this feels like compromising design for backwards compatibility. That being said, I'm not too married to this idea and certainly not enough to block this PR

chenbh (approved):

PR #1226, opened 2023-05-26
Followup on 1221 - further condition improvements

evankanderson:
I incorporated [Daniel's suggestion](https://github.com/pivotal/kpack/pull/1221#discussion_r1206015959), which caused me to also re-evaluate the failed build error message.  I incorporated the name of the build, which seemed like an improvement.


pkg/reconciler/image/reconcile_build.go
```diff
@@ -76,53 +77,36 @@ func (c *Reconciler) reconcileBuild(ctx context.Context, image *buildapi.Image,
...
+		ready.Reason = buildapi.BuilderNotReady
+		ready.Message = builderError(builder)
+	case buildNeeded == corev1.ConditionUnknown && !sourceResolver.Ready():
+		ready.Status = corev1.ConditionUnknown
+		ready.Reason = ResolverNotReadyReason
+		ready.Message = fmt.Sprintf("SourceResolver %s is not ready", sourceResolver.GetName())
+	case buildNeeded == corev1.ConditionUnknown && sourceResolver.Ready():
+		ready.Status = corev1.ConditionUnknown
+		ready.Reason = UnknownStateReason
+		ready.Message = "Build status unknown"
```
evankanderson, line 98: Excitingly, this code isn't covered, as I changed the reason and no tests failed.

(I've been carrying this state around, but I'm not sure what it's supposed to represent...)

evankanderson (commented):
`+34 -49` makes me happy, thanks for the suggestion!

pkg/reconciler/image/reconcile_build.go
chenbh: @matthewmcnew is this related to that libgit2 bug where the source resolver hangs forever? I can't think of any scenarios that would result in status unknown when both builder and source resolver are ready.

evankanderson: This was https://github.com/pivotal/kpack/blob/7857820a20481e34b338c345ee4640ada53bb195/pkg/reconciler/image/reconcile_build.go#L74-L88, which would have had a message of `""` and status `Unknown` with no `Reason`.

This is now explicit enough that at least it's obvious that we don't know how we got into this state.  😁

evankanderson: Looking at the source, it appears that _today_ the only way to get to this state is when SourceResolver or Builder are not ready.  Of course, a future refactor could change that, so I'd like to have some sort of default "not okay" status.

PR #1231, opened 2023-06-06
Resolve concurrent map write error in remote git resolver

tomkennedy513:
- Move logic that can cause concurrent map writes to an init function


pkg/git/fetch_test.go
```diff
@@ -108,15 +106,6 @@ func testGitCheckout(t *testing.T, when spec.G, it spec.S) {
...
 		})
-
-		it("uses the http proxy env vars", func() {
-			require.NoError(t, os.Setenv("HTTPS_PROXY", "http://invalid-proxy"))
-			defer os.Unsetenv("HTTPS_PROXY")
-
-			err := fetcher.Fetch(testDir, "https://github.com/git-fixtures/basic", "master", metadataDir)
-			require.Error(t, err)
-			require.Contains(t, err.Error(), "proxyconnect tcp: dial tcp")
-		})
```
chenbh, line 119: Were you able to verify that `HTTPS_PROXY` is supported out of the box? I still think it's valuable to test it, but I  can't get this test to work.

tomkennedy513: Yeah I ran a proxy locally and saw that it was attempting to reach git through the proxy. I think the problem with the tests is that they're running in parallel and attempting to use a shared environment variable. I personally don't see a ton of value on this test because we are basically testing the go http proxy package for this particular use case and not any of the other instances where kpack may need to utilize a proxy with the same default transport (i.e. registry access)

tomkennedy513: If you focus that test with the changes in the PR then you it will use the proxy.

chenbh: There might be a conversation for a integration test instead of unit, but as long as it was at least tested manually I'm fine with this

chenbh (approved):

PR #1232, opened 2023-06-06
Deduplicate builder layers by diff id

### LINKED ISSUE #645: Builder images contain duplicate layers
I noticed [this issue with pack](https://github.com/buildpacks/pack/issues/1081) where `pack builder create` produces builder images with multiple copies of an identical layer. This happens when a builder contains metabuildpacks that include some overlapping implementation buildpacks. (A common example is the Paketo Procfile buildpack, which is included in the Paketo Go, Dotnet Core, and Ruby buildpacks, to name a few.)

Out of curiosity, I took a look to see if kpack has the same problem. It seems to. Take a look at [this gist](https://gist.github.com/fg-j/72bc224a8770a5094678385d661cd610) for an analysis of the layers in a builder whose metabuildpacks contain overlapping buildpacks.

The layer duplication creates unnecessary builder size bloat. This can slow down image pulling. But it also poses a larger problem: builders grow rapidly with the addition of each metabuildpack. Since there is an underlying limit to the number of layers that a filesystem driver can package into an image ([for docker's backend this is about 128 layers](https://docs.docker.com/storage/storagedriver/overlayfs-driver/#how-the-overlay2-driver-works)), there will probably come a point where a builder gets too large and can't be built/pushed to the registry!

tomkennedy513:
- This will remove duplicate layers that result from two buildpacks in a builder that have the same layer diff id

resolves #645 

Co-authored-by Matthew McNew <mmcnew@vmware.com>

### DIFF
--- a/pkg/cnb/builder_builder.go
+++ b/pkg/cnb/builder_builder.go
@@ -126,7 +126,7 @@ func (bb *builderBlder) WriteableImage() (v1.Image, error) {
 	}
 
 	image, err := mutate.AppendLayers(bb.baseImage,
-		layers(
+		deduplicateLayers(layers(
 			[]v1.Layer{
 				defaultLayer,
 				bb.lifecycleLayer,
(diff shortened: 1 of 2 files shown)


chenbh (approved):
Small question, but otherwise lgtm

chenbh:
Related `pack` PR: https://github.com/buildpacks/pack/pull/1787

PR #1239, opened 2023-06-29
Add support for known_hosts file

### LINKED ISSUE #1230: GitHub private repositories fail due to `unable to find any valid known_hosts file, set SSH_KNOWN_HOSTS env variable`
Kpack 0.11.0 [replaced git2go with go-git](#1159) and doing so GitHub private repositories fail due to this error:
`unable to fetch references for repository: unable to find any valid known_hosts file, set SSH_KNOWN_HOSTS env variable`

The same configuration works with kpack-0.9.4 without issue. This appears to be a regression due git2go -> go-git and may be caused by the handling of known_hosts in the context of the underlying SSH connection to GitHub's git repositories. 

In kpack-0.9.4, private repositories on GitHub required a secret that included GitHub's SSH public key(s). For example KNOWNHOSTS64 below would be `ssh-keyscan github.com` encoded base64. This requirement was pieced together from aged documentation references including #413. It is possible that this issue is from some unrelated misconfiguration we have, so we'd like this to be confirmed prior to giving this issue any more attention!

```
apiVersion: v1
data:
  known_hosts: KNOWNHOSTS64
  ssh-privatekey: PRIVATEKEY64
kind: Secret
metadata:
  annotations:
    kpack.io/git: git@github.com
```  

This has been tested by trying identical Image configurations on kpack-0.9.4 (working) and kpack-0.11.0 (not working).

chenbh:
fixes #1230 

For more details see the docs change

- Introduce a flag `INSECURE_SSH_TRUST_UNKNOWN_HOSTS` to automatically trust unknown hosts (default value is `true` to keep backwards compatibilty
- If the flag is disabled, a `known_hosts` field can be specified on the ssh auth secret to configure it
- Add e2e tests for authenticated git sources (basic auth + ssh auth)

### DIFF
--- a/pkg/buildpod/generator.go
+++ b/pkg/buildpod/generator.go
@@ -11,7 +11,6 @@ import (
 	ggcrv1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
-	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/dynamic"
@@ -44,6 +43,7 @@ type Generator struct {
 	DynamicClient             dynamic.Interface
 	MaximumPlatformApiVersion *semver.Version
 	InjectedSidecarSupport    bool
+	SSHTrustUnknownHost       bool
 }
 
 type BuildPodable interface {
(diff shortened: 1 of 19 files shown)


pkg/buildpod/generator.go
```diff
@@ -11,7 +11,6 @@ import (
 	ggcrv1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
-	v1 "k8s.io/api/core/v1"
```
tomkennedy513, line 14: nice!

tomkennedy513 (approved):
Nice work!

PR #1253, opened 2023-07-11
Remove add-to-project workflow

tomkennedy513:
- Now that the kpack project is no longer owned by VMware, we should not be adding the kpack prs to our project as part of a workflow


chenbh (approved):
Neat that Github projects can add issues from across orgs now

PR #1254, opened 2023-07-11
upgrading some lifecycle api versions

joe-kimmel-vmw:


pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -1154,10 +1154,10 @@ func (b *Build) setupCosignVolumes(secrets []corev1.Secret) ([]corev1.Volume, []
 }
 
 var (
-	lowestSupportedPlatformVersion = semver.MustParse("0.3")
+	lowestSupportedPlatformVersion = semver.MustParse("0.7")
 
-	supportedPlatformAPIVersionsWithWindowsAndReportToml = []*semver.Version{semver.MustParse("0.9"), semver.MustParse("0.8"), semver.MustParse("0.7"), semver.MustParse("0.6"), semver.MustParse("0.5"), semver.MustParse("0.4")}
```
joe-kimmel-vmw, line 1159: go look for the bits and pieces that check platform API version

pkg/apis/build/v1alpha2/build_pod_test.go
```diff
@@ -903,34 +762,11 @@ func testBuildPod(t *testing.T, when spec.G, it spec.S) {
 			}
 			assert.Equal(t, append(append([]string{
 				"-layers=/layers",
-				"-analyzed=/layers/analyzed.toml",
-				"-run-image=builderregistry.io/run"},
```
joe-kimmel-vmw, line 907: I think this is correct that the run-image wouldn't be there anymore, because we've upgraded the platform API version,  and _think_ that really is the main change,  but uh... cc @natalieparellano  :)

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -1154,17 +1115,13 @@ func (b *Build) setupCosignVolumes(secrets []corev1.Secret) ([]corev1.Volume, []
 }
 
 var (
-	lowestSupportedPlatformVersion = semver.MustParse("0.3")
+	lowestSupportedPlatformVersion = semver.MustParse("0.7")
```
chenbh, line 1118: I wonder if this var should've really been named `platformAPIEquals03`, because it's only used to determine if the `--process-type` and `-report` flags are passed to the exporter. 

@joe-kimmel-vmw do you know if these flags are guaranteed to be supported >0.7? If so we can also clean up line 553 thru 569 and remove this var.

 https://github.com/buildpacks-community/kpack/blob/37a9406669b68f5285895501d4afbce0a3b27a7d/pkg/apis/build/v1alpha2/build_pod.go#L553-L569

chenbh (changes requested):

pkg/apis/build/v1alpha2/build_pod.go
joe-kimmel-vmw: cool,  good point.  yeah - maybe they're not guaranteed to be supported until the heat death of the universe, but for the foreseeable future they should be in there afaik.

joe-kimmel-vmw: btw this should be fine but i did have to change the unit tests just a little more.   `proces-type=web` is the default, so now we're just passing in the default explicitly rather than sometimes silently letting the default happen.  should be NBD but it's always possible there's some tripping hazard I'm not seeing :)

pkg/apis/build/v1alpha2/build_pod.go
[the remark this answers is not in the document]
joe-kimmel-vmw: Thanks-  agree I was being too conservative in my first pass.

pkg/apis/build/v1alpha2/build_pod_test.go
natalieparellano: We should still provide `-run-image` to the analyzer. It is only removed as a flag to the exporter on newer platforms.

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -274,24 +270,13 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
 		Args: args(
 			[]string{"-layers=/layers", "-analyzed=/layers/analyzed.toml"},
 			func() []string {
-				if !platformAPILessThan07 {
-					return []string{"-run-image=" + runImage}
-				}
 				return []string{}
 			}(),
```
chenbh, line 274: This func, along with L276-L278 can be removed. The `args` func just appends a bunch of string slices together, so if we return an empty slice, it's a noop either way.

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -315,12 +297,8 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
...
 			workspaceVolume,
 			homeMount,
-		}, func() []corev1.VolumeMount {
-			if platformAPILessThan07 {
-				return cacheVolumes
-			}
-			return []corev1.VolumeMount{}
-		}()),
+		},
+			[]corev1.VolumeMount{}),
```
chenbh, line 301: The `volumeMounts` func does the same thing as `args`, so passing in an empty slice is noop

```suggestion
		}),
```

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -588,27 +537,12 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
...
-
-							}(),
-							func() []string {
-								if platformAPILessThan07 {
-									return []string{"-run-image=" + runImage}
-								}
-								return []string{}
-							}(),
+							[]string{fmt.Sprintf("-report=%s", ReportTOMLPath)},
+							[]string{},
```
chenbh, line 545: Empty slice is noop
```suggestion
```

chenbh (changes requested):

pkg/apis/build/v1alpha2/build_pod.go
```diff
@@ -588,27 +537,12 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
 							exporterCacheArgs,
 							func() []string {
 								if b.DefaultProcess() == "" {
-									if platformAPI.Equal(lowestSupportedPlatformVersion) || platformAPI.GreaterThan(semver.MustParse("0.5")) {
-										return nil
-									} else {
-										return []string{fmt.Sprintf("-process-type=web")}
-									}
+									return []string{fmt.Sprintf("-process-type=web")}
```
natalieparellano, line 540:
```suggestion
									return nil
```

tomkennedy513:
superseded by #1305

Issue #1255, opened 2023-07-11
Build status should show why it failed when the build pod runs out of ephemeral disk space


davewalter:
#### Details

When the build pod errors due to a full ephemeral storage disk, the Build's Ready status condition is set to false without a useful reason or message that indicates why.

```yaml
status:
  conditions:
  - lastTransitionTime: "2023-07-11T23:37:02Z"
    status: "False"
    type: Succeeded
```

Inspecting the build pod's status, the `Ready` status condition simply indicates that the pod has failed:

```yaml
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2023-07-11T23:34:51Z"
    reason: PodFailed
    status: "False"
    type: Ready
```

The `status.initContainerStatuses[{name: "build"}].state` shows that the container has terminated with an exit code of `2` and a reason of `Error`:

```yaml
status:
  initContainerStatuses:
  - name: build
    ...
    ready: false
    state:
      terminated:
        containerID: ...
        exitCode: 2
        finishedAt: "2023-07-11T23:37:01Z"
        reason: Error
        startedAt: "2023-07-11T23:34:56Z"
```

The only useful status from the build pod is the `status.message` field, which states `"Pod ephemeral local storage usage exceeds the total limit of containers"`.

It would be helpful to be able to see the useful error message reflected in the `Build` (and `Image`) succeeded status conditions in the human-readable `message` field. It would also be nice to see the machine-readable `reason` field copied from the build pod's ready status condition into the succeeded conditions for both resources.

tomkennedy513:
Hey @davewalter, what version of kpack are you using?

davewalter:
Hi @tomkennedy513,

This testing was performed with [v0.10.0](https://github.com/pivotal/kpack/releases/tag/v0.10.0). I see that v0.11.0 includes changes #1221 and #1226 to improve the status conditions on the `Image` resource based on the value from the corresponding `Build` resource. From what I can tell, it doesn't look like the build reconciler's [`conditionForPod` helper function](https://github.com/buildpacks-community/kpack/blob/6a1e5b77fced7ca10955ef5350703efbe1e31c40/pkg/reconciler/build/build.go#L236) sets a reason/message in the Build's Succeeded condition (other than when the status is unknown).

I have taken a stab at updating the build reconciler to set suitable reasons/messages based on the status of the build pod. Please let me know if you have any questions or comments.

pviraj59:
Resolved by #1258

PR #1258, opened 2023-07-13
Improve Build resource conditions

davewalter:
Inspired by #1221, this PR updates the build reconciler to set status condition reasons and messages on Builds based on the status of the corresponding build pod.

See #1255 for the issue that this tries to address.


chenbh (approved):
sorry for the long delay, but lgtm!

PR #1268, opened 2023-07-20
[0.10.x] Backport Deduplicate builder layers by diff id

tomkennedy513:
- This will remove duplicate layers that result from two buildpacks in a builder that have the same layer diff id


matthewmcnew:
It looks like this is backfix. Could you add a description of why?

tomkennedy513:
> It looks like this is backfix. Could you add a description of why?

Hey Matt! I was working on patching this backwards to deal with issues where flattened versions of paketo buildpacks potentially causing issues with older versions of kpack.

samj1912 (approved):

PR #1269, opened 2023-07-21
[0.10.x] Backport ci

tomkennedy513:
When we recreated our ci to be open source, we never brought those changes into the release branches, so there is no way to release patches to the older versions if needed currently.


.github/workflows/ci.yaml
```diff
@@ -1,20 +1,536 @@
...
+        kapp deploy -a kpack -y -f prerelease.yaml
+
+        export IMAGE_REGISTRY=${{ env.REGISTRY_URL }}
+        export IMAGE_REGISTRY_USERNAME=${{ env.REGISTRY_USER }}
+        export IMAGE_REGISTRY_PASSWORD=${{ env.REGISTRY_PASS }}
+
+        export GIT_PRIVATE_REPO=${{ vars.E2E_PRIVATE_REPO }}
+        export GIT_BASIC_USERNAME="${{ secrets.E2E_PRIVATE_REPO_USERNAME }}"
+        export GIT_BASIC_PASSWORD="${{ secrets.E2E_PRIVATE_REPO_PASSWORD }}"
+        export GIT_SSH_PRIVATE_KEY="${{ secrets.E2E_PRIVATE_REPO_PRIVATE_KEY }}"
```
chenbh, line 371: These aren't needed in <0.11.x, but there's no real harm in having them here I guess

chenbh (approved):

.github/workflows/ci.yaml
tomkennedy513: I can remove them

PR #1274, opened 2023-07-21
Trigger unit tests on prs to release branch

tomkennedy513:


chenbh:
Shouldn't we just run unit tests on all PRs regardless of destination branch? In fact I would even be fine with running unit tests on all commits

tomkennedy513:
> Shouldn't we just run unit tests on all PRs regardless of destination branch? In fact I would even be fine with running unit tests on all commits

what other branches would someone pr to

chenbh (approved):

PR #1282, opened 2023-07-24
Update schema to match pre 0.11 changes

tomkennedy513:
- On main the config directory schema has changed. When backporting the ci workflow that was not taken into consideration so it is using the wrong schema.


chenbh (approved):
In particular, this is a partial revert of #1261

tomkennedy513:
> In particular, this is a partial revert of #1261

yes that change was intended to be only going forward

PR #1295, opened 2023-07-31
Create local script to replace existing local dev scripts

tomkennedy513:
- Support building with pack or ko locally


DEVELOPMENT.md
```diff
@@ -6,7 +6,7 @@ Access to a Kubernetes cluster is needed in order to install the kpack controlle
```
chenbh, line 6: We should mention somewhere that the dev flow is only for linux images

hack/local.sh
```diff
@@ -0,0 +1,100 @@
...
+Builds and generates a deployment yaml for kpack.
+
+Prerequisites:
+- pack or ko installed
+- kubectl installed
+- docker login to your registry
+
+OPTIONS
+  --help                          -h  prints the command usage
+  --buildType <buildType>             build system to use. valid options are pack or ko.
```
chenbh, line 21: I think CLI flags usually uses `kebab-case`

hack/build.sh
```diff
@@ -0,0 +1,174 @@
+#!/bin/bash
+
+function lifecycle_image_build() {
+    image=$1
+    go run hack/lifecycle/main.go --tag=${image}
+}
+
+function generate_kbld_config_pack() {
```
chenbh, line 8: There's something weird about this when you try to rebuild the image again. For some reason it results in a repo digest mismatch. I wonder if kbld is adding something that's causing the build to not be reproducible?

hack/local.sh
tomkennedy513: ya I can change, I took the loop from paketo which uses camel case

hack/build.sh
chenbh: This was due to my env being polluted, nuking the registry and docker pruning fixed the problem

hack/build.sh
```diff
@@ -0,0 +1,203 @@
...
+  completion_image=${IMAGE_PREFIX}completion
+  lifecycle_image=${IMAGE_PREFIX}lifecycle
+
+  echo "Building Lifecycle"
+  lifecycle_image_build ${lifecycle_image}
+
+  echo "Generating kbld config"
+  temp_dir=$(mktemp -d)
+  kbld_config_path="${temp_dir}/kbld-config"
+  if [ $type = "ko" ]; then
```
samj1912, line 192: Any reason to support both types? Should we just use the one that has the best devx?

tomkennedy513: The main reason I added both was a bit of a proof of concept. Inside VMware we have some teams that use ko, but are experimenting with using buildpacks, so I was trying to show a way that you could build the same image with either ko or pack.

Which one has the best developer experience is a bit tough to answer since it depends a bit on what you're doing. Iterating locally, ko will be a bit quicker since there is no docker involved, but it is very dependent on your local environment. Building with pack will replicate our ci so as a developer you can have confidence that the changes you are working on locally will behave the same in ci. 

I personally think there is room for both, but if I had to chose one I would pick pack because consistency would be more important to me than speed.

chenbh: The trade off is `pack` is what we use to generate our actual releases (and something-something dog fooding), so you would use it to get as close as possible to the release process/candidate. On the other hand, `ko` is fast, like 1 min vs 7-10 min. The difference is even greater when both are started from empty caches.

samj1912 (approved):

chenbh (changes requested):
Whoops, forgot that `pack` deposits the built binaries at `/cnb/process/<app>`, which we hard code at https://github.com/buildpacks-community/kpack/blob/2646bcfa7b744677a4261f26214490d4a7f03fba/pkg/apis/build/v1alpha2/build_pod.go#L399.

samj1912 (approved):

PR #1297, opened 2023-08-04
[WIP] Give the option to build image from local directory

bsyouness:


chenbh:
Hi, I'm not quite sure how would building from a local directory work. Ignoring the code changes in this PR, which seems to be a copy paste of the blob source, I don't see a way to even provide a local directory to the build pod. I guess we can mount PVCs, but I'm not sure the usefulness it.

What use cases are you trying to solve here?

bsyouness:
Hey! Thanks for commenting on the PR. Yes, that was the idea: the use for this would be to avoid having to figure out a way to serve a blob and instead mount the directory that would be the target of kpack. Thoughts?

chenbh:
I guess I don't really see a use case for it. In order for this to work, the app's source code must be made available somehow to the build pod. Which probably means that it'll have to be mounted from a kubernetes volume of some sort. But the question then is who would be populating this volume with the source code? 

I would love to learn more about how and where the app's source code will come from before thinking about how kpack will consume that source code.

chenbh:
If you're looking to use kpack for local development, i.e. use kpack to build a source code on your local machine, the kp CLI has a way to [upload a local directory as the source code via the `--local-path` flag](https://github.com/buildpacks-community/kpack-cli/blob/main/docs/kp_image_create.md) when creating Images. (Btw, this is implemented as uploading the local directory as a registry image and using registry source on the kpack Image)

samj1912:
@chenbh I think a better alternative would be allowing users to use a configmap as source code. IIRC there was an issue to support this.

samj1912:
@bsyouness would you like to tackle https://github.com/buildpacks-community/kpack/issues/237 instead? I think that might fit your use cases better.

bsyouness:
Thanks @samj1912. I'll try to spend some time on this at some point. Thank you both for the help.

PR #1307, opened 2023-08-23
Propagates container error logs to build resource and adds message to stepStates

fcaroline2020:


pkg/buildchange/progress_logger.go
```diff
@@ -0,0 +1,51 @@
...
+type ProgressLogger struct {
+	K8sClient k8sclient.Interface
+}
+
+// GetTerminationMessage gets the container logs for a given pod
+func (p *ProgressLogger) GetTerminationMessage(pod *corev1.Pod, s *corev1.ContainerStatus) (string, error) {
+	containerLog, _ := p.getContainerLogs(pod, s)
+
+	moreInfoCmd := createMoreInfoCommand(pod.Namespace, pod.Name, s.Name)
+	return " " + s.State.Terminated.Message + " " + containerLog + moreInfoCmd, nil
```
tomkennedy513, line 23: Is there a case where this can get too long that we should account for ?

fcaroline2020: That's a good idea. I've added a truncate method that will cut off at 800 characters. I think this is only relevant for containerLog since the others are mostly predefined ones that predetermined length
- `s.State.Terminated.Message`  is `Container %s terminated with error`
- `moreInfoCmd` is `kubectl logs -n %s %s -c %s`

PR #1316, opened 2023-08-30
Updates pr-1309 with code improvements suggested by Daniel Chen

fcaroline2020:
- toss the error from func 'waitBuild' up the chain instead of trying to handle it in wait_for_image
- remove else from func 'waitBuild' 
- use 'buildFailure' func to format error message


fcaroline2020:
@chenbh

chenbh (approved):

PR #1331, opened 2023-09-26
chore: remove refs to deprecated io/ioutil

testwill:


pkg/blob/fetch_test.go
```diff
@@ -78,14 +77,14 @@ func testBlobFetcher(t *testing.T, when spec.G, it spec.S) {
...
-		files, err := ioutil.ReadDir(dir)
+		files, err := os.ReadDir(dir)
 		require.NoError(t, err)
 		require.Len(t, files, 1)
 
 		testFile := files[0]
 		require.Equal(t, "some-file.txt", testFile.Name())
-
-		require.Equal(t, os.FileMode(0777).String(), testFile.Mode().String())
+		info, _ := testFile.Info()
```
tomkennedy513, line 86: maybe we should assert there is no error here

samj1912 (approved):

tomkennedy513 (approved):

PR #1339, opened 2023-10-02
[0.11.x] Fix cloning with Azure DevOps Git

tomkennedy513:
- remove multi_ack and multi_ack_detailed from unsupported capabilities
- see https://github.com/go-git/go-git/blob/52c2972976737a00fce91d7deb1278a6a460cae6/_examples/azure_devops/main.go\#L21-L36 for more info
- we should be able to remove this once go-git implements full support for v2 git with multi_ack

PR #1343, opened 2023-10-04
[0.11.x] Change name of executable in test

tomkennedy513:
- Also pinning to a commit so this doesn't happen in the future


tomkennedy513:
closing because I combined with #1339

PR #1349, opened 2023-10-09
Run ci on PRs to release branches

tomkennedy513:


chenbh:
Don't we already run the unit tests on PRs? Why do we need to full suite?

chenbh:
Oh is the expectation that since PRs to release branches are always from this repo, they will always have the correct creds to run the full suite?

tomkennedy513:
Ya I am trying to avoid ever having the release branch in a non releasable state

tomkennedy513:
Only maintainers can make prs to release branches so it will always happen from the repo

samj1912 (approved):

PR #1358, opened 2023-10-20
Support subfolder reconciliation logic

### LINKED ISSUE #1068: Support sub folder reconciliation logic
Currently if a git repository with a sub folder is used, kpack performs a rebuild even if the subfolder contents are not changed at all.

This is problematic for monorepos with lots of changes. We should support such use cases and only build when necessary.

xtreme-shane-lattanzio:
Resolves #1068

### DIFF
--- a/docs/image.md
+++ b/docs/image.md
@@ -90,7 +90,7 @@ The `source` field is a composition of a source code location and a `subpath`. I
         - `url`: The git repository url. Both https and ssh formats are supported; with ssh format requiring a [ssh secret](secrets.md#git-secrets).
         - `revision`: The git revision to use. This value may be a commit sha, branch name, or tag.
         - `initializeSubmodules`: Initialize submodules inside repo, recurses up to a max depth of 10 submodules.
-    - `subPath`: A subdirectory within the source folder where application code resides. Can be ignored if the source code resides at the `root` level.
+    - `subPath`: A subdirectory within the source folder where application code resides. New builds will only be created if a change was made in this directory. Can be ignored if the source code resides at the `root` level.
 
 * Blob
 
--- a/pkg/git/remote_git_resolver.go
+++ b/pkg/git/remote_git_resolver.go
@@ -14,6 +14,8 @@ import (
 
 const defaultRemote = "origin"
 
+var regex = regexp.MustCompile("[a-f0-9]{40}")
+
 type remoteGitResolver struct{}
 
 func (*remoteGitResolver) Resolve(auth transport.AuthMethod, sourceConfig corev1alpha1.SourceConfig) (corev1alpha1.ResolvedSourceConfig, error) {
@@ -123,7 +125,7 @@ func resolveSourceWithSubpath(auth transport.AuthMethod, sourceConfig corev1alph
 func latestCommitForSubpath(r *gogit.Repository, subPath string) (string, error) {
 	logOutput, err := r.Log(&gogit.LogOptions{
 		PathFilter: func(s string) bool {
-			if strings.Contains(s, subPath) {
+			if strings.HasPrefix(s, subPath) {
 				return true
 			} else {
 				return false
@@ -132,7 +134,6 @@ func latestCommitForSubpath(r *gogit.Repository, subPath string) (string, error)
 	})
 
 	latestCommit, err := logOutput.Next()
-	regex, _ := regexp.Compile("[a-f0-9]{40}")
 	effectiveCommit := regex.FindString(latestCommit.String())
 
 	return effectiveCommit, err
--- a/pkg/git/remote_git_resolver.go
+++ b/pkg/git/remote_git_resolver.go
@@ -2,29 +2,26 @@ package git
 
 import (
 	gogit "github.com/go-git/go-git/v5"
-	"github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/transport"
 	"github.com/go-git/go-git/v5/storage/memory"
-
 	corev1alpha1 "github.com/pivotal/kpack/pkg/apis/core/v1alpha1"
+	"regexp"
+	"strings"
 )
-
-const defaultRemote = "origin"
 
 type remoteGitResolver struct{}
 
 func (*remoteGitResolver) Resolve(auth transport.AuthMethod, sourceConfig corev1alpha1.SourceConfig) (corev1alpha1.ResolvedSourceConfig, error) {
-	remote := gogit.NewRemote(memory.NewStorage(), &config.RemoteConfig{
-		Name: defaultRemote,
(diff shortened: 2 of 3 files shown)


pkg/git/remote_git_resolver.go
```diff
@@ -2,59 +2,94 @@ package git
...
 
 func (*remoteGitResolver) Resolve(auth transport.AuthMethod, sourceConfig corev1alpha1.SourceConfig) (corev1alpha1.ResolvedSourceConfig, error) {
-	remote := gogit.NewRemote(memory.NewStorage(), &config.RemoteConfig{
-		Name: defaultRemote,
-		URLs: []string{sourceConfig.Git.URL},
-	})
+	var resolvedConfig corev1alpha1.ResolvedSourceConfig
 
-	refs, err := remote.List(&gogit.ListOptions{
+	r, err := gogit.Clone(memory.NewStorage(), nil, &gogit.CloneOptions{
```
tomkennedy513, line 18: to improve performance can we just do the clone if there is a subpath and do a list if there isn't?

chenbh (commented):
Can you also update the docs to indicate the new behaviour when using `subpath` with `git` source?

pkg/git/remote_git_resolver.go
```diff
@@ -2,59 +2,94 @@ package git
...
-	return corev1alpha1.ResolvedSourceConfig{
-		Git: &corev1alpha1.ResolvedGitSource{
-			URL:      sourceConfig.Git.URL,
-			Revision: sourceConfig.Git.Revision,
-			Type:     corev1alpha1.Commit,
-			SubPath:  sourceConfig.SubPath,
+	return resolvedConfig, err
+}
+
+func getLogs(r *gogit.Repository, subPath string) (string, error) {
```
chenbh, line 77: nit: lets make this more descriptive

- `latestCommitForSubpath`
- `latestCommit`
- `lastModifiedCommit`

pkg/git/remote_git_resolver.go
```diff
@@ -2,59 +2,94 @@ package git
...
+	logOutput, err := r.Log(&gogit.LogOptions{
+		PathFilter: func(s string) bool {
+			if strings.Contains(s, subPath) {
+				return true
+			} else {
+				return false
+			}
 		},
```
chenbh, line 85: Kinda expected, but this means that if the `subpath` is a symlink, the only time a new commit would be registered is if the symlink itself changed (i.e. to a different path), and not when the underlying target dir changes.

I guess we can call it out in the release notes, but I feel like this is obvious enough that we don't need to worry about it

pkg/git/remote_git_resolver.go
xtreme-shane-lattanzio: I can do that but this does clone a bare repo with the worktree being nil. Is it worth it to add logic to do this? The two flows are pretty different since one uses remote refs and the other uses the actual repository.

docs/image.md
```diff
@@ -90,7 +90,7 @@ The `source` field is a composition of a source code location and a `subpath`. I
         - `url`: The git repository url. Both https and ssh formats are supported; with ssh format requiring a [ssh secret](secrets.md#git-secrets).
         - `revision`: The git revision to use. This value may be a commit sha, branch name, or tag.
         - `initializeSubmodules`: Initialize submodules inside repo, recurses up to a max depth of 10 submodules.
-    - `subPath`: A subdirectory within the source folder where application code resides. Can be ignored if the source code resides at the `root` level.
+    - `subPath`: A subdirectory within the source folder where application code resides. New builds will only be created if a change was made in this directory. Can be ignored if the source code resides at the `root` level.
```
tomkennedy513, line 93: nice

pkg/git/remote_git_resolver.go
```diff
@@ -48,16 +76,66 @@ func (*remoteGitResolver) Resolve(auth transport.AuthMethod, sourceConfig corev1
...
-	}, nil
+	})
+
+	latestCommit, err := logOutput.Next()
+	regex, _ := regexp.Compile("[a-f0-9]{40}")
```
tomkennedy513, line 135: We can move this to a global const and use must parse

pkg/git/remote_git_resolver.go
chenbh: Oh, it should also match on `strings.HasPrefix`, `strings.Contains` will match if any subdir has the same name (i.e. `internal` should only match `./internal`, not `./pkg/some-pkg/internal`).

You might also need to watch out for `internal` and `./internal`, which should logically match the same thing.

matthewmcnew (approved):

pkg/git/remote_git_resolver.go
```diff
@@ -48,16 +77,66 @@ func (*remoteGitResolver) Resolve(auth transport.AuthMethod, sourceConfig corev1
...
+				return true
+			} else {
+				return false
+			}
 		},
-	}, nil
+	})
+
+	latestCommit, err := logOutput.Next()
+	regex := regexp.MustCompile(commitRegex)
```
chenbh, line 136: Let's move this regex to a top level var, compiling a regex is non-trivial and doing it every minute per git source for the exact same static regex is not ideal.

chenbh (changes requested):

pkg/git/remote_git_resolver.go
```diff
@@ -48,16 +78,65 @@ func (*remoteGitResolver) Resolve(auth transport.AuthMethod, sourceConfig corev1
...
+
+	rIter, _ := r.References()
+
+	rIter.ForEach(func(ref *plumbing.Reference) error {
+		if ref.Name().Short() == sourceConfig.Git.Revision {
+			refSourceType := sourceType(ref)
+			effectiveCommit := ""
+
+			if refSourceType == corev1alpha1.Branch {
+				effectiveCommit, err = latestCommitForSubpath(r, sourceConfig.SubPath)
```
tomkennedy513, line 100: maybe we should do a `path.Clean(sourceConfig.SubPath)` or something when the values are compared to normalize the user provided subpath to handle the case that @chenbh mentioned where currently the use can provide `./subpath` or `subpath` and these may not both match correctly even though they are functionally equivalent. It may also make sense to normalize the paths provided by gogit to ensure they match what we expect. Everything else lgtm

chenbh: I've thought about this some more and I actually think we shouldn't do this. Supporting `.` can be taken as we also support `..` (because otherwise why use relative paths at all?) which is definitely not true. `path.Clean` might be counterproductive here because it'll actually processes `..` properly.

tomkennedy513: I think the issue is that we support ./ today so we shouldn't break existing users. If someone uses ../ then that's fine. It probably won't work but it will be the same as before

tomkennedy513 (approved):

PR #1370, opened 2023-11-11
Add Builder UpToDate Condition

### LINKED ISSUE #1365: Allow builds to be created if the builder has a latest image
When kpack is determining if a build is required, it checks the status of the builder resource and will refuse to stamp out a build if the builder isn't ready. This is problematic because the builder will be not ready anytime it experiences a reconcile failure. There are many cases where a build could be kicked off using the available latest image from the builder even if the latest reconcile failed.

kpack should have a separate condition to indicate that the builder is failing to reconcile properly, leaving the ready condition to indicate whether or not he builder is capable of participating in a build (it has a latest image).

tomkennedy513:
Adds a condition for Builder UpToDate. This tracks the status of the latest reconcile of the builder, while the Ready condition indicates that the builder is able to be used in a build. The reason behind this change is that a builder reconcile may fail for a variety of reasons, but as long as there is a latestImage, builds should be able to continue.

resolves #1365

### DIFF
--- a/config/builder.yaml
+++ b/config/builder.yaml
@@ -37,6 +37,9 @@ spec:
     - name: Ready
       type: string
       jsonPath: ".status.conditions[?(@.type==\"Ready\")].status"
+    - name: UpToDate
+      type: string
+      jsonPath: ".status.conditions[?(@.type==\"UpToDate\")].status"
   conversion:
     strategy: Webhook
     webhook:
--- a/config/clusterbuilder.yaml
+++ b/config/clusterbuilder.yaml
@@ -37,6 +37,9 @@ spec:
     - name: Ready
       type: string
       jsonPath: ".status.conditions[?(@.type==\"Ready\")].status"
+    - name: UpToDate
+      type: string
+      jsonPath: ".status.conditions[?(@.type==\"UpToDate\")].status"
   names:
     kind: ClusterBuilder
     listKind: ClusterBuilderList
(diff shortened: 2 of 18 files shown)


chenbh:
Feel free to push back if you think this is scope creep, but what do you think about printing a warning in the build logs if the builder is not up to date?

I'm a little bit worried about how to surface this info to the user, I highly doubt people are in the habit of checking their builders/stacks/buildpacks if their Images are happy. While they probably won't check build logs that much either, it would at be something. Maybe we can also bubble this up in the Image as well, but that might be too complex to introduce a "Warning" state in addition to the current "Error" states.

tomkennedy513:
> Feel free to push back if you think this is scope creep, but what do you think about printing a warning in the build logs if the builder is not up to date?
> 
> 
> 
> I'm a little bit worried about how to surface this info to the user, I highly doubt people are in the habit of checking their builders/stacks/buildpacks if their Images are happy. While they probably won't check build logs that much either, it would at be something. Maybe we can also bubble this up in the Image as well, but that might be too complex to introduce a "Warning" state in addition to the current "Error" states.

I thought about this but the build doesn't really know anything about the builder resource so I wasn't sure if it made a ton of sense. I'm still open to it but I think I  would prefer it in a follow up pr

chenbh:
> I thought about this but the build doesn't really know anything about the builder resource so I wasn't sure if it made a ton of sense. 

I was imagining something simple like a `--warn-builder-not-up-to-date=clusterbuilder/my-builder` flag in `cmd/build-init`. So the build reconciler would decide to inject this flag and the `prepare` step would just dutifully print it.

> I'm still open to it but I think I would prefer it in a follow up pr

👍

chenbh (approved):

PR #1374, opened 2023-11-13
RFC: SLSA attestation support

chenbh:
[Readable](https://github.com/buildpacks-community/kpack/blob/slsa/rfcs/0000-slsa-attestation.md)


rfcs/0000-slsa-attestation.md
```diff
@@ -0,0 +1,361 @@
...
+          "content": "{\"name\":\"test-build-1\",\"resourceVersion\":\"5\"}"
+        },
+        {
+          "name": "ServiceAccount",
+          "content": "{\"name\":\"default\",\"resourceVersion\":\"269\"}"
+        },
+        {
+          "name": "Secrets",
+          "content": "[{\"name\":\"my-registry-creds\",\"resourceVersion\":\"53\"},{\"name\":\"my-git-creds\"}]"
+        },
```
chenbh, line 207: I'm not too sold on this, I mainly included it as a way to trace an attestation back to a particular k8s object version (hence the `resourceVersion`). But I'm not too sure how practical this information will be, I guess it lets you know if the current k8s object is or is not the same as the one used during the build?

rfcs/0000-slsa-attestation.md
```diff
@@ -0,0 +1,365 @@
...
+          "paketo-buildpacks/executable-jar": "6.8.2",
+          "paketo-buildpacks/git": "1.0.7",
+          "paketo-buildpacks/dist-zip": "5.6.7",
+          "paketo-buildpacks/environment-variables": "4.5.6",
+          "paketo-buildpacks/environment-variables": "4.5.5",
+          "paketo-buildpacks/image-labels": "4.5.5",
+          "paketo-buildpacks/image-labels": "4.5.4",
+          "paketo-buildpacks/java": "10.3.1",
+          "paketo-buildpacks/go": "4.6.0"
+        }
```
chenbh, line 180: While this records the version of buildpacks that was used in the builder, it doesn't record the order. I feel like if you have the builder image's digest it shouldn't matter too much since you're more interested in the integrity of the build rather than reproducing it.

rfcs/0000-slsa-attestation.md
```diff
@@ -0,0 +1,365 @@
...
+[dsse-envelope]: https://github.com/secure-systems-lab/dsse/blob/v1.0.0/envelope.md
+[cosign-attestation]: https://github.com/sigstore/cosign/blob/main/specs/ATTESTATION_SPEC.md
+[cosign-tag]: https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md#tag-based-discovery
+
+### Key types
+
+At minimum, the implementation should support the following key types:
+- [cosign key][cosign-key-format]. This is to ensure interop with `cosign verify-attestation`.
+- ECDSA private keys. This is due to it being the recommended SLSA suite.
+- RSA private keys. This is due to it being an extreamly popular key type.
```
chenbh, line 277: See the additional considerations down below, tldr: the only verification tool that works with key-based attestation is `cosign`, but they don't support non cosign-format keys. So there's no (nice) way of verifying ECDSA and RSA key signatures.

rfcs/0000-slsa-attestation.md
matthewmcnew: Do we expect consumers of the SLSA provenance to have access to the builder?

rfcs/0000-slsa-attestation.md
matthewmcnew: Should we include the resource version as well?

rfcs/0000-slsa-attestation.md
chenbh: I'm going to interpret your question as "do we expect them to have access to the registry that the builder and app images are pushed to" and not "do we expect them to have permissions to pull the builder image". I consider permissions a human/procedural problem and not an infrastructure concern.

If they don't care about build reproducibility then it shouldn't matter, simply recording the digest is enough to verify the integrity of the build.

If they do want to be able to reproduce the build bit-for-bit, then there's really no way to do it without having access to the registry. Because if you want to completely recreate the builder from yaml, then at some point you would need the buildpackage images (and the stack images) which also requires you to have the access to the registry.

rfcs/0000-slsa-attestation.md
chenbh: Good catch, don't know why I included it in the first secret but not the second

rfcs/0000-slsa-attestation.md
matthewmcnew: >  I feel like if you have the builder image's digest it shouldn't matter too much since you're more interested in the integrity of the build rather than reproducing it.

How do we define the integrity of the build? I asked because this provides the buildpacks under the assumption that the builder digest and the buildpacks are sufficient. But, can't the output of the build be different with different orders of buildpacks?

chenbh: I think the core of my thinking is that anything with a checksum can be treated as a immutable basic dependency and verifying it is outside the scope of SLSA (or at least kpack's involvement in SLSA). And from my understanding, a different order would result in a different image config which would result in a different digest. 

To put it differently, I think kpack should treat the builder image as a dependency from an external system, not an artifact that was generated by kpack itself. This is to further emphasize that SLSA should apply to only the Build resource, not the entire system that kpack provides (ClusterStack, Buildpacks, Builder, Image).

chenbh: > I think the core of my thinking is that anything with a checksum can be treated as a immutable basic dependency and verifying it is outside the scope of SLSA (or at least kpack's involvement in SLSA)

The reason I hold this axiom is that it would get really ugly if we have to start considering "how can I verify that the checksums in the `go.sum` of my app is correct?"

chenbh: After thinking about it a bit more, I think I'm going to generalize the annotations section to be "the labels associated with the builder image". This would save us from doing any special parsing of the image (like figuring out buildpack version or order grouping), while surfacing more metadata (stack id, release date, distro, etc) and being more future-proof.

chenbh:
@matthewmcnew I reworked the RFC so that there's 2 attestations schemas generated, the app image (Build resource), and the builder image (Builder/ClusterBuilder).

samj1912 (approved):

rfcs/0000-slsa-attestation.md
```diff
@@ -0,0 +1,528 @@
...
+
+- SLSA is relatively early in its life, while it's part of the Open Source Security Foundation (OpenSSF), industry
+  adoption is still getting started.
+
+## Additional considerations and open questions
+- Should the signing key secrets require an additional annotation instead of just matching on the type/field? This would
+  bring more inline with the [Cluste wide cosign RFC][cluster-wide-cosign-rfc]
+- Currently there's no tools on the market that can actually verify generic key-based attestations.
+  [slsa-verifier][slsa-verifier] uses keyless signing and ony supports GitHub Actions and Google Cloud Build,
+  `cosign` does key based verification, but only with its own key type.
```
natalieparellano, line 514: Where did we land here? How do we expect users of kpack to validate these attestations?

tomkennedy513 (approved):

tomkennedy513:
@chenbh can we merge this?

chenbh:
> @chenbh can we merge this?

Yeah we can, I completely forgot about this PR. I'll force push it to get the required tests triggering again

PR #1389, opened 2023-11-22
Get dependabot to watch the `release/**` branches

chenbh:
I haven't fully tested this and am not really sure how well this will turn out.

I don't think you can configure dependabot to only create security related dep bumps, so I assume we're gonna get hit by a bajillion PRs for the older releases.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,3 +15,25 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+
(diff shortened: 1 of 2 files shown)


tomkennedy513 (approved):

chenbh:
My plan is to turn this on and re-evaluate how much maintainer burden it is to triage these PRs in about a month or 2

samj1912 (approved):

PR #1436, opened 2023-11-29
Change where the keychain dockercreds are read

### LINKED ISSUE #1353: Secrets mounted as volumes are not watched for changes
When docker creds are passed in as [`CREDENTIAL_PROVIDER_SECRET_PATH`](https://github.com/buildpacks-community/kpack/blob/4e9236b54ec77232ced637653de09bd4f3176981/pkg/dockercreds/volume_secret_keychain.go#L13C1-L13C1), the values are only read [once at startup](https://github.com/buildpacks-community/kpack/blob/4e9236b54ec77232ced637653de09bd4f3176981/pkg/dockercreds/parse_secrets.go#L14).

If these secrets ever change (i.e. as part of a rotation), the controller has to be restarted in order to use the new values.

xtreme-shane-lattanzio:
Resolves #1353


chenbh (changes requested):
I don't think this is a good idea, opening and reading the file _every time_ a secret is requested (aka every single reconcile) is going to be a lot of wasted resources.

What do you think of calling `os.Stat()` instead? We could cache the last modified time of the file and re-parse the file if the new modified time is newer. `stat()`should be a lot more performant than `open()` since it's only doing a lookup and not allocating anything that needs to be cleaned up (i.e. every `open()` must have a `close()`)

matthewmcnew:
Is this an opportunity a single reconcile loop could continually re-read them?

xtreme-shane-lattanzio:
@chenbh I can add a condition in the NewVolumeSecretKeychain method to os.stat the secret file but where are you thinking would we cache the time to compare against?

chenbh:
@xtreme-shane-lattanzio I would have the `NewVolumeSecretKeychain` return a struct that implements the `authn.Keychain` interface, and so every time `Resolve` is called, it can compare last modified time against its internal state. I'm not 100% sure, but you might need to secure it with a mutex cause the KeychainFactor is shared among all the reconcilers which will run in separate goroutines.

@matthewmcnew I'm not too convinced about the benefits of continuously reading the file especially since we don't expect the underlying Secret to be rotated often.

PR #1448, opened 2023-12-11
Add in the initial configuration to add the image labels

### LINKED ISSUE #1442: Add the ability to add arbitrary image labels to a built `Builder` image
https://github.com/buildpacks-community/kpack/blob/5a99b7df74938dc1e3c8f0802ed962a2bb119c8f/pkg/cnb/builder_builder.go#L154

As you can see here there is nothing in the spec that allows for folks to specify any custom metadata about a Builder image.

This can cause issues for groups that assign meaning/filtering functions based on labels of images inside of their repositories.

alexbarbato:
Closes #1442

### DIFF
--- a/pkg/apis/build/v1alpha2/builder_types.go
+++ b/pkg/apis/build/v1alpha2/builder_types.go
@@ -36,7 +36,8 @@ type BuilderSpec struct {
 	Stack corev1.ObjectReference `json:"stack,omitempty"`
 	Store corev1.ObjectReference `json:"store,omitempty"`
 	// +listType
-	Order []BuilderOrderEntry `json:"order,omitempty"`
+	Order            []BuilderOrderEntry `json:"order,omitempty"`
+	AdditionalLabels map[string]string   `json:"additionalLabels,omitempty"`
 }
 
 // +k8s:openapi-gen=true
--- a/pkg/cnb/builder_builder.go
+++ b/pkg/cnb/builder_builder.go
@@ -51,6 +51,7 @@ type builderBlder struct {
 	runImage          string
 	mixins            []string
 	os                string
+	additionalLabels  map[string]string
 }
 
 func newBuilderBldr(kpackVersion string) *builderBlder {
@@ -150,6 +151,8 @@ func (bb *builderBlder) WriteableImage() (v1.Image, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	image, err = imagehelpers.SetStringLabels(image, bb.additionalLabels)
 
 	return imagehelpers.SetLabels(image, map[string]interface{}{
 		buildpackOrderLabel:  bb.order,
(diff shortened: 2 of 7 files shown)


alexbarbato:
Mostly opening for feedback and to learn what I'm missing.

Open Q:
- I intentionally put the custom images before the standard images so that way any conflict that a user were attempt to make would just get overwritten. Open to thoughts on if we should provide errors or something for "protected labels" - feels like overkill to me.
- Anything I'm missing?

pkg/cnb/builder_builder.go
```diff
@@ -151,6 +152,8 @@ func (bb *builderBlder) WriteableImage() (v1.Image, error) {
 		return nil, err
 	}
 
+	image, err = imagehelpers.SetStringLabels(image, bb.additionalLabels)
```
matthewmcnew, line 155: Add an err check

matthewmcnew, line 155: Add an error check here.

pkg/cnb/create_builder.go
```diff
@@ -65,6 +65,8 @@ func (r *RemoteBuilderCreator) CreateBuilder(ctx context.Context, builderKeychai
 		builderBldr.AddGroup(buildpacks...)
 	}
 
+	builderBldr.additionalLabels = spec.AdditionalLabels
```
matthewmcnew, line 68: All the other steps are modifications via setters on the builderBldr. Can we add a set additional labels?

alexbarbato: happy to, do you think it adds value though?

I guess said differently I'm probably going to have that setter effectively is going to say exactly what I say here - which may be because I'm not writing good go code or something

alexbarbato:
Thank you for the review btw @matthewmcnew

pkg/cnb/builder_builder.go
alexbarbato: updated!

alexbarbato:
Anything else y'all needed from me here?

pkg/apis/build/v1alpha2/builder_types.go
```diff
@@ -36,7 +36,8 @@ type BuilderSpec struct {
 	Stack corev1.ObjectReference `json:"stack,omitempty"`
 	Store corev1.ObjectReference `json:"store,omitempty"`
 	// +listType
-	Order []BuilderOrderEntry `json:"order,omitempty"`
+	Order            []BuilderOrderEntry `json:"order,omitempty"`
+	AdditionalLabels map[string]string   `json:"additionalLabels,omitempty"`
```
samj1912, line 40: Can you regenerate the CRDs as well?

alexbarbato: I just added a new commit.

Please let me know if you think it's not the right place

matthewmcnew (approved):

PR #1449, opened 2023-12-13
SLSA attestations for the Build resource (aka app image attestations)

chenbh:
Implementation of #1374, the feature is marked experimental until the RFC has been merged.

This PR adds SLSA attestations to the Build resource:
- `EXPERIMENTAL_GENERATE_SLSA_ATTESTATION="true"` can be set on the controller to enable these behaviours
- A new status `.status.latestAttestationImage` will stamp out the immutable digest of the attestation
- The attestation is also pushed to a mutable image tag based off of the app image's digest (digest `my-app-image@sha:1234` will have the attestation stored at tag `my-app-image:sha256-1234`). This is consistent with how `cosign` stores their attestations and can be verified with `cosign verify-attestation ...`
- Recommend viewing signed images with `cosign verify-attestation --type=slsaprovenance1 ...`, alternatively unsigned or signed attestations can be viewed with

    ```
    crane export <ATTESTATION_IMAGE> - | jq -r '.payload' | base64 -D | jq
    ```
- Any secrets attached to the Build's service account, or the kpack controller's service account with the annotation `kpack.io/slsa: ""` will be used as private keys for signing
- Signing RSA, ECDSA, ED25519, and Cosign keys are supported, the way to use them is specified in the RFC

Things still needed:
- [x] E2E tests
- [x] Docs about how to use/configure it
- [x] Docs about SLSA level provided by kpack


chenbh:
@matthewmcnew @samj1912 @AidanDelaney can I get some non-Broadcom 👀 on this or the RFC?

config/controller.yaml
```diff
@@ -98,6 +98,8 @@ spec:
           value: "false"
         - name: INJECTED_SIDECAR_SUPPORT
           value: "false"
+        - name: EXPERIMENTAL_GENERATE_SLSA_ATTESTATION
```
tomkennedy513, line 101: what do you see the process looking like when we want to make this non experimental

tomkennedy513: could we just document it as experimental but keep the env as `GENERATE_SLSA_ATTESTATION` to minimize any tech debt we might take on

chenbh: The only changes needed is to update a single line for parsing command-line/env var flag:
https://github.com/buildpacks-community/kpack/blob/84c91fa7dbecedcf590de63439f02ac510df85cb/cmd/controller/main.go#L96

All the internals use the non-prefixed name
https://github.com/buildpacks-community/kpack/blob/84c91fa7dbecedcf590de63439f02ac510df85cb/pkg/config/config.go#L16
https://github.com/buildpacks-

chenbh: I'm a little hesitant to remove the `EXPERIMENTAL` part since the RFC hasn't been merged and there hasn't been a lot of feedback regarding this feature just yet

pkg/blob/fetch.go
```diff
@@ -36,9 +41,15 @@ func (f *Fetcher) Fetch(dir string, blobURL string, stripComponents int) error {
...
 
+	checksum, err := sha256sum(file)
+	if err != nil {
+		return err
+	}
+
 	switch mediaType {
 	case "application/zip":
-		info, err := file.Stat()
+		var info fs.FileInfo
```
AidanDelaney, line 51: Why do we move from short declaration to `var` here?

pkg/blob/fetch.go
```diff
@@ -57,6 +68,27 @@ func (f *Fetcher) Fetch(dir string, blobURL string, stripComponents int) error {
 		return err
 	}
 
+	projectMetadataFile, err := os.Create(path.Join(metadataDir, "project-metadata.toml"))
```
AidanDelaney, line 71: Should `"project-metadata.toml"` be defined as a constant somehwere?  We use that string in this function and in the associated tests.

pkg/blob/fetch.go
```diff
@@ -105,3 +137,36 @@ func classifyFile(reader io.ReadSeeker) (string, error) {
...
 }
+
+func sha256sum(reader io.ReadSeeker) (string, error) {
+	hash := sha256.New()
+	_, err := io.Copy(hash, reader)
+	if err != nil {
+		return "", err
+	}
+
+	_, err = reader.Seek(0, 0)
```
AidanDelaney, line 148: Nice! :) I would have assumed the reader was zero'ed and then spent hours debugging this when the inevitable bug report landed.

pkg/cosign/image_signer.go
```diff
@@ -210,40 +213,24 @@ func (s *ImageSigner) SignBuilder(
 			},
 		)
 
-		if _, found := os.LookupEnv(cosignutil.CosignDockerMediaTypesEnv); found {
-			err = os.Unsetenv(cosignutil.CosignDockerMediaTypesEnv)
+		if _, found := os.LookupEnv(CosignDockerMediaTypesEnv); found {
+			err = os.Unsetenv(CosignDockerMediaTypesEnv)
```
AidanDelaney, line 217: Is this to prevent an end-user from injecting an incorrect media type through an environmental variable?

pkg/cosign/image_signer_test.go
```diff
@@ -93,16 +93,16 @@ func testImageSigner(t *testing.T, when spec.G, it spec.S) {
...
+				publicKey1 = path.Join(secretLocation, "secret-name-1", secret.CosignSecretPublicKey)
+				publicKey2 = path.Join(secretLocation, "secret-name-2", secret.CosignSecretPublicKey)
+				passwordFile1 = path.Join(secretLocation, "secret-name-1", secret.CosignSecretPassword)
+				passwordFile2 = path.Join(secretLocation, "secret-name-2", secret.CosignSecretPassword)
 
 				report = createReportToml(t, expectedImageName, imageDigest)
 
-				os.Unsetenv(cosignutil.CosignRepositoryEnv)
-				os.Unsetenv(cosignutil.CosignDockerMediaTypesEnv)
+				os.Unsetenv(CosignRepositoryEnv)
```
AidanDelaney, line 104: Do we have any guarantees about the order of the tests?  AFAIK unsetting these env vars here have an impact on tests that run subsequently.  Do we want to change the environment for subsequent tests?

pkg/reconciler/build/build.go
```diff
@@ -154,7 +188,7 @@ func (c *Reconciler) reconcile(ctx context.Context, build *buildapi.Build) error
 			})
 
 			if err != nil {
-				return errors.Wrap(err, "unable to create app image keychain")
+				return fmt.Errorf("unable to create app image keychain: %v", err)
```
AidanDelaney, line 191: Why do  we choose to stringify the message instead of wrapping it?  Do we wish to drop the stack trace here?

pkg/reconciler/build/build.go
```diff
@@ -164,12 +198,22 @@ func (c *Reconciler) reconcile(ctx context.Context, build *buildapi.Build) error
...
-				return errors.Wrap(err, "failed to get build metadata from build pod")
+				return fmt.Errorf("failed to get build metadata from build pod: %v", err)
 			}
 		}
+
+		var attestDigest string
+		if c.FeatureFlags.GenerateSlsaAttestation {
+			attestDigest, err = c.attestBuild(ctx, build, buildMetadata, pod)
+			if err != nil {
+				return fmt.Errorf("attesting build: %v", err)
```
AidanDelaney, line 209: Could we add a "failed to" like our other error messages?

```suggestion
				return fmt.Errorf("failed to attest build: %v", err)
```

pkg/reconciler/build/build.go
```diff
@@ -358,6 +402,105 @@ func (c *Reconciler) buildMetadataFromBuildPod(pod *corev1.Pod) (*cnb.BuildMetad
...
+		Namespace:        build.Namespace,
+		ImagePullSecrets: build.Spec.Builder.ImagePullSecrets,
+	})
+	if err != nil {
+		return "", err
+	}
+
+	controllerSecrets, err := c.SecretFetcher.SecretsForSystemServiceAccount(ctx)
+	if err != nil {
+		return "", fmt.Errorf("getting controller secrets: %v", err)
```
AidanDelaney, line 417:
```suggestion
		return "", fmt.Errorf("failed to get attestation controller secrets: %v", err)
```

pkg/reconciler/build/build.go
```diff
@@ -358,6 +402,105 @@ func (c *Reconciler) buildMetadataFromBuildPod(pod *corev1.Pod) (*cnb.BuildMetad
...
+	}
+
+	buildSecrets, err := c.SecretFetcher.SecretsForServiceAccount(ctx, build.ServiceAccount(), build.Namespace)
+	if err != nil {
+		return "", fmt.Errorf("getting service account secrets: %v", err)
```
AidanDelaney, line 422:
```suggestion
		return "", fmt.Errorf("failed to get service account secrets: %v", err)
```

pkg/reconciler/build/build.go
```diff
@@ -358,6 +402,105 @@ func (c *Reconciler) buildMetadataFromBuildPod(pod *corev1.Pod) (*cnb.BuildMetad
...
+	}
+
+	secrets := append(controllerSecrets, buildSecrets...)
+	signingKeys, err := secret.FilterAndExtractSLSASecrets(secrets)
+	if err != nil {
+		return "", fmt.Errorf("parsing slsa secrets: %v", err)
```
AidanDelaney, line 428:
```suggestion
		return "", fmt.Errorf("failed to parse slsa secrets: %v", err)
```

pkg/reconciler/build/build.go
```diff
@@ -358,6 +402,105 @@ func (c *Reconciler) buildMetadataFromBuildPod(pod *corev1.Pod) (*cnb.BuildMetad
...
+	}
+
+	buildId := slsa.UnsignedBuildID
+	if len(signers) > 0 {
+		buildId = slsa.SignedBuildID
+	}
+
+	deps, err := c.attestBuildDeps(ctx, build, pod, secrets)
+	if err != nil {
+		return "", fmt.Errorf("gathering build deps: %v", err)
```
AidanDelaney, line 453:
```suggestion
		return "", fmt.Errorf("failed to gather build deps: %v", err)
```

pkg/reconciler/build/build.go
```diff
@@ -358,6 +402,105 @@ func (c *Reconciler) buildMetadataFromBuildPod(pod *corev1.Pod) (*cnb.BuildMetad
...
+	}
+
+	statement, err := c.Attester.AttestBuild(build, buildMetadata, pod, keychain, buildId, deps...)
+	if err != nil {
+		return "", fmt.Errorf("generating statement: %v", err)
```
AidanDelaney, line 458:
```suggestion
		return "", fmt.Errorf("failed to generate attestation statement: %v", err)
```

pkg/reconciler/build/build.go
```diff
@@ -358,6 +402,105 @@ func (c *Reconciler) buildMetadataFromBuildPod(pod *corev1.Pod) (*cnb.BuildMetad
...
+	}
+
+	payload, err := c.Attester.Sign(ctx, statement, signers...)
+	if err != nil {
+		return "", fmt.Errorf("signing statement: %v", err)
```
AidanDelaney, line 463:
```suggestion
		return "", fmt.Errorf("failed to sign attestation statement: %v", err)
```

pkg/reconciler/build/build.go
```diff
@@ -358,6 +402,105 @@ func (c *Reconciler) buildMetadataFromBuildPod(pod *corev1.Pod) (*cnb.BuildMetad
...
+	}
+
+	_, digest, err := c.Attester.Write(ctx, buildMetadata.LatestImage, payload, keychain)
+	if err != nil {
+		return "", fmt.Errorf("writting attestation: %v", err)
```
AidanDelaney, line 468:
```suggestion
		return "", fmt.Errorf("failed to write attestation: %v", err)
```

pkg/registry/fetch.go
```diff
@@ -59,10 +65,31 @@ func (f *Fetcher) Fetch(dir, registryImage string) error {
...
 	}
 
-	if err := handler(img, dir); err != nil {
+	if err = handler(img, dir); err != nil {
 		return err
 	}
 
+	projectMetadataFile, err := os.Create(path.Join(metadataDir, "project-metadata.toml"))
+	if err != nil {
+		return errors.Wrapf(err, "invalid metadata destination '%s/project-metadata.toml' for image: %s", metadataDir, registryImage)
```
AidanDelaney, line 74: We're using `errors.Wrap` where and not `fmt.Errorf`.  When do we wrap and when we `Errorf`?

pkg/secret/signing_secret.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func FilterAndExtractSLSASecrets(secrets []*corev1.Secret) ([]SigningKey, error) {
+	cosignSecrets := filterCosignSecrets(secrets, SLSASecretAnnotation)
+	privKeySecrets := filterPrivateKeySecrets(secrets, SLSASecretAnnotation)
+
+	return extractAttestationKeyFromSecrets(cosignSecrets, privKeySecrets)
+}
+
+func filterCosignSecrets(serviceAccountSecrets []*corev1.Secret, annotation string) []*corev1.Secret {
```
AidanDelaney, line 34: Is this `filterCosignSecrets` or `gatherCosignSecrets`?  I'd expect a filter to apply a predicate to some collection.  Here it looks like we're gathering the secrets and packing them into a struct, should they exist.

pkg/slsa/attest.go
```diff
@@ -0,0 +1,258 @@
...
+
+	Images   config.Images
+	Features config.FeatureFlags
+	Config   config.Config
+}
+
+func (a *Attester) AttestBuild(build *buildv1alpha2.Build, buildMetadata *cnb.BuildMetadata, pod *corev1.Pod, builderAndAppKeychain authn.Keychain, builderId BuilderID, depFns ...BuilderDependencyFn) (intoto.Statement, error) {
+	builderRepo, builderSha, builderLabels, err := a.ImageReader.Read(builderAndAppKeychain, build.Spec.Builder.Image)
+	if err != nil {
+		return intoto.Statement{}, fmt.Errorf("reading builder image: %v", err)
```
AidanDelaney, line 49:
```suggestion
		return intoto.Statement{}, fmt.Errorf("failed to read builder image: %v", err)
```

pkg/slsa/attest.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	appRepo, appSha, appLabels, err := a.ImageReader.Read(builderAndAppKeychain, buildMetadata.LatestImage)
+	if err != nil {
+		return intoto.Statement{}, fmt.Errorf("reading app image: %v", err)
```
AidanDelaney, line 54:
```suggestion
		return intoto.Statement{}, fmt.Errorf("failed to read app image: %v", err)
```

pkg/slsa/attest.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	source, sourceDigest, err := extractSourceFromLabel(appLabels)
+	if err != nil {
+		return intoto.Statement{}, fmt.Errorf("extracting source from label: %v", err)
```
AidanDelaney, line 59:
```suggestion
		return intoto.Statement{}, fmt.Errorf("failed to extract source from label: %v", err)
```

pkg/slsa/attest.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	start, stop, err := getStartStopTime(pod)
```
AidanDelaney, line 62: I don't expect this to ever fail, but it _appears_ that the `StartedOn`/`FinishedOn` is optional in https://pkg.go.dev/github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v1#BuildMetadata.  Should this failure stop the attestation process?

AidanDelaney (commented):
I _hate_ to provide a partial review, but I have limited time today.  Will complete the review tomorrow.  Most of my comments are trivial and can be ignored as style changes.  I have one question on whether the start/stop time is _necessary_ in an attestation.

pkg/blob/fetch.go
chenbh: 💯. But I would like to address this in a follow-up PR. I was going to put this const in the `github.com/buildpacks-community/kpack/pkg/cnb` package, but ran into #1473 and that requires some decent refactoring that I would rather not be in this PR.

pkg/slsa/attest.go
chenbh: Good catch, I've made `getStartStopTime` do a best effort in parsing and have no failure states

pkg/blob/fetch.go
AidanDelaney: Thanks

pkg/blob/fetch.go
AidanDelaney: Makes sense to me

tomkennedy513 (approved):
this is awesome

docs/slsa.md
```diff
@@ -0,0 +1,240 @@
...
+```
+
+The envelope is stored as uncompressed text in the first layer of the attestation image. The image (and the registry)
+is treated as a blobstore and isn't intended to be a container image. That is, `docker pull $ATTESTATION_TAG` or
+trying to run the image in any way will **not** work.
+
+If you want to access the attestation, you must use one of the tools that interact with the registry directly.
+
+All of the following examples assume you have [jq](https://jqlang.github.io/jq/) installed. Given an `IMAGE_DIGEST`
+`registry.com/my/repo@sha256:1234`, the `ATTESTATION_TAG` would be `registry.com/my/repo:sha256-1234`
```
natalieparellano, line 92:
```suggestion
`registry.com/my/repo@sha256:1234`, the `ATTESTATION_TAG` would be `registry.com/my/repo:sha256-1234.att`
```

natalieparellano (commented):
Amazing work. I only reviewed the docs files

PR #1483, opened 2024-01-16
RFC to create a lifecycle crd

tomkennedy513:
[readable](https://github.com/buildpacks-community/kpack/blob/lifecycle-crd/rfcs/0000-lifecycle-crd.md)

### DIFF
--- a/rfcs/0000-lifecycle-crd.md
+++ b/rfcs/0000-lifecycle-crd.md
@@ -1,0 +1,137 @@
+`[Readable](https://github.com/your-name-or-org/kpack/blob/<pr-branch-name>/rfcs/0000-<my-feature>.md)`
+
+**Problem:**
+
+The CNB lifecycle is provided to kpack as in image reference in a configMap similar to how the completion and prepare
+images are passed to the kpack controller. This means that every builder will always use the same exact lifecycle, with
+no ability to specify an alternate one that may support different buildpack api or platform api versions. Due to kpack
+support of windows, a custom lifecycle image with specific layers for the linux and windows binaries is required
+(with a label specifying the layer diff id for each os). Because of this, a cve present in the lifecycle that has been 
+patched upstream, requires any user to build the kpack specific lifecycle image themselves if the kpack provided one
+has not been bumped.
+
+As we [move to deprecate windows](https://github.com/buildpacks-community/kpack/discussions/1366), there is an opportunity for us to utilize
+the upstream lifecycle image shipped by CNB, but switching to that image without changing the interface for providing the
+lifecycle to kpack may open the door to incompatibilities if the wrong image is used on a particular kpack version.
+
+**Outcome:**
+
+Users will be able to create a CR that references a lifecycle image. The file structure in the image must match 
+that of the [CNB Lifecycle image](https://hub.docker.com/r/buildpacksio/lifecycle/tags).
+
+```
+.
+└── cnb/
+    └── lifecycle/
+        ├── analyzer
+        ├── builder
+        ├── detector
+        ├── exporter
+        ├── launcher
+        ├── lifecycle.toml
+        └── restorer
+```
+
+They will be able to use this lifecycle CR in any builder. A lifecycle will be shipped out of the box in kpack and 
+users will not be required to provide a lifecycle reference in the builder if they do not wish to override the default.
+Users should see no difference in functionality compared to the existing kpack lifecycle if they do not choose to 
+provide their own lifecycle CR.
+
+**Actions to take:**
+
+A new ClusterLifecycle CRD will be created. It will be very similar in structure to the ClusterStack Resource. 
+One open question is whether we should follow the Clusterstack model and keep this resource strictly cluster scoped or
+follow the ClusterBuildpack pattern and provide a namespaced scoped version of the CRD. 
+
+This is an example instance of the proposed CRD:
+```yaml
+apiVersion: kpack.io/v1alpha2
+kind: ClusterLifecycle
+metadata:
+  name: sample-cluster-lifecycle
+spec:
+  serviceAccountRef:
+    name: sample-sa
+    namespace: sample-namespace
+  image: buildpacksio/lifecycle@sha256:f4ce143ea6bbc6b5be5f4d151aba8214324adb93bbd7e3b1f43cd134ad450bf7
+```
+
+If we were to create a namespaced version of this CRD it would look like this. 
+Similar to Buildpack CRs, ClusterBuilders will not be able to use the namespaced CR:
+```yaml
+apiVersion: kpack.io/v1alpha2
+kind: Lifecycle
+metadata:
+  name: sample-cluster-lifecycle
+  namespace: test
+spec:
+  serviceAccountName: sample-sa
+  image: buildpacksio/lifecycle@sha256:f4ce143ea6bbc6b5be5f4d151aba8214324adb93bbd7e3b1f43cd134ad450bf7
+```
+
+The existing lifecycle reconciler can be updated to reconcile this new CRD instead of ConfigMaps.
+
+A new field will be added to the ClusterBuilder/Builder resources allowing the user to provide a reference to a lifecycle CR.
+This field will be optional and will default to the highest semantic version that is supported by the buildpacks in the builder and the platform. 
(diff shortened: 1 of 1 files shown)


rfcs/0000-lifecycle-crd.md
```diff
@@ -0,0 +1,137 @@
...
+
+The CNB lifecycle is provided to kpack as in image reference in a configMap similar to how the completion and prepare
+images are passed to the kpack controller. This means that every builder will always use the same exact lifecycle, with
+no ability to specify an alternate one that may support different buildpack api or platform api versions. Due to kpack
+support of windows, a custom lifecycle image with specific layers for the linux and windows binaries is required
+(with a label specifying the layer diff id for each os). Because of this, a cve present in the lifecycle that has been 
+patched upstream, requires any user to build the kpack specific lifecycle image themselves if the kpack provided one
+has not been bumped.
+
+As we [move to deprecate windows](https://github.com/buildpacks-community/kpack/discussions/1366), there is an opportunity for us to utilize
```
natalieparellano, line 13: We do ship "lifecycle images" for Windows, so in theory you could specify the Windows image if Windows support is desired.

rfcs/0000-lifecycle-crd.md
```diff
@@ -0,0 +1,137 @@
...
+They will be able to use this lifecycle CR in any builder. A lifecycle will be shipped out of the box in kpack and 
+users will not be required to provide a lifecycle reference in the builder if they do not wish to override the default.
+Users should see no difference in functionality compared to the existing kpack lifecycle if they do not choose to 
+provide their own lifecycle CR.
+
+**Actions to take:**
+
+A new ClusterLifecycle CRD will be created. It will be very similar in structure to the ClusterStack Resource. 
+One open question is whether we should follow the Clusterstack model and keep this resource strictly cluster scoped or
+follow the ClusterBuildpack pattern and provide a namespaced scoped version of the CRD.
```
natalieparellano, line 44: What are the pros and cons here? I could imagine, con of providing additional namespaced scoepd resource is increased complexity. Anything else here?

I'd vote to do whatever is simplest and add more stuff later if it is ever needed or requested.

rfcs/0000-lifecycle-crd.md
```diff
@@ -0,0 +1,137 @@
...
+  namespace: test
+spec:
+  serviceAccountName: sample-sa
+  image: buildpacksio/lifecycle@sha256:f4ce143ea6bbc6b5be5f4d151aba8214324adb93bbd7e3b1f43cd134ad450bf7
+```
+
+The existing lifecycle reconciler can be updated to reconcile this new CRD instead of ConfigMaps.
+
+A new field will be added to the ClusterBuilder/Builder resources allowing the user to provide a reference to a lifecycle CR.
+This field will be optional and will default to the highest semantic version that is supported by the buildpacks in the builder and the platform.
```
natalieparellano, line 75: Is that basically like how we hard-code the lifecycle version here? https://github.com/buildpacks-community/kpack/blob/e1722c0dbbf6d8ff3c20a84aa730a5d7e9ac221e/hack/lifecycle/main.go#L35

Would the future world deprecate that part of the code entirely? Like, if no `ClusterLifecycle` is specified do you get a default `ClusterLifecycle` or do you fallback to pulling the layer out of the `lifecycleImage`?

tomkennedy513: ya this would get deprecated and we would pull in an upstream lifecycle in the ci pipeline instead of building our own like this file is doing. 

we would still ship a default ClusterLifecycle to ease burden on our users and avoid a breaking change on upgrade

natalieparellano (approved):

samj1912:
What happens when there are multiple lifecycle crds? Can we specify a default? Is this a cluster only resource? What happens when the lifecycle gets updated? Will we do a rebuild/rebase?

rfcs/0000-lifecycle-crd.md
chenbh: Do we want to support multi-arch lifecycle images? I.e. if I have an Image Index with linux/amd64, linux/arm64, and windows/amd64 images, do I need to break them out into 3 separate ClusterLifecycles?

natalieparellano:
> What happens when there are multiple lifecycle crds? Can we specify a default?

I think @tomkennedy513 was suggesting we use a webhook for that

> Is this a cluster only resource?

Yes (this is detailed [here](https://github.com/buildpacks-community/kpack/pull/1483/files#diff-56876ddb2609355e5cb3ef2904a19346fbe40f320916d77badf51bc2cc8beb69R41-R44))

> What happens when the lifecycle gets updated? Will we do a rebuild/rebase?

Good question. For rebuild, I think yes? This is philosophically the same as updating your builder which already triggers a rebuild. For rebase, I think no? There would be no benefit if the run image didn't change. @tomkennedy513 @chenbh could you confirm my understanding?

> For rebase, I think no?

BUT we may want to explore (on the CNB-spec side) "launcher layer rebase" to update the go version in the launcher which (when outdated) can cause pain.

rfcs/0000-lifecycle-crd.md
natalieparellano: That is a good question. I suspect the simpler approach would be to keep them separate, but I could be wrong. Would appreciate other more knowledgeable folks weighing in here.

tomkennedy513:
> @tomkennedy513 @chenbh I'd like to work on this. Is there an issue for it? Could we create one (pending the approval of this RFC)?

I dont think we have one, are you ok making one

natalieparellano:
Made an issue to track the work on this: https://github.com/buildpacks-community/kpack/issues/1609

rfcs/0000-lifecycle-crd.md
```diff
@@ -0,0 +1,125 @@
...
+    name: sample-sa
+    namespace: sample-namespace
+  image: buildpacksio/lifecycle@sha256:f4ce143ea6bbc6b5be5f4d151aba8214324adb93bbd7e3b1f43cd134ad450bf7
+```
+
+
+The existing lifecycle reconciler can be updated to reconcile this new CRD instead of ConfigMaps.
+
+A new field will be added to the ClusterBuilder/Builder resources allowing the user to provide a reference to a lifecycle CR.
+This field will be optional and will default to the highest semantic version that is supported by the buildpacks in the builder and the platform.
```
natalieparellano, line 63:
```suggestion
This field will be optional and will default to the highest semantic version that is supported by the platform. 
```
Buildpacks are configured by the operator, so I don't think we can know which Buildpack APIs will be required... right? In theory (assuming we think most buildpacks in the wild have upgraded to at least Buildpack API 0.7 - since lifecycle 0.18.x and higher drops support for earlier APIs) it should be the highest semantic version that supports the Platform APIs [here](https://github.com/buildpacks-community/kpack/blob/ae5eb15b8b9632fda50bafc4d2fad0f8c2c8a2a8/pkg/apis/build/v1alpha2/build_pod.go#L1098)...right? In practice, that is going to mean the latest available lifecycle (until we deprecate more Platform APIs).

tomkennedy513: we do validate that any buildpack referenced in the order is using buildpack apis supported by the lifecycle when constructing a builder https://github.com/buildpacks-community/kpack/blob/ae5eb15b8b9632fda50bafc4d2fad0f8c2c8a2a8/pkg/cnb/builder_builder.go#L188-L204

natalieparellano: Ah, I see - but that validation happens after the ClusterLifecycle is defined, no? So in the case that the provided lifecycle doesn't support a buildpack in the builder we just fail, right? (Instead of selecting a different lifecycle?)

tomkennedy513: I think the only check we'd want when the cluster lifecycle is reconciled is the platform api compatibility. The buildpack api compatibility will continue to happen at builder creation

natalieparellano:
FYI the PR implementing the changes proposed here is ready for feedback: https://github.com/buildpacks-community/kpack/pull/1628

So, we should probably circle back to this RFC to see if it is acceptable as-is or if it needs changes.

cc @tomkennedy513 @chenbh @samj1912

PR #1533, opened 2024-02-21
Relocate the run image to the builder repo

tomkennedy513:
- Also make sure the builder is tagged with the name and kind if a tag is not provided

### DIFF
--- a/pkg/cnb/builder_builder.go
+++ b/pkg/cnb/builder_builder.go
@@ -60,7 +60,6 @@ func newBuilderBldr(kpackVersion string) *builderBlder {
 		kpackVersion:    kpackVersion,
 	}
 }
-
 func (bb *builderBlder) AddStack(baseImage v1.Image, clusterStack *buildapi.ClusterStack) error {
 	file, err := baseImage.ConfigFile()
 	if err != nil {
@@ -70,7 +69,6 @@ func (bb *builderBlder) AddStack(baseImage v1.Image, clusterStack *buildapi.Clus
 	bb.os = file.OS
 	bb.baseImage = baseImage
 	bb.stackId = clusterStack.Status.Id
-	bb.runImage = clusterStack.Status.RunImage.Image
 	bb.mixins = clusterStack.Status.Mixins
 	bb.cnbUserId = clusterStack.Status.UserID
 	bb.cnbGroupId = clusterStack.Status.GroupID
@@ -181,6 +179,10 @@ func (bb *builderBlder) WriteableImage() (v1.Image, error) {
 			Buildpacks: buildpacks,
 		},
 	})
+}
+
+func (bb *builderBlder) AddRunImage(runImage string) {
+	bb.runImage = runImage
 }
 
 func (bb *builderBlder) validateBuilder(sortedBuildpacks []DescriptiveBuildpackInfo) error {
--- a/pkg/cnb/create_builder.go
+++ b/pkg/cnb/create_builder.go
@@ -2,14 +2,15 @@ package cnb
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	ggcrv1 "github.com/google/go-containerregistry/pkg/v1"
-	"github.com/pivotal/kpack/pkg/cosign"
 	corev1 "k8s.io/api/core/v1"
 
 	buildapi "github.com/pivotal/kpack/pkg/apis/build/v1alpha2"
 	corev1alpha1 "github.com/pivotal/kpack/pkg/apis/core/v1alpha1"
+	"github.com/pivotal/kpack/pkg/cosign"
 	"github.com/pivotal/kpack/pkg/registry"
 )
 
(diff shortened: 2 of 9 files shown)


pkg/cnb/create_builder.go
```diff
@@ -110,6 +128,20 @@ func (r *RemoteBuilderCreator) CreateBuilder(ctx context.Context, builderKeychai
...
+	}
+
+	// this happens if there is no tag
+	if parsedRef.Identifier() == "latest" {
+		return parsedRef.Context().Tag(fmt.Sprintf("%s-%s", builderKind, builderName)).Name(), nil
+	}
+
+	return parsedRef.Name(), nil
+}
+
```
chenbh, line 144: nit: this function should be tested

pkg/cnb/create_builder.go
```diff
@@ -30,14 +33,29 @@ type RemoteBuilderCreator struct {
...
+	if err != nil {
+		return buildapi.BuilderRecord{}, err
+	}
+	runImage, _, err := r.RegistryClient.Fetch(stackKeychain, clusterStack.Status.RunImage.LatestImage)
 	if err != nil {
 		return buildapi.BuilderRecord{}, err
 	}
 
-	builderBldr := newBuilderBldr(r.KpackVersion)
+	relocatedRunImage, err := r.RegistryClient.Save(builderKeychain, fmt.Sprintf("%s-run-image", resolvedBuilderTag), runImage)
```
matthewmcnew, line 56: Should this error be handled?

tomkennedy513: Yes I completely forgot to do that

pkg/cnb/create_builder.go
```diff
@@ -110,6 +128,20 @@ func (r *RemoteBuilderCreator) CreateBuilder(ctx context.Context, builderKeychai
 	return builder, nil
 }
 
+func ResolveBuilderRef(ref, builderKind, builderName string) (string, error) {
```
matthewmcnew, line 131: It seems that this is called by the consumer of CreateBuilder. Should it be in this package?

pkg/cnb/builder_builder.go
```diff
@@ -54,10 +54,11 @@ type builderBlder struct {
 	additionalLabels  map[string]string
 }
 
-func newBuilderBldr(kpackVersion string) *builderBlder {
+func newBuilderBldr(kpackVersion string, runImage string) *builderBlder {
```
matthewmcnew, line 57: Should this remain in AddStack or a `RunImage` method? I think we were trying to keep `BuilderBldr` expressive.

pkg/cnb/create_builder.go
tomkennedy513: I mainly put it here so I wouldn't need to duplicate the logic in both reconcilers

pkg/cnb/builder_builder.go
```diff
@@ -70,7 +71,6 @@ func (bb *builderBlder) AddStack(baseImage v1.Image, clusterStack *buildapi.Clus
 	bb.os = file.OS
 	bb.baseImage = baseImage
 	bb.stackId = clusterStack.Status.Id
-	bb.runImage = clusterStack.Status.RunImage.Image
```
matthewmcnew, line 73: Is `clusterStack.Status.RunImage.Image` fully qualified?

pkg/cnb/create_builder.go
matthewmcnew: Then should it be within the RemoteBuilderCreator?

pkg/cnb/builder_builder.go
tomkennedy513: Ya it's the resolved image

pkg/cnb/create_builder.go
tomkennedy513: I'm fine with that

pkg/cnb/builder_builder.go
matthewmcnew: It looks like it just the spec runImage which is not necessarily fully qualified. 

https://github.com/buildpacks-community/kpack/blob/03acfbb54d58af0a006da0c0d94a6d62e53c5735/pkg/cnb/remote_stack_reader.go#L71-L74

matthewmcnew: I think that was intentional. The run image shouldn't necessarily be fully qualified. (E.g if you are using the builder with pack and the run image was updated)

matthewmcnew: I suppose here, it needs to be relocated by kpack so, it is not necessarily useful.

matthewmcnew: Which I guess exposes the downside with this approach. It is only really useful if the source of the stack is not available otherwise, it provides a worse experience.

pkg/cnb/create_builder.go
```diff
@@ -30,14 +33,29 @@ type RemoteBuilderCreator struct {
...
+	spec buildapi.BuilderSpec,
+	serviceAccountSecrets []*corev1.Secret,
+	resolvedBuilderTag string,
+) (buildapi.BuilderRecord, error) {
 
+	buildImage, _, err := r.RegistryClient.Fetch(stackKeychain, clusterStack.Status.BuildImage.LatestImage)
+	if err != nil {
+		return buildapi.BuilderRecord{}, err
+	}
+	runImage, _, err := r.RegistryClient.Fetch(stackKeychain, clusterStack.Status.RunImage.LatestImage)
```
matthewmcnew, line 51: Would it be possible to only save the image if it is in a different registry than the builder?

pkg/cnb/builder_builder.go
tomkennedy513: Ya I guess the trade off is full build reproducibility though which a floating tag doesn't guarantee

pkg/cnb/create_builder.go
tomkennedy513: I was thinking you might still run into a case where the pull rules could be different for a different repo

matthewmcnew: Fair. There is a trade-off from having the run-image be relocated everywhere, though.

tomkennedy513: I think I'm ok with just checking the registry and saying you need to be able to pull both. Maybe putting the original image in the run image mirror would also help as a backup

pkg/cnb/create_builder.go
tomkennedy513: this would require us to pass in the builder name and kind into createBuilder though

pkg/cnb/create_builder.go
matthewmcnew: I was thinking the downside is more about more confusing provenance. It is useful to see that the run image is `my-registry.com/my-stack/specific-stack-name/run` vs `my-registry.com/some-team/custom-builder/run`.

but, I agree it is not a fatal flaw.

tomkennedy513: ya that is definitely a concern, but I was hoping the tags on the relocated run image would simplify this. It also makes registry management a bit easier so you can easily see the run image associated with a particular builder

pkg/cnb/create_builder.go
tomkennedy513: which is fine imo

pkg/cnb/create_builder.go
tomkennedy513: ended up just moving the implementations into each reconciler. I realized I needed to include the namespace in the builder tag so the implementation is slightly different.

pkg/cnb/create_builder.go
tomkennedy513: mentioned the same above, but I ended up just moving the implementations into each reconciler since the ended up looking slightly different. The alternative i was thinking was passing an interface to create_builder that exposes the name, kind, and build spec (or just the specific fields from the build spec).

matthewmcnew (approved):

Issue #1594, opened 2024-04-09
Support Azure Managed Identities for download blob 


zhoufenqin:
Similar to #1503 feature request. 

When using Azure storage account file share to store source code, we should use blob url with sas token, Currently, kpack [doesn't support fetch blob with keychains](https://github.com/buildpacks-community/kpack/blob/ae5eb15b8b9632fda50bafc4d2fad0f8c2c8a2a8/cmd/build-init/main.go#L222), it should be supported.

See the reference about [Azure Files identity-based authentication options for SMB access](https://learn.microsoft.com/en-us/azure/storage/files/storage-files-active-directory-overview)

chenbh:
Talked to @zhoufenqin on another forum, the main thing asked here is the ability to use workload identity (i.e. service account identity) to access Iaas specific blob stores. What we came up with is an additional field to the source configuration

```
source:
  blob:
    url: "blah.blob.azure.com/1234"
    auth: "" | "helper" | "secret"
```

- `""` means no auth performed
- `secret` will do the stuff in #1503 
- `helper` will delegate to IaaS specific SDKs to get an oauth2 token, which it will use as a bearer token for the GET request. 
  For now I'll implement a GCP and Azure helper (purely because those are the only 2 environments I have access to test on). But the cred interface should be simple enough for anyone to implement and contribute.

chenbh:
Would [`azidentity.DefaultAzureCredentials`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#readme-defaultazurecredential) work instead of `azidentity.ManagedIdentityCredential`? It looks it's the recommended approach by the SDK and will fall back to Managed Identity if none of the other options work.

zhoufenqin:
> appr

yes, it works

zhoufenqin:
@chenbh  Do you know which kpack version will release this feature, I didn't see the version tag in PR

chenbh:
@zhoufenqin this feature was released in [v0.14.0](https://github.com/buildpacks-community/kpack/releases/tag/v0.14.0), but you probably want [v0.14.1](https://github.com/buildpacks-community/kpack/releases/tag/v0.14.1) which has the fix in #1637

PR #1616, opened 2024-05-17
Support authentication methods for blobs

### LINKED ISSUE #1503: Download Blob with Basic Auth Secret
Currently, blob downloads need to either be publicly accessible or have the access token in the URL. Basic Authentication can be forced into the URL here, but that results in the username & password being exposed in build logs.

Similar to image pull secrets, it would be ideal to store these credentials in a Kubernetes secret & reference them from the Image CRD.

chenbh:
closes #1503 and #1594 

Introduce a `.spec.source.blob.auth` field with 3 options:
- `""`: no auth performed
- `"secret"`: find all secrets with `kpack.io/blob` annotation (with same domain matching rules as git secret). And use it for basic auth username/password, oauth2 bearer token, or arbitrary `Authorization` header.
- `"helper"`: use IaaS specific IAM mechanisms (i.e. mapping IaaS service accounts/IDs/roles to k8s ServiceAccount) to retrieve an oauth2 token. This PR only implements it for GCP and Azure as those are the 2 envs I have easy access to, but the general interface and registration is simple enough for anyone to contribute.

### DIFF
--- a/pkg/blob/azure_keychain.go
+++ b/pkg/blob/azure_keychain.go
@@ -1,0 +1,50 @@
+package blob
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/sas"
+)
+
+var (
+	azScope      = "https://storage.azure.com/.default"
+	azApiVersion = sas.Version
+	azRegex      = regexp.MustCompile(`.*[a-z0-9]+\.([a-z]+)\.core\.windows\.net\/.*`)
+)
+
+type azKeychain struct{}
+
+func (a azKeychain) Resolve(url string) (string, map[string]string, error) {
+	submatches := azRegex.FindStringSubmatch(url)
+	if len(submatches) != 2 {
+		return "", nil, fmt.Errorf("not an azure url")
+	}
+	service := submatches[1]
+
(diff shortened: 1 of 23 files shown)


pkg/blob/azure_keychain.go
```diff
@@ -0,0 +1,35 @@
...
+
+type azKeychain struct{}
+
+func (a azKeychain) Resolve(url string) (string, map[string]string, error) {
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	if err != nil {
+		return "", nil, err
+	}
+
+	tk, err := cred.GetToken(context.Background(), policy.TokenRequestOptions{Claims: claim})
```
zhoufenqin, line 25: should set Scopes thant than Claims
`policy.TokenRequestOptions{Scopes: []string{xxx}}`

tomkennedy513 (approved):

pkg/blob/file_keychain.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	for _, cred := range f.creds {
+		if u.Hostname() != cred.domain {
+			continue
+		}
+
+		var authHeader []string
+		if cred.username != "" || cred.password != "" {
+			authHeader = append(authHeader, base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("%v:%v", cred.username, cred.password))))
```
yilims, line 98: missing "Basic " here: 
authHeader = append(authHeader, "Basic "+base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("%v:%v", cred.username, cred.password))))

yilims (changes requested):

pkg/blob/file_keychain.go
chenbh: Good catch!

PR #1732, opened 2024-10-14
Update fetch.go to include a Depth specifier

ollie-kane-CB:
For repositories with large histories, since there doesn't appear to be caching of the git repository, we shouldn't pull all history as it's not needed for the purpose of building.  This one line change significantly improves build performance when irrelevant git history is large.

### DIFF
--- a/pkg/git/fetch.go
+++ b/pkg/git/fetch.go
@@ -50,6 +50,7 @@ func (f Fetcher) Fetch(dir, gitURL, gitRevision, metadataDir string) error {
 	err = remote.Fetch(&gogit.FetchOptions{
 		RefSpecs: []config.RefSpec{"refs/*:refs/*"},
 		Auth:     auth,
+		Depth:     1,
 	})
 	if err != nil && err != transport.ErrAuthenticationRequired {
 		return errors.Wrapf(err, "unable to fetch references for repository")
--- a/pkg/git/fetch.go
+++ b/pkg/git/fetch.go
@@ -51,6 +51,7 @@ func (f Fetcher) Fetch(dir, gitURL, gitRevision, metadataDir string) error {
 		RefSpecs: []config.RefSpec{"refs/*:refs/*"},
 		Auth:     auth,
 		Depth:     1,
+		Force:    true,
 	})
 	if err != nil && err != transport.ErrAuthenticationRequired {
 		return errors.Wrapf(err, "unable to fetch references for repository")
--- a/pkg/git/fetch.go
+++ b/pkg/git/fetch.go
@@ -1,6 +1,7 @@
 package git
 
 import (
+	corev1alpha1 "github.com/pivotal/kpack/pkg/apis/core/v1alpha1"
 	"log"
 	"os"
 	"path"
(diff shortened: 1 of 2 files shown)


ollie-kane-CB:
I see the tests have failed.  I'm a bit unfamiliar with Go, but if the effort is needed, i can attempt to figure out what's wrong.  I expected this to be the equiliv of adding the --depth 1 flag to a git clone

```bash
git clone --depth 1 REPO_URL .
```
rather than 

```bash
git clone REPO_URL .
```

ollie-kane-CB:
After reading on the web some, it seems like this link may have something to do with it.

https://github.com/rancher/rancher/issues/42659

They suggest that in addition to the Depth, one must also specify the Force param, otherwise it won't create the local ref.  I'll attempt this and see if it passes the test suite. If not, rather than generate a lot of noise, I'll see if I can determine how to run your test suite myself.

ollie-kane-CB:
Modified this to Draft. I figured out how to build and run the project locally. Will iterate and return.

ollie-kane-CB:
Updated this to pass the unit tests.  Ended up being slightly more complicated than I expected.

In order to support the ```Depth: 1```, the Hash resolver process needed to be updated.  Instead of resolving the gitRevision using the results of a Fetch --all, instead use the GitRemove.List operation to get a list of refs only, and from those resolve the hash. Once the hash is known, use that in the FetchOptions to configure only fetching of the data required. 

A small side effect is the 'doesnotexist' test needed to have it's error message updated as the error thrown when there's no match is different now.

Sorry for the delay and noise here!

pkg/git/fetch.go
```diff
@@ -47,9 +48,23 @@ func (f Fetcher) Fetch(dir, gitURL, gitRevision, metadataDir string) error {
...
+			URL:                  gitURL,
+			Revision:             gitRevision,
+			InitializeSubmodules: f.InitializeSubmodules,
+		},
+		SubPath: "",
+	})
+	if err != nil {
+		return errors.Wrap(err, "resolving source config")
+	}
+
```
ollie-kane-CB, line 63: Use existing git_resolver method to list the remote refs instead of Fetching them all.

pkg/git/fetch.go
```diff
@@ -47,9 +48,23 @@ func (f Fetcher) Fetch(dir, gitURL, gitRevision, metadataDir string) error {
...
 	err = remote.Fetch(&gogit.FetchOptions{
-		RefSpecs: []config.RefSpec{"refs/*:refs/*"},
+		RefSpecs: []config.RefSpec{config.RefSpec(resolvedSourceConfig.Git.Revision + ":" + resolvedSourceConfig.Git.Revision)},
```
ollie-kane-CB, line 65: Configure RefSpecs to not pull EVERYTHING but instead just the specific resolved commit we want.

pkg/git/fetch.go
```diff
@@ -47,9 +48,23 @@ func (f Fetcher) Fetch(dir, gitURL, gitRevision, metadataDir string) error {
...
 		Auth:     auth,
+		Depth:    1,
```
ollie-kane-CB, line 67: Don't get all the history. we just need that single commit.

pkg/git/fetch_test.go
```diff
@@ -79,7 +79,7 @@ func testGitCheckout(t *testing.T, when spec.G, it spec.S) {
 
 		it("returns error on non-existent ref", func() {
 			err := fetcher.Fetch(testDir, "https://github.com/git-fixtures/basic", "doesnotexist", metadataDir)
-			require.EqualError(t, err, "resolving revision: reference not found")
+			require.EqualError(t, err, "unable to fetch references for repository: couldn't find remote ref \"doesnotexist\"")
```
ollie-kane-CB, line 82: Error message has changed. Updated to newly expected error message such that the tests pass.

ollie-kane-CB:
> lgtm

Thanks for the merge!  

I'm looking into adopting this project into my stack.  I don't know if I'll ever get around to it, but if you were to pick a feature/bug/issue that'd be the best value-add, what would that be? If I'm swimming in this lane, it's possible that a weekend here i may take a stab at the backlog.

tomkennedy513:
> > lgtm
> 
> Thanks for the merge!
> 
> I'm looking into adopting this project into my stack. I don't know if I'll ever get around to it, but if you were to pick a feature/bug/issue that'd be the best value-add, what would that be? If I'm swimming in this lane, it's possible that a weekend here i may take a stab at the backlog.

hmm, I'm not sure I have a particular issue I would recommend, but I would say that playing around with kpack and seeing what doesn't work right or what annoys you that could be improved might be a good way to start

PR #1742, opened 2024-10-28
Add `cascadeDelete` option to the `Image` and `Build` objects

pbusko:
When `cascadeDelete` is set to `true` on a `Build` object, the produced image by the build will be removed from the registry upon object deletion. If the `cascadeDelete` is set on the `Image` object, it will be propagated to `Build`s


pkg/reconciler/build/build.go
```diff
@@ -352,6 +370,64 @@ func (c *Reconciler) conditionForPod(pod *corev1.Pod, stepsCompleted []string) c
...
+	if build.Finished() {
+		keychain, err := c.KeychainFactory.KeychainForSecretRef(ctx, registry.SecretRef{
+			ServiceAccount: build.Spec.ServiceAccountName,
+			Namespace:      build.Namespace,
+		})
+		if err != nil {
+			return err
+		}
+
+		if err := c.RegistryClient.Delete(keychain, build.Status.LatestImage); err != nil {
```
sambhav, line 387: We would also need to delete all of the additional tags as well.

pkg/reconciler/build/build.go
```diff
@@ -352,6 +370,64 @@ func (c *Reconciler) conditionForPod(pod *corev1.Pod, stepsCompleted []string) c
...
+			return err
```
tomkennedy513, line 388: Returning error might mean that builds get stuck if the finalized can't delete the image due to immutability rules in the registry. I think this should be a best effort cleanup

pkg/reconciler/build/build.go
```diff
@@ -352,6 +370,64 @@ func (c *Reconciler) conditionForPod(pod *corev1.Pod, stepsCompleted []string) c
...
+		keychain, err := c.KeychainFactory.KeychainForSecretRef(ctx, registry.SecretRef{
+			ServiceAccount: build.Spec.ServiceAccountName,
+			Namespace:      build.Namespace,
+		})
+		if err != nil {
+			return err
+		}
+
+		if err := c.RegistryClient.Delete(keychain, build.Status.LatestImage); err != nil {
+			//logger.Printf(errors.Wrapf(err, "Could not delete image %q", build.Status.LatestImage))
```
nicolasbender, line 388: How should we handle the error in that case? It seems like error logging is not supposed to happen at the reconciliation stage.

tomkennedy513: You can log and it will print to the controller logs

pkg/reconciler/build/build.go
tomkennedy513: A build can push multiple images (see additionalTags). There is also the signed images and image attestations but that might be too much to try to delete and we probably don’t want to delete the attestations

modulo11: According to [docs](https://github.com/buildpacks-community/kpack/blob/main/docs/build.md#status), the `status.latestImage` field contain the image's digest. It should be enough to delete just this (no necessity to delete all tags individually): https://github.com/opencontainers/distribution-spec/blob/main/spec.md#deleting-manifests.

> Once deleted, a GET to /v2/<name>/manifests/<digest> and any tag pointing to that digest will return a 404.

pkg/reconciler/build/build.go
pbusko: Isn't it fair to keep the build if the desired outcome is not reached? The solution is not to enable the cleanup when the registry forbids objects deletion is used. If I set the cleanup to `true` but the image is still present, it should be considered an error, not a warning message

PR #1767, opened 2024-12-06
Enable using cache registry for the first build

mtparet:
Registry caching was disabled for the first build although it permits to share cache between images. This will speed lot of first builds.


mtparet:
Ho @chenbh, any chance to get a review ? Thanks 🙏

chenbh (approved):

chenbh:
@mtparet it looks like GitHub Actions didn't pick up your commit for whatever reason, can you try regenerating the commit hash and force pushing?
```
git commit --amend --no-edit
git push -f
```

chenbh:
I swear that button wasn't there last time 😅. 

@mtparet One of the tests is failing

mtparet:
@chenbh I believe you :D 

I fixed the test, the way we select the container based on its position in the initContainers array is not very reliable and we were not testing the right container. I added a function to select the container based on the name, I can open a future PR to convert all existing tests so we ensure we tests right containers everywhere.

mtparet:
(I need an approval on the workflow to trigger the test again)

mtparet:
Thanks @tomkennedy513 !

PR #1820, opened 2025-01-22
Do not reconcile images and builds being deleted

danail-branekov:
When an image is being deleted, k8s does not delete the object immediately, instead it sets the deletion timestamp awaiting finalization. See
[k8s docs](https://github.com/kubernetes/apimachinery/blob/45d29dc4d66fc2ac83e736e79752ad81a9c6195f/pkg/apis/meta/v1/types.go#L190-L209) for details

If the kpack image (or its builds) has a finalizer being added by an external component (such as
[Korifi](https://github.com/cloudfoundry/korifi/blob/17557eb68fed830f3f57abd651882a712fc25f5f/kpack-image-builder/controllers/webhooks/finalizer/finalizer_webhook.go#L25)), then when the client request the object to be deleted, the image/build reconcilers keep reconciling the image/build causing new build pods to be created as the image is being deleted. Even though eventually the image gets deleted, it takes significant amount of time.

In order to address this, this PR "short-circuits" the image and build reconcilers to immediately return whenever an image or a build has their deletion timestamp set, thus preventing builds and build pods being rescheduled.

### DIFF
--- a/pkg/reconciler/build/build.go
+++ b/pkg/reconciler/build/build.go
@@ -136,6 +136,10 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 		return err
 	}
 
+	if build.DeletionTimestamp != nil {
+		return nil
+	}
+
 	build = build.DeepCopy()
 	build.SetDefaults(ctx)
 
--- a/pkg/reconciler/build/build.go
+++ b/pkg/reconciler/build/build.go
@@ -136,10 +136,6 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 		return err
 	}
 
-	if build.DeletionTimestamp != nil {
-		return nil
-	}
-
 	build = build.DeepCopy()
 	build.SetDefaults(ctx)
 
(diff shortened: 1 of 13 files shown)


chenbh (commented):
It feels kinda awkward to have this short circuit on only the Build and Image, I wonder if we should just add this to all our reconcilers.

I imagine we can make use of the `FilterFunc` to skip any resource with non-nil DeletionTimestamps https://github.com/buildpacks-community/kpack/blob/d3e70fe1cbb0829025e65748a8c2e43cf69fa5c8/pkg/reconciler/image/image.go#L68-L71

Something like

```go
pkg/reconciler/filter.go:
func FilterDeletionTimestamp(nextFilter func({}interface) bool) func(obj interface{}) bool {
	return func(obj interface{}) bool {
		object, ok := obj.(metav1.Object)
		if !ok {
			return false
		}

		if object.GetDeletionTimestamp() != nil {
			return false
		}
                return nextFilter(obj)
	}
}

And in each reconciler:
buildInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{ 
 	FilterFunc: reconciler.FilterDeletionTimestamp(
            controller.FilterControllerGK(buildapi.SchemeGroupVersion.WithKind(Kind).GroupKind()),
        ), 
 	Handler:    controller.HandleAll(impl.EnqueueControllerOf), 
 }) 
```

This way we can also get away without modifying all the reconciler tests

chenbh (approved):
LGTM

PR #1833, opened 2025-02-26
Do not attempt to download lifecycle binary for Windows

natalieparellano:
As of lifecycle 0.20.6, Windows support is removed, and no lifecycle binary is available for Windows on GitHub.

Without this change, we'd be unable to upgrade the lifecycle version in kpack beyond 0.20.5.

This change avoids attempting to download the Windows binary, but still produces an empty Windows layer in the lifecycle image, to minimize the surface area of changes.

This is a first step toward deprecating Windows support in kpack (see https://github.com/buildpacks-community/kpack/discussions/1366), which requires further changes throughout the code base to remove Windows-specific logic entirely.

### DIFF
--- a/hack/lifecycle/main.go
+++ b/hack/lifecycle/main.go
@@ -11,7 +11,6 @@ import (
 	"net/http"
 	"os"
 	"path"
-	"reflect"
 	"regexp"
 	"time"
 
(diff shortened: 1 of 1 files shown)


natalieparellano:
I verified that `go run ./hack/lifecycle/main.go` creates a pullable image. Below is the Windows layer that gets created - I think this is what we want. 

Screenshot 2025-02-26 at 2 19 27 PM

tomkennedy513 (approved):

Issue #1836, opened 2025-03-04
GCR shutdown on March 18, 2025 – impact on kpack stored images


tgrzesik:
Hi team,

I wanted to bring to your attention that Google Container Registry (GCR) will be permanently shut down on March 18, 2025.

It looks like kpack is currently storing images in GCR. Given the upcoming shutdown, it might be worth considering migrating to Google Artifact Registry or another supported registry to avoid disruptions.

Thanks! 😊

PR #1853, opened 2025-04-15
Migrate images to GHCR

### LINKED ISSUE #1385: Publish release images to GHCR
We're currently publishing to Google Container Registry under a VMware account because of legacy reasons. We should migrate to GitHub Container Registry

chenbh:
Closes #1385, #1836

The current GAR repository the kpack images are published to is owned by Broadcom. We've been informed that this GCP project will be shut down as per Broadcom policy that all GCP projects needs to be internal and behind a firewall.

So this PR moves the destination for both dev (aka nightly) and release images to ghcr.io

There's also a small change in here to use a mirror for the lifecycle image as we were getting throttled by index.docker.io

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -49,9 +49,9 @@ jobs:
     - name: Docker Login
       uses: docker/login-action@v3.3.0
       with:
-        registry: ${{ secrets.REGISTRY_HOST }}
-        username: ${{ secrets.REGISTRY_USER }}
-        password: ${{ secrets.REGISTRY_PASSWORD }}
+        registry: ghcr.io
+        username: ${{ github.actor }}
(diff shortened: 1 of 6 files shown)


tomkennedy513 (approved):

chenbh:
@sambhav this probably means that older kpack releases will be broken as the release.yml will be pointing to a defunct gcr.io repository. Lmk if you have any concerns with this from either a consumer or CNB perspective, I do plan on cutting a new patch release asap to ensure at least the latest version works

sambhav:
We should make an announcement somewhere?

tomkennedy513:
> We should make an announcement somewhere?

Maybe Slack?

tomkennedy513:
I'm going to merge this now so that we can cut another release before the other registry is deleted. I will follow up with a slack message.

PR #1936, opened 2026-02-05
Implement subPath filtering

### LINKED ISSUE #1068: Support sub folder reconciliation logic

mogsie:
This replaces the "git ls-remote" technique of discovering revisions with a bare, shallow clone, which is filtered for blobs.  Only metadata is downloaded.

* The Status of the SourceResolver includes the "tree" of the detected source
* The tree is only filled out if a subpath is requested
* When the SourceResolver status is being updated, the "tree" is compared between the old status and the new status.  If the tree didn't change, then the old status' commit is kept.

The resolver does the equivalent of

  git clone --bare --depth=1 --filter=blob:none <repo> <ref>

This downloads only the metadata for the desired reference

If subPath is filled out, it performs the equivalent of

  git ls-tree <ref> -- <subpath>

which provides a unique hash of the subpath at the commit in question.

fixes #1068


mogsie:
I had to upgrade go git from v5 to v6, and I have no problem if we want to do that in a separate pull-request.  I mainly wanted to prove that this is possible, and it does seem to do the trick.

(Converted to draft because I forgot the actual _status update_ code.)

pkg/git/fetch.go
```diff
@@ -23,9 +23,9 @@ type Fetcher struct {
 
 func init() {
 	//remove multi_ack and multi_ack_detailed from unsupported capabilities to enable Azure DevOps git support
-	transport.UnsupportedCapabilities = []capability.Capability{
-		capability.ThinPack,
-	}
+	//	transport.UnsupportedCapabilities = []capability.Capability{
```
mogsie: I need to find out what happened to this API, and its replacement in go-git v6

tomkennedy513:
Hey @mogsie what are you thoughts around putting this new behavior behind a feature flag on the controller

mogsie:
> Hey @mogsie what are you thoughts around putting this new behavior behind a feature flag on the controller

I'd be happy to.  The "tree" in the source resolver spec would have to be part of the API of course.  Where can I learn how kpack provides feature flags?

Should I split the upgrade to go-git/v6 into its own PR?

mogsie:
Ok, I understood it as being adding a variable to config.FeatureFlags, CLI parsing and environment variable support, pass it on to the Reconciler construction and store it in the struct.  Sending this feature flag would require quite a few changes to signatures and structs, but I guess that's the way it is.

tomkennedy513:
> > Hey @mogsie what are you thoughts around putting this new behavior behind a feature flag on the controller
> 
> 
> 
> I'd be happy to.  The "tree" in the source resolver spec would have to be part of the API of course.  Where can I learn how kpack provides feature flags?
> 
> 
> 
> Should I split the upgrade to go-git/v6 into its own PR?

Ya, normally we handled this with the flags that get parsed during controller startup which would provide as env vars. It is a bit cumbersome but it gives us an escape hatch for potentially large behavior changes. 

For the go git stuff, I don't have a super strong opinion about the upgrade happening in a separate pr, whichever is easier for you

mogsie:
I split it into three commits, for the benefit of review.

* Upgrade v5-v6
* Forward featureflags down to the reconciler and resolvers
* Added the new code behind feature flags.

The test suite for the remote git resolver is run twice; for both variants of the feature flag.

The only difference is that the "tree" is filled out when the feature flag is enabled.

pkg/git/fetch.go
tomkennedy513: Were you able to find anything about this?

tomkennedy513:
Hey @mogsie, I was reading the migration guide for go-git v5 to v6 (https://go-git.github.io/docs/tutorials/migrating-from-v5-to-v6/) and it said that v6 isn't officially released yet. When checking the repo i also don't see any v6 releases yet. is there anything specific your impl needs from v6? I am a little concerned about bumping before the api is fully ironed out.

mogsie:
Yes, initially, I wasn't aware that v6 wasn't released, and it was on my list of things to check out.  I agree that this might be a bit premature.  The main feature is the blob filtering, which ensures that the clone only contains metadata about the commits, and no actual files, which is (likely) critical in avoiding the memory issues that plagued earlier attempts.

I'll look again to see what's possible in v5.

mogsie:
I will move this back to go-git v5 with a shallow bare clone instead, seeing as it'd be behind a feature flag.  Then we can revisit the blob filtering once go-git v6 is stable.  :+1:

mogsie:
Summary of changes:

* Removed the upgrade to go-git v6 commit
* removed the now nonexistent _Bare: true_ and _Filter: packp.FilterBlobNone()` options.

Since the worktree is set to nil, it should be performing a bare clone-

pkg/git/fetch.go
mogsie: No, I never got round to it. I was going to look through the v5-v6 migration guide, but that's moot now.

pkg/reconciler/sourceresolver/sourceresolver.go
```diff
@@ -138,10 +138,28 @@ func (c *Reconciler) updateStatus(ctx context.Context, desired *buildapi.SourceR
...
 	_, err = c.Client.KpackV1alpha2().SourceResolvers(desired.Namespace).UpdateStatus(ctx, desired, v1.UpdateOptions{})
 	return err
 }
+
+func (*Reconciler) retainCommitIfTreeUnchanged(original buildapi.SourceResolverStatus, desired *buildapi.SourceResolverStatus) {
+	if original.Source.Git == nil || desired.Source.Git == nil {
+		return
+	}
+
+	if original.Source.Git.Tree == "" || desired.Source.Git.Tree == "" {
```
tomkennedy513, line 158: wouldn't this prevent the tree from being added if there wasn't one before?

tomkennedy513: In my test it didn't look like an issue so I may be wrong. I'm guessing the sub resolver got replaced when I added the subpath but I didn't go validate if thats true

mogsie: The func in question is about *rollling back* a change to the `revision` if the _tree_ was unchanged.

So the first time, yes, this func will quit here and it will be a no-op.  The desired status _already has a tree_.

The second time we get here (next reconcile) both the statuses will have the tree, possibly different.

Maybe it should have been called `retainRevisionIfTreeIsUnchanged` though?

mogsie:
Just pointing out that using this on extremely large repositories (which unfortunately is its prime use case — monorepos) is not ideal... Maybe this should be mentioned in the (to be written) docs for this feature flag?

A bare, shallow git clone of the linux repository is 250MiB, while if you filter away the blobs, the metadata clocks in at ~1% of that, 2½MiB.

pkg/reconciler/sourceresolver/sourceresolver.go
tomkennedy513: ok ya I see that now, I don't have a strong opinion on the name.

tomkennedy513:
> Just pointing out that using this on extremely large repositories (which unfortunately is its prime use case — monorepos) is not ideal... Maybe this should be mentioned in the (to be written) docs for this feature flag?
> 
> A bare, shallow git clone of the linux repository is 250MiB, while if you filter away the blobs, the metadata clocks in at ~1% of that, 2½MiB.

That's unfortunate. I was thinking earlier that we could potentially use go-git v6 for the cloning codepath and leave v5 for the existing code path. Obvious downside is that we would have 2 dependencies so for go-git so images would be larger, but it might be a worthwhile tradeoff. What are your thoughts on that

mogsie:
Huh, I never considered that. I'll give it a shot and provide the size of the resulting binaries.

tomkennedy513:
> Huh, I never considered that. I'll give it a shot and provide the size of the resulting binaries.

I personally haven't tried something like this before but think it might work because they are different majors. We still have that azure devops workaround to deal with on the v6 side, but as far as the rest of the pr is concerned I think everything looks good.

mogsie:
`pack build` with `paketobuildpacks/builder-jammy-base:latest`

| commit | size of container | size in bytes of controller | comment |
|--|--|--|--|
| bf0baf9 | 667,652,700 | 150,614,712 | git v5 |
| ea3c2d7 | 672,267,091 | 152,615,375 | mix of git v5 and v6 |

So, 4,614,391 bytes bigger.

tomkennedy513 (approved):
lgtm. Thank you @mogsie

mogsie:
@tomkennedy513 should I make a separate PR to document this feature flag?

tomkennedy513:
> @tomkennedy513 should I make a separate PR to document this feature flag?

Yes I completely forgot about that when I merged. If you wouldn't mind that would be great

PR #1939, opened 2026-02-10
Update codegen

tomkennedy513:
This is needed after bumping k8s.io/code-generator to v0.34.3 in #1931


chenbh (approved):
nit: can you update the commit message to indicate why this was needed? Had to do a little bit of digging to realize it's probably something to do with us bumping k8s.io/code-generator in https://github.com/buildpacks-community/kpack/pull/1931

PR #1941, opened 2026-03-09
registry client use currentPlatform

tpaul1611:
As discussed in https://github.com/buildpacks-community/kpack/issues/1940 even if custom built for arm64, kpack controller is still hardwired to pull amd64 only. This PR rectifies this by always using the current arch, effectively solving the problem when building for arm64, while not changing anything for amd64.

### DIFF
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -3,6 +3,7 @@ package registry
 import (
 	"fmt"
 	"net/http"
+	"runtime"
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/authn"
@@ -14,6 +15,11 @@ import (
 	"github.com/pkg/errors"
 )
 
+var currentPlatform = remote.WithPlatform(v1.Platform{
+	Architecture: runtime.GOARCH,
+	OS:           "linux",
+})
+
 type Client struct {
 }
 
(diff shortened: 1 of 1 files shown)


tomkennedy513:
Sorry for the delay. What will happen if requesting an architecture that doesn't exist? Also maybe we can add some tests for this new behavior

tms-at-celum:
> Sorry for the delay. What will happen if requesting an architecture that doesn't exist? Also maybe we can add some tests for this new behavior

It will fail :) For me this is a low level change, that lets ppl who build their own images for their own choice of arch do their thing :) For a complete multi arch concept for kpack, alot more thought and changes have to be done.

tomkennedy513:
> > Sorry for the delay. What will happen if requesting an architecture that doesn't exist? Also maybe we can add some tests for this new behavior
> 
> 
> 
> It will fail :) For me this is a low level change, that lets ppl who build their own images for their own choice of arch do their thing :) For a complete multi arch concept for kpack, alot more thought and changes have to be done.

I think that's ok but we should really have tests for to is then to make sure we don't break existing behavior

tomkennedy513:
closing in favor of #1951 -> this work was cherry picked onto that branch and tests were added. Thank you @tpaul1611 (and @BaptisteCentreon for adding the tests)

PR #1944, opened 2026-04-28
reimplement subpath filtering experiment

mogsie:
The error was incorrectly being ignored, and resulted in wrong error messages about branches not found.

The subpath filtering experiment now
* Is only turned on if `subPath` is set
* creates an in-memory bare repository
* tries to fetch the remote reference as a branch, a tag, and a commit individually, every time without fetching any blobs.

Includes a document describing it in some detail.

### DIFF
--- a/pkg/git/remote_git_resolver_clone.go
+++ b/pkg/git/remote_git_resolver_clone.go
@@ -31,6 +31,18 @@ func (r *remoteGitResolver) ResolveByCloning(auth transport.AuthMethod, sourceCo
 		Filter:        packp.FilterBlobNone(),
 	})
 
+	if err != nil {
+		return corev1alpha1.ResolvedSourceConfig{
+			Git: &corev1alpha1.ResolvedGitSource{
+				URL:                  sourceConfig.Git.URL,
+				Revision:             sourceConfig.Git.Revision,
+				Type:                 corev1alpha1.Unknown,
+				SubPath:              sourceConfig.SubPath,
+				InitializeSubmodules: sourceConfig.Git.InitializeSubmodules,
+			},
+		}, fmt.Errorf("cloning repository: %w", err)
+	}
+
 	var resolvedRef *plumbing.Reference
 	var hash plumbing.Hash
 	var kind corev1alpha1.GitSourceKind
--- a/pkg/git/remote_git_resolver_clone.go
+++ b/pkg/git/remote_git_resolver_clone.go
@@ -44,19 +44,22 @@ func (r *remoteGitResolver) ResolveByCloning(auth transport.AuthMethod, sourceCo
 		return emptyResult, fmt.Errorf("creating remote: %w", err)
 	}
 
-	var resolvedRef *plumbing.Reference
-	var hash plumbing.Hash
-	var kind corev1alpha1.GitSourceKind
+	input := remoteGitCloneResolverInput{
+		repository: repository,
+		auth:       auth,
+		revision:   sourceConfig.Git.Revision,
+	}
+	var output *remoteGitCloneResolverOutput
 	errs := []error{}
 
 	for _, resolver := range resolvers {
-		resolvedRef, hash, kind, err = resolver(repository, auth, sourceConfig.Git.Revision)
+		output, err = resolver(input)
 		if err != nil {
 			errs = append(errs, err)
 			fmt.Printf("resolver %T failed: %v\n", resolver, err)
 			continue
 		}
-		if resolvedRef != nil {
+		if output != nil {
 			errs = nil
 			break
 		}
@@ -79,9 +82,9 @@ func (r *remoteGitResolver) ResolveByCloning(auth transport.AuthMethod, sourceCo
 	return corev1alpha1.ResolvedSourceConfig{
 		Git: &corev1alpha1.ResolvedGitSource{
 			URL:                  sourceConfig.Git.URL,
-			Revision:             hash.String(),
-			Tree:                 hashOfSubpath(sourceConfig.SubPath, hash, repository),
-			Type:                 kind,
+			Revision:             output.hash.String(),
+			Tree:                 hashOfSubpath(sourceConfig.SubPath, output.hash, repository),
+			Type:                 output.kind,
 			SubPath:              sourceConfig.SubPath,
 			InitializeSubmodules: sourceConfig.Git.InitializeSubmodules,
 		},
@@ -113,18 +116,29 @@ func hashOfSubpath(subPath string, hash plumbing.Hash, repository *gogit.Reposit
 
 }
 
-type resolverFunc func(repository *gogit.Repository, auth transport.AuthMethod, revision string) (*plumbing.Reference, plumbing.Hash, corev1alpha1.GitSourceKind, error)
+type remoteGitCloneResolverInput struct {
+	repository *gogit.Repository
+	auth       transport.AuthMethod
+	revision   string
+}
+type remoteGitCloneResolverOutput struct {
+	reference *plumbing.Reference
+	hash      plumbing.Hash
+	kind      corev1alpha1.GitSourceKind
+}
+
+type resolverFunc func(remoteGitCloneResolverInput) (*remoteGitCloneResolverOutput, error)
 
 var resolvers = []resolverFunc{resolveBranch, resolveTag, resolveRevision, looksLikeACommit}
 
-func resolveBranch(repository *gogit.Repository, auth transport.AuthMethod, branch string) (*plumbing.Reference, plumbing.Hash, corev1alpha1.GitSourceKind, error) {
+func resolveBranch(input remoteGitCloneResolverInput) (*remoteGitCloneResolverOutput, error) {
+	branch := input.revision
 	if !strings.HasPrefix(branch, "refs/heads/") {
 		branch = "refs/heads/" + branch
 	}
-	fmt.Println("fetching branch", branch)
-	err := repository.Fetch(&gogit.FetchOptions{
+	err := input.repository.Fetch(&gogit.FetchOptions{
 		RemoteName: defaultRemote,
-		Auth:       auth,
+		Auth:       input.auth,
 		RefSpecs: []gogitconfig.RefSpec{
 			gogitconfig.RefSpec(branch + ":" + branch),
 		},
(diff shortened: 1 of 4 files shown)


mogsie:
I have this code running in our clusters, with the experiment turned on. When enabling the experiment, the clone step failed with being unable to clone because of missing _known hosts_ file.  The error message then said to add a known hosts file environment variable SSH_KNOWN_HOSTS. Adding the `known_hosts` to the git secret did not help; adding the environment variable + an empty file to the kpack process did.

I'm not sure if this is go-git v6 https://github.com/go-git/go-git/issues/1234, or something else related to the fact that we are initializing `auth` with go-git v5 and using it with v6 in this branch of the code. I worked around it by providing an EmptyDir with kpack and an empty known_hosts file in it.

mogsie:
@tomkennedy513 since you reviewed #1936 I was wondering if you could provide some guidance on how to resolve this conundrum.

(I have some clean-up to do before marking the PR as ready for review.)

The current tests for _tags_ fails because of the situation when a Build references an annotated tag, the SourceResolver (currently) resolves to the sha of the annotated tag.  My changes would change this so that it resolves to the _commit sha_ that the annotated tag points to, as evidenced by the test message:

``` diff
--- Expected
+++ Actual
@@ -3,5 +3,5 @@
   URL: (string) (len=40) "https://github.com/git-fixtures/tags.git",
-  Revision: (string) (len=40) "ad7897c0fb8e7d9a9ba41fa66072cf06095a6cfc",
+  Revision: (string) (len=40) "f7b877701fbf855b44c0a9e86f3fdce2c298b07f",
   SubPath: (string) (len=5) "/tree",
-  Tree: (string) "",
+  Tree: (string) (len=40) "e69de29bb2d1d6434b8b29ae775ad8c2e48c5391",
   Type: (v1alpha1.GitSourceKind) (len=3) "Tag",
```

Test:  returns_tag_with_resolved_commit

* ad7897c is the annotated tag from the fixture.
* f7b8777 is the commit that the tag points to.

The point here being that if I change my Image spec to build _a different tag_ the "subpath filtering" would *skip* building the image if the subpaths of the two tags contained the same code.

Is it important to check out annotated tags?  If so, subpath filtering won't work for annotated tags, and the way I see it, you could get that behaviour just by not including the subpath.

I *think* that the correct behaviour here is (when subpath filtering is active) to let the buildpack check out the annotated tag (since it ends up in image metadata), but that it ends up not being rebuilt if (when you swith to a new tag) the subpath didn't change.

Thanks in advance for your guidance.

tomkennedy513:
> @tomkennedy513 since you reviewed #1936 I was wondering if you could provide some guidance on how to resolve this conundrum.
> 
> 
> 
> (I have some clean-up to do before marking the PR as ready for review.)
> 
> 
> 
> The current tests for _tags_ fails because of the situation when a Build references an annotated tag, the SourceResolver (currently) resolves to the sha of the annotated tag.  My changes would change this so that it resolves to the _commit sha_ that the annotated tag points to, as evidenced by the test message:
> 
> 
> 
> ``` diff
> 
> --- Expected
> 
> +++ Actual
> 
> @@ -3,5 +3,5 @@
> 
>    URL: (string) (len=40) "https://github.com/git-fixtures/tags.git",
> 
> -  Revision: (string) (len=40) "ad7897c0fb8e7d9a9ba41fa66072cf06095a6cfc",
> 
> +  Revision: (string) (len=40) "f7b877701fbf855b44c0a9e86f3fdce2c298b07f",
> 
>    SubPath: (string) (len=5) "/tree",
> 
> -  Tree: (string) "",
> 
> +  Tree: (string) (len=40) "e69de29bb2d1d6434b8b29ae775ad8c2e48c5391",
> 
>    Type: (v1alpha1.GitSourceKind) (len=3) "Tag",
> 
> ```
> 
> 
> 
> Test:  returns_tag_with_resolved_commit
> 
> 
> 
> * ad7897c is the annotated tag from the fixture.
> 
> * f7b8777 is the commit that the tag points to.
> 
> 
> 
> The point here being that if I change my Image spec to build _a different tag_ the "subpath filtering" would *skip* building the image if the subpaths of the two tags contained the same code.
> 
> 
> 
> Is it important to check out annotated tags?  If so, subpath filtering won't work for annotated tags, and the way I see it, you could get that behaviour just by not including the subpath.
> 
> 
> 
> I *think* that the correct behaviour here is (when subpath filtering is active) to let the buildpack check out the annotated tag (since it ends up in image metadata), but that it ends up not being rebuilt if (when you swith to a new tag) the subpath didn't change.
> 
> 
> 
> Thanks in advance for your guidance.

Your proposal sounds like a valid option provided we document this behavior. We could also potentially add a status condition saying build skipped or something of that nature

pkg/git/remote_git_resolver_clone.go
```diff
@@ -95,50 +121,125 @@ func hashOfSubpath(subPath string, hash plumbing.Hash, repository *gogit.Reposit
...
+		return nil, fmt.Errorf("fetching: %w", err)
 	}
 
-	return plumbing.NewHashReference(plumbing.ReferenceName(revision), h), h, corev1alpha1.Commit, nil
+	reference, err := input.repository.CommitObject(plumbing.NewHash(commitHash))
+	if err != nil {
+		return nil, fmt.Errorf("tag: %w", err)
+	}
+
+	return &fetchResolverOutput{
```
mogsie, line 230: code coverage of the tests seem to indicate that we never get here.  The fetch fails with `fetching: decoding shallow-update: upload-pack: not our ref a755256fc0a57241b92167eb748b333449a3d7e9`

I can't get gogit to play along to fetch a raw commit.

The only reason to get this to work is that _if we can fetch the commit metadata_ here, then the subpath filtering will work in these cases too.

Unfixed, it means that subpath filtering won't work if you use commit shas as the source revision.

mogsie:
I documented it in a new markdown file, and the subpath filtering should now work with private repos and ssh logins etc.  I have this running at $WORK and it is humming along nicely.

pkg/git/remote_git_resolver_clone.go
```diff
@@ -95,50 +121,125 @@ func hashOfSubpath(subPath string, hash plumbing.Hash, repository *gogit.Reposit
...
+type fetchResolverOutput struct {
+	reference *plumbing.Reference
+	hash      plumbing.Hash
+	kind      corev1alpha1.GitSourceKind
+}
+
+type fetchResolver func(fetchResolverInput) (*fetchResolverOutput, error)
 
-var resolvers = []resolverFunc{resolveBranch, resolveTag, resolveRevision, looksLikeACommit}
+var fetchResolvers = []fetchResolver{resolveBranch, resolveTag, resolveRevision, looksLikeACommit}
```
mogsie: One optimization one could do is to look at the revision being resolved: If it starts with "refs/heads/" we could do _resolveBranch_ only.  If it starts with "refs/tags/" we could use resolveTag only.  If it looks like a commit, we could resolve revisions.

Also, I wonder if it should support arbitrary remote references, i.e. if it starts with "refs/", e.g. pull-request references (refs/pull/837/head), which aren't even branches, but should be fetchable.

mogsie: I implemented this (refs/pull/), trying to keep some parity with the existing implementation.  But I think maybe "refs/pull" and so on aren't *really* supported by kpack Images?  I can revert this if this should go through a more formal feature proposal process.

tomkennedy513: I think one issue with pull is going to be the ephemerality of where you will eventually have the image start failing when the pr is merged. I'm somewhat indifferent to the feature existing assuming that it won't somehow break tag/branch/commit resolution

mogsie: Yes, the use case for this is of course in CI, to prove that the kpack build will work when a PR is merged. Something like ArgoCD (ApplicationSet) could be creating the Image object with the refs/pull/123/head as the reference, with kpack rebuilding as commits are pushed to it.  There are other ways of doing this, of course (ArgoCD could use the commit sha), so it is perhaps a non-issue.  I can remove it.

tomkennedy513:
@mogsie is this is a good state now for me to review?

mogsie:
Yes.  It should all be OK to review @tomkennedy513.

pkg/git/remote_git_resolver_clone.go
```diff
@@ -95,50 +121,141 @@ func hashOfSubpath(subPath string, hash plumbing.Hash, repository *gogit.Reposit
...
+		Depth:  1,
+		Filter: packp.FilterBlobNone(),
+	})
 	if err != nil {
-		return nil, plumbing.Hash{}, corev1alpha1.Unknown, err
+		return nil, fmt.Errorf("fetching: %w", err)
 	}
 
-	return resolvedTag, resolvedTag.Hash(), corev1alpha1.Tag, nil
+	reference, err := input.repository.Tag(tag)
```
tomkennedy513, line 198: Tag function adds refs/tags to the front (https://github.com/go-git/go-git/blob/bc4de5eb6836294af1b31bd44886dc8cb67caf4b/repository.go#L1030) so you would probably need to trim the tag here

pkg/git/remote_git_resolver_test.go
```diff
@@ -100,15 +101,20 @@ func testRemoteGitResolver(t *testing.T, when spec.G, it spec.S) {
...
 					Git: &corev1alpha1.ResolvedGitSource{
 						URL:      tagsUrl,
 						Revision: tagCommit,
 						Type:     corev1alpha1.Tag,
 						SubPath:  "/tree",
 					},
-				}, resolvedGitSource)
+				}
+
+				if featureFlags.GitResolverUseShallowClone {
```
tomkennedy513, line 113: couldnt we have the same test, but with the feature flag set so we always test both cases? maybe a table test or something

pkg/git/remote_git_resolver_clone.go
```diff
@@ -20,33 +21,51 @@ const commitSHARegex = `^[a-f0-9]{40}$`
...
+			errs = nil
 			break
 		}
 	}
 
-	if resolvedRef == nil {
+	err = errors.Join(errs...)
+	if err != nil {
 		return corev1alpha1.ResolvedSourceConfig{
 			Git: &corev1alpha1.ResolvedGitSource{
```
tomkennedy513: this is the same as emptyResult right?

mogsie: Indeed it is. Fixed it in c7a3de6d

pkg/git/remote_git_resolver_test.go
mogsie: I'm not sure I understand.  The entire test is run twice, once with the feature flag, once without:

```
func TestRemoteGitResolver(t *testing.T) {
	featureFlags.GitResolverUseShallowClone = false
	spec.Run(t, "TestRemoteGitResolverFetch", testRemoteGitResolver)
	featureFlags.GitResolverUseShallowClone = true
	spec.Run(t, "TestRemoteGitResolverClone", testRemoteGitResolver)
}
```

When testing with the feature flag enabled, the Tree field is filled out; it isn't when the feature flag is disabled.

pkg/git/remote_git_resolver_clone.go
mogsie: Simplified it in eb8b177a

pkg/git/remote_git_resolver_test.go
tomkennedy513: Ya I totally missed that somehow

tomkennedy513 (approved):

Issue #207, opened 2019-10-31
Document/support way of providing CA file needed to talk to image registry.


GrahamDumpleton:
Mentioned this in https://github.com/pivotal/kpack/issues/196#issuecomment-547232594 but creating separate issue to give better visibility since that other issue was for different problem.

Issue is how can one supply a CA file fir secure SSL connection when taking to an image registry?

I can't see that this is documented, nor can I determine if it is supported in some way.

For an image registry hosted in the cluster, the normal CA used with secure connections internal to the registry may be used. Or a separate non public CA might be used. In either case, need a way of supplying the CA file in the secret with login credentials.

matthewmcnew:
At this time we don't currently have a mechanism to provide a CA cert for the image registry. 

Theoretically you could add support by adding the CA cert to a custom builder image and mounting the CA cert in the kpack controller via an init container.

p-alexisv:
I was able to successfully get kpack to connect to a private registry (Harbor) with self-signed certs.  

I had to clone two images (for ClusterBuilder and BUILD_INIT_IMAGE) and then copy the CA cert file into these (in /etc/ssl/certs) and then upload them to a registry.  Then, modified the ClusterBuilder and the kpack-controller Deployment to use my custom images.

I also had to create a configmap with the CA cert data.  Then, modified the kpack-controller Deployment to create a volume using that configmap and had it mounted (in /etc/ssl/certs) to the controller container.

In all 3 cases, the filename of the CA cert is in the hashed format (e.g., 9xx2x831.0) when I put it in /etc/ssl/certs.

cameronbanowsky:
@p-alexisv : when you say you uploaded your custom images, did you use your private registry, or did you need to use a public registry at that point?

cameronbanowsky:
Cool thank you @p-alexisv ... I will update this thread to show our hack to this problem

jeffellin:
> ClusterBuilder

@p-alexisv when you say ClusterBuilder Image what image are you referring to?  In my case the builder never becomes ready because it can't pull from harbor.  I updated the build-init-image as you described but am unsure which one I need to update for the ClusterBuilder

mattmoor:
I'm guessing this boils down to your usage of github.com/google/go-containerregistry, which we (knative/serving) use for digest resolution.  I'd wager a similar workaround to what we have documented here would work for kpack as well (though I haven't tried it): https://knative.dev/docs/serving/tag-resolution/#custom-certificates

[...]

PR #1946, opened 2026-06-11
Allow providing custom CA to images created by controller

mtatheonly:
As the images are being built with pack, it's possible to use ca-certificates buildpack features to allow providing a custom CA (referenced in #207). This commit adds the environment variable and mounts for the service bindings needed by the ca-certificates runtime (the helper is added to image /layers/paketo-buildbacks_ca-certificates/helper/helper)

### DIFF
--- a/pkg/apis/build/v1alpha2/build_pod.go
+++ b/pkg/apis/build/v1alpha2/build_pod.go
@@ -299,6 +299,7 @@ func (b *Build) BuildPod(images BuildPodImages, buildContext BuildContext) (*cor
 		ImagePullPolicy: corev1.PullIfNotPresent,
 		Env: []corev1.EnvVar{
 			platformApiVersionEnvVar,
+			serviceBindingRootEnv,
 		},
 		SecurityContext: containerSecurityContext(),
 	}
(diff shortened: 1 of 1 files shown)


tomkennedy513:
I'm not opposed to this but historically we have recommended the use of cert-injection-webhook for this case and the reasoning why we didnt do this is escaping me. i think it might be worthwhile to at least filter the bindings that are mounted so not everything is attached.

mtatheonly:
Hi @tomkennedy513, the mounts are the same as in the build container, just adding them to the build-init and completion steps. To provide custom CA it's just an example of things that can be done  (maybe rephrase the issue title?)

tomkennedy513:
> Hi @tomkennedy513, the mounts are the same as in the build container, just adding them to the build-init and completion steps. To provide custom CA it's just an example of things that can be done  (maybe rephrase the issue title?)

Outside of custom cas what use case is there for generic service bindings in build init and completion

mtatheonly:
right now I'm not devising another use other than providing custom CAs, but it might be possible the usage of others bindings into those containers

tomkennedy513:
> right now I'm not devising another use other than providing custom CAs, but it might be possible the usage of others bindings into those containers

i admittedly haven't thought a ton of possible attack vectors that might open up by allowing this so i may have no reason to be concerned, but I wonder if we should only mount volumes that match preapproved types (say ca-certificates for starters)

mtatheonly:
thinking a little more, I've remembered other kinds of bindings that could be mounted into aforementioned containers: dependency-mapping, dependency-mirror (according to https://paketo.io/docs/howto/configuration/)

tomkennedy513:
> thinking a little more, I've remembered other kinds of bindings that could be mounted into aforementioned containers: dependency-mapping, dependency-mirror (according to https://paketo.io/docs/howto/configuration/)

but those wouldn't need to be present in build init

Issue #1940, opened 2026-02-16
ARM64 support


brainfair:
Solution works perfectly on x86, but when I try to run it on an arm64 instance, I'm getting an error `exec /cnb/process/controller: no such file or directory`, which is obviously an issue with the wrong architecture. Is there any way to run the controller and builders on arm64? This also limits me from building arm64 images.

tpaul1611:
I also ran into the same issue, the controller injects build-init, build-waiter, completion, and rebase images into build pods that are all not available for arm64. Neither is the image for the kpack-controller.

tpaul1611:
I built all the images myself for am64, and hosted kpack on an arm64 machine, but it still doesnt work because kpack seems to be hardwired to pull amd64 images.

tomkennedy513:
Hi @brainfair kpack doesnt currently support arm64 as you have seen. The main reason being that we never had a strong reason to support it considering that the vast majority of k8s clusters were amd64 and there is very little buildpack support for arm64. I wouldn't be opposed to someone implementing it, but I don't think it's something I would want to take on personally right now.

PR #1949, opened 2026-06-19
feat: Support running kpack on arm64 clusters

BaptisteCentreon:
## Why

We run our clusters on graviton arm nodes. They cost around 20% less than the amd equivalents at comparable performance.

Today kpack's component images are published linux/amd64 only, so kpack pods crash on arm64 with `exec format error` (#1940), and on a mixed cluster nothing pins a build to a node matching its builder's architecture.

## What changed

This builds the six component images (controller, webhook, build-init, build-waiter, rebase, completion) as multi-arch amd64+arm64 with ko, using the work from #1947, but now wired into CI. ko cross-compiles per arch and the platforms are set with `IMAGE_PLATFORMS`.

It also adds a kubernetes.io/arch node selector to each build pod, matching the builder's architecture, so the pod lands on a compatible node. Users can't set kubernetes.io/arch themselves, the same way kubernetes.io/os is already blocked (RFC 0008). Rebase pods don't run the builder, so they're left alone. Existing amd64 users see no change.

Finally, I updated the samples with jammy stack to match docs/stack.md.

## Validation

Verified on a real arm64 cluster: the controller and webhook run, and a Go app builds end-to-end into a linux/arm64 image.

## Note

To build arm64 apps, the ClusterStack and ClusterLifecycle also need to point at arm64 images. kpack still defaults to amd64 when reading multi-arch images (#1941), so for now those have to be pinned to arm64 digests. This PR is safe on its own and doesn't depend on that.


BaptisteCentreon:
Hey @tomkennedy513, If you get some time, I'd really appreciate a review :)

tomkennedy513:
Hi @BaptisteCentreon sorry for the delay. We definitely want to continue building our production images with buildpacks. I was playing around this weekend using pack with and I think this might work but I haven't thoroughly tested it. (https://github.com/buildpacks-community/kpack/tree/review-1949-pack-multiarch). I Ran the ci job off this branch if you want to test the built images work for you https://github.com/buildpacks-community/kpack/actions/runs/28547019135 and if they do we can pull those changes into your branch.

tomkennedy513:
> Hi @BaptisteCentreon sorry for the delay. We definitely want to continue building our production images with buildpacks. I was playing around this weekend using pack with and I think this might work but I haven't thoroughly tested it. (https://github.com/buildpacks-community/kpack/tree/review-1949-pack-multiarch). I Ran the ci job off this branch if you want to test the built images work for you https://github.com/buildpacks-community/kpack/actions/runs/28547019135 and if they do we can pull those changes into your branch.

update: ended up just opening the pr to your fork in case that's easier https://github.com/BaptisteCentreon/kpack/pull/1

BaptisteCentreon:
Thanks Tom, and no worries at all about the delay!

Happy to keep using buildpacks, I pulled your changes into the branch.

I tested it end to end with both M4 locally and Graviton on EKS. The images run fine on arm64, and the build pods land on the right architecture. To build arm64 I just had to pin the stack/store/lifecycle to their arm64 digests (#1941).

Everything looked good.

tomkennedy513 (approved):

tomkennedy513:
lgtm, @BaptisteCentreon do you want to pick up #1941 to drive it over the line. The main thing it needs is testing.

BaptisteCentreon:
Of course, I'll have a look :)

Thanks for merging.

PR #1951, opened 2026-07-02
 fix: resolve images for the current architecture in the registry client

BaptisteCentreon:
to follow up on #1949, with this, kpack running on arm64 resolves stacks/builders to arm64 automatically, without pinning digests.

Continues #1941 by @tpaul1611 (cherry-picked here) and adds the unit test the maintainers asked for in review.

kpack's registry client always pulled linux/amd64, so on arm64 it read the wrong manifest. This uses runtime.GOARCH instead, fixing multi-arch resolution on arm64 while keeping amd64 unchanged.

### DIFF
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -3,6 +3,7 @@ package registry
 import (
 	"fmt"
 	"net/http"
+	"runtime"
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/authn"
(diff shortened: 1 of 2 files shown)


BaptisteCentreon:
@tomkennedy513 as requested :)

tomkennedy513 (approved):

brainfair:
Hi, I see that this change is not included in a published release yet. Is it possible to schedule a new release? Thank you.

tomkennedy513:
> Hi, I see that this change is not included in a published release yet. Is it possible to schedule a new release? Thank you.

just cut a release

